PR #3, opened 2022-07-07
DOC Update references for how to update documentation (4 branch)

GuySartorelli:
Also add composer.json and readme.

composer.json is needed so that cow can install the docs and use them to get commits for the changelog and to commit the changelog.

## Parent issue
- https://github.com/silverstripe/silverstripe-framework/issues/10215

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,23 @@
+{
+    "name": "silverstripe/developer-docs",
+    "description": "Developer documentation for the Silverstripe framework and CMS. Added to packagist for release purposes.",
+    "homepage": "https://silverstripe.org",
+    "license": "CC-BY-3.0",
+    "keywords": [
+        "silverstripe",
+        "framework",
+        "cms",
+        "developer",
+        "documentation"
+    ],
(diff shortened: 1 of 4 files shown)


GuySartorelli:
After merging, add this repo to packagist.

composer.json
```diff
@@ -0,0 +1,23 @@
+{
```
emteknetnz, line 1: Why do we need a composer file here?

If we need it, worth just writing this explanation into the description attribute or somewhere else in this file to explain to future generations

emteknetnz (changes requested):

composer.json
ChloeSartorelli: Per the description in this PR:
> composer.json is needed so that cow can install the docs and use them to get commits for the changelog and to commit the changelog.

It also means we already have it checked out during development which makes it easier to update docs as we go.

I'll update the composer.json description.

ChloeSartorelli: done

emteknetnz (approved):

PR #5, opened 2022-07-07
DOC Document searching across fields by default

GuySartorelli:
Docs for https://github.com/silverstripe/silverstripe-framework/pull/10382

## Parent issue
- https://github.com/silverstripe/silverstripe-framework/issues/9356

### DIFF
--- a/en/02_Developer_Guides/00_Model/11_Scaffolding.md
+++ b/en/02_Developer_Guides/00_Model/11_Scaffolding.md
@@ -18,9 +18,8 @@ An example is `DataObject`, Silverstripe CMS will automatically create your CMS 
 ```php
 use SilverStripe\ORM\DataObject;
 
-class MyDataObject extends DataObject 
+class MyDataObject extends DataObject
 {
-    
     private static $db = [
         'IsActive' => 'Boolean',
         'Title' => 'Varchar',
@@ -78,7 +77,7 @@ your `$summary_fields` config specifies fields that are not stored in the databa
 ```php
 use SilverStripe\ORM\DataObject;
 
-class MyDataObject extends DataObject 
+class MyDataObject extends DataObject
 {
 
    private static $searchable_fields = [
@@ -88,18 +87,93 @@ class MyDataObject extends DataObject
 }
 ```
 
+### General search field
+
+Tabular views such as `GridField` or `ModalAdmin` include a search bar. As of Silverstripe CMS 4.12, the search bar will search across all of your searchable fields by default. It will return a match if the search terms appear in any of the searchable fields.
+
+#### Exclude fields from the general search
+
+If you have fields which you do _not_ want to be searched with this general search (e.g. date fields which need special consideration), you can mark them as being explicitly excluded by setting `general` to false in the searchable fields configuration for that field:
+
+```php
+use SilverStripe\ORM\DataObject;
+
+class MyDataObject extends DataObject
+{
+    private static $searchable_fields = [
+        'Name',
+        'BirthDate' => [
+            'general' => false
+        ],
+    ];
+}
+```
+
+#### Customise the general search field name
+
+By default the general search field uses the name "q". If you already use that field name or search query in your [SearchContext](/developer_guides/search/searchcontext), you can change this to whatever name you prefer either globally or per class:
+
+[hint]
+If you set `general_search_field_name` to any empty string, general search will be disabled entirely. Instead, the first field in your searchable fields configuration will be used, which was the default behaviour prior to Silverstripe CMS 4.12.
+[/hint]
+
+**Globally change the general search field name via yaml config**
+```yml
+SilverStripe\ORM\DataObject:
+  general_search_field_name: 'my_general_field_name'
+```
+
+**Customise the general search field name via yaml _or_ php config**
+```php
+use SilverStripe\ORM\DataObject;
+
+class MyDataObject extends DataObject
+{
+    private static string $general_search_field_name = 'my_general_field_name';
+}
+```
+
+#### Specify a search filter for general search
+
+By default, the general search will search across your fields using a [PartialMatchFilter](api:SilverStripe\ORM\Filters\PartialMatchFilter) regardless of what filters you have [specified for those fields](#specify-a-form-field-or-search-filter).
+
+You can configure this to be a specific filter class, or else disable the general search filter. Disabling the filter will result in the filters you have specified for each field being used when searching against that field in the general search.
+
+Like the general search field name, you can set this either globally or per class:
+
+**Globally change the general search filter via yaml config**
+```yml
+# use a specific filter
+SilverStripe\ORM\DataObject:
+  general_search_field_filter: 'SilverStripe\ORM\Filters\EndsWithFilter'
+
+# or disable the filter to fall back on individual fields' filters
+SilverStripe\ORM\DataObject:
+  general_search_field_filter: ''
+```
+
+**Customise the general search filter via yaml _or_ php config**
+```php
+use SilverStripe\ORM\DataObject;
+use SilverStripe\ORM\Filters\EndsWithFilter;
+
+class MyDataObject extends DataObject
+{
+    private static string $general_search_field_name = EndsWithFilter::class;
+}
+```
+
 ### Specify a form field or search filter
 
 Searchable fields will appear in the search interface with a default form field (usually a [TextField](api:SilverStripe\Forms\TextField)) and a 
-default search filter assigned (usually an [ExactMatchFilter](api:SilverStripe\ORM\Filters\ExactMatchFilter)). To override these defaults, you can specify 
+default search filter assigned (usually a [PartialMatchFilter](api:SilverStripe\ORM\Filters\PartialMatchFilter)). To override these defaults, you can specify
 additional information on `$searchable_fields`:
 
 ```php
 use SilverStripe\ORM\DataObject;
 
-class MyDataObject extends DataObject 
+class MyDataObject extends DataObject
 {
-
     private static $searchable_fields = [
         'Name' => 'PartialMatchFilter',
         'ProductCode' => NumericField::class
@@ -107,15 +181,14 @@ class MyDataObject extends DataObject
 }
 ```
 
-If you assign a single string value, you can set it to be either a [FormField](api:SilverStripe\Forms\FormField) or [SearchFilter](api:SilverStripe\ORM\Filters\SearchFilter). To specify 
-both, you can assign an array:
+If you assign a single string value, you can set it to be either a [FormField](api:SilverStripe\Forms\FormField) or [SearchFilter](api:SilverStripe\ORM\Filters\SearchFilter). To specify
+both or to combine this with other configuration, you can assign an array:
 
 ```php
 use SilverStripe\ORM\DataObject;
 
-class MyDataObject extends DataObject 
+class MyDataObject extends DataObject
 {
-
    private static $searchable_fields = [
        'Name' => [
           'field' => TextField::class,
@@ -171,6 +244,10 @@ class Player extends DataObject
 
 Use a single search field that matches on multiple database fields with `'match_any'`. This also supports specifying a field and a filter, though it is not necessary to do so.
 
+[alert]
+If you don't specify a field, you must use the name of a real database field instead of a custom name so that a default field can be determined.
+[/alert]
+
 ```php
 class Order extends DataObject
 {
@@ -198,10 +275,6 @@ class Order extends DataObject
 }
 ```
 
-[alert]
-If you don't specify a field, you must use the name of a real database field instead of a custom name so that a default field can be determined.
-[/alert]
-
 ## Summary Fields
 
 Summary fields can be used to show a quick overview of the data for a specific [DataObject](api:SilverStripe\ORM\DataObject) record. The most common use 
@@ -210,7 +283,7 @@ is their display as table columns, e.g. in the search results of a [ModelAdmin](
 ```php
 use SilverStripe\ORM\DataObject;
 
-class MyDataObject extends DataObject 
+class MyDataObject extends DataObject
 {
     private static $db = [
         'Name' => 'Text',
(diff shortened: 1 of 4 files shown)


en/02_Developer_Guides/00_Model/11_Scaffolding.md
```diff
@@ -88,34 +87,138 @@ class MyDataObject extends DataObject
...
+}
+```
+
+[warning]
+You may get unexpected results using some filters if you don't disable splitting the query into terms - for example if you use an [ExactMatchFilter](api:SilverStripe\ORM\Filters\ExactMatchFilter), each term in the query _must_ exactly match the value in at least one field to get a match. If you disable splitting terms, the whole query must exactly match a field value instead.
+[/warning]
+
+#### Splitting search queries into individual terms
+
+By default the primary search field qill split your search query into individual terms, and search across your searchable field for each term. At least one field must match each term to get a match.
```
emteknetnz:
```suggestion
By default the primary search field will split your search query on spaces into individual terms, and search across your searchable field for each term. At least one field must match each term to get a match.
```

emteknetnz: > At least one field must match each term to get a match.

May need to change this based on decision around matching all or matching a single term

emteknetnz (changes requested):

en/02_Developer_Guides/00_Model/11_Scaffolding.md
```diff
@@ -88,34 +87,138 @@ class MyDataObject extends DataObject
...
+
+For example: with the search query "farm house" at least one field must have a match for the word "farm", and at least one field must have a match for the word "house". There does _not_ need to be a field which matches the full phrase "farm house".
+
+You can disable this behaviour by setting `DataObject.primary_search_split_terms` to false. This would mean that for the example above a `DataObject` would need a field that matches "farm house" to be included in the results. Simply amtching "farm" or "house" alone would not be sufficient.
```
emteknetnz:
```suggestion
You can disable this behaviour by setting `DataObject.primary_search_split_terms` to false. This would mean that for the example above a `DataObject` would need a field that matches "farm house" to be included in the results. Simply matching "farm" or "house" alone would not be sufficient.
```

GuySartorelli:
Requested changes made. Pending UX decision.

en/02_Developer_Guides/00_Model/11_Scaffolding.md
```diff
@@ -88,34 +87,137 @@ class MyDataObject extends DataObject
 }
 ```
 
+### General search field
+
+Tabular views such as `GridField` or `ModalAdmin` include a search bar. As of Silverstripe CMS 4.12, the search bar will search across all of your searchable fields by default. It will return a match if the search terms appear in any of the searchable fields.
+
+#### Exclude fields from the general search
+
+If you have fields which you do _not_ want to be searched with this general search (e.g. date fields which need special consideration), you can mark them as being explicitly excluded by setting `general` to false in the searchable fields configuration for that field:
```
emteknetnz, line 96: I think we'll need to mention that DataExtension's do not work for switching things off, e.g.

```yml
---
Name: myextensions
After: '*'
---

SilverStripe\Security\Member:
  extensions:
    - MemberExtension
```

```php
<?php

use SilverStripe\ORM\DataExtension;

class MemberExtension extends DataExtension
{
    private static $searchable_fields = [
        'Surname' => [
            'general' => false
        ],
    ];
}
```

Doesn't work

en/02_Developer_Guides/00_Model/11_Scaffolding.md
```diff
@@ -88,34 +87,137 @@ class MyDataObject extends DataObject
...
+
+You can configure this to be a specific filter class, or else disable the general search filter. Disabling the filter will result in the filters you have specified for each field being used when searching against that field in the general search.
+
+Like the general search field name, you can set this either globally or per class:
+
+**Globally change the general search filter via yaml config**
+```yml
+# use a specific filter
+SilverStripe\ORM\DataObject:
+  general_search_field_filter: 'SilverStripe\ORM\Filters\EndsWithFilter'
```
emteknetnz, line 148: I think the implementation here might need to be changed

We're defaulting this to `PartialMatchFilter::class` in the DataObject, so right out of the gate we're essentially overwriting and configured 'filter' on individual fields.  This essentially makes this an opt out.  Yes we should default to PartialMatchFilter, though I don't think it should have precedence over configured filters

e.g. a project with

```php
    private static $searchable_fields = [
        'Title' => [
            'filter' => EndsWithFilter::class,
        ],
```
Would search with PartialMatchFilter in the general search, though in the specific search it would match on EndsWithFilter.  That's inconsistent.

en/02_Developer_Guides/00_Model/11_Scaffolding.md
```diff
@@ -88,34 +87,137 @@ class MyDataObject extends DataObject
...
+```
+
+**Customise the general search filter via yaml _or_ php config**
+```php
+use SilverStripe\ORM\DataObject;
+use SilverStripe\ORM\Filters\EndsWithFilter;
+
+class MyDataObject extends DataObject
+{
+    private static string $general_search_field_name = EndsWithFilter::class;
```
emteknetnz: This should be `$general_search_field_filter` ?

en/02_Developer_Guides/00_Model/11_Scaffolding.md
```diff
@@ -88,34 +87,137 @@ class MyDataObject extends DataObject
...
+#### Exclude fields from the general search
+
+If you have fields which you do _not_ want to be searched with this general search (e.g. date fields which need special consideration), you can mark them as being explicitly excluded by setting `general` to false in the searchable fields configuration for that field:
+
+```php
+use SilverStripe\ORM\DataObject;
+
+class MyDataObject extends DataObject
+{
+    private static $searchable_fields = [
```
emteknetnz, line 103:
```suggestion
    private static array $searchable_fields = [
```

en/04_Changelogs/4.12.0.md
```diff
@@ -20,6 +21,28 @@ While it is still advised that you perform your own due diligence when upgrading
...
+The general search text field which appears in the search interface now searches across all of your searchable fields by default. It will return a match if the search terms appear in any of the searchable fields.
+
+If you have fields which you do _not_ want to be searched with this general search (e.g. date fields which need special consideration), you can mark them as being explicitly excluded by setting `general` to false in the searchable fields configuration for that field:
+
+```php
+use SilverStripe\ORM\DataObject;
+
+class MyDataObject extends DataObject
+{
+    private static $searchable_fields = [
```
emteknetnz, line 35:
```suggestion
    private static array $searchable_fields = [
```

en/04_Changelogs/4.12.0.md
```diff
@@ -20,6 +21,28 @@ While it is still advised that you perform your own due diligence when upgrading
 
 ## Features and enhancements {#features-and-enhancements}
 
+### Search multiple `searchable_fields` by default {#general-search-field}
+
+The general search text field which appears in the search interface now searches across all of your searchable fields by default. It will return a match if the search terms appear in any of the searchable fields.
```
emteknetnz, line 26:
```suggestion
The general search text field which appears in the search interface now searches across all of your searchable fields by default. It will return a match if the search terms appears in any of the searchable fields.
```

emteknetnz (changes requested):

en/02_Developer_Guides/00_Model/11_Scaffolding.md
ChloeSartorelli: This is a bug: https://github.com/silverstripe/silverstripe-framework/issues/10405
We could mention it in the changelog but I definitely wouldn't want to mention it in the main doc. There's no precedent for pointing out bugs in the docs.

en/02_Developer_Guides/00_Model/11_Scaffolding.md
ChloeSartorelli: > I don't think it should have precedence over configured filters

The way I see this is that the configured filters are _primarily_ for the specific search fields in the dropdown. They are separate from the functionality for the general search field. I would find it very strange if when I search "Big mac" in the general search if doesn't find someone who's last name is "Cormac" because someone decided last name should have a `StartsWithFilter`.

If developers do explicitly want their specific search field filters applied to the general search they can configure it to do that, but my instinct is that it's not intuitive to do so by default.
Might be another question for Cassie though.

en/02_Developer_Guides/00_Model/11_Scaffolding.md
ChloeSartorelli: Can we do that if it's not specified as `array` in `DataObject`?

en/04_Changelogs/4.12.0.md
ChloeSartorelli: "search terms" is plural. "search terms appear" is grammatically correct.

en/02_Developer_Guides/00_Model/11_Scaffolding.md
emteknetnz: > If developers do explicitly want their specific search field filters applied to the general search they can configure it to do that

How would they do that?

ChloeSartorelli: > You can configure this to be a specific filter class, or else disable the general search filter. Disabling the filter will result in the filters you have specified for each field being used when searching against that field in the general search.

en/02_Developer_Guides/00_Model/11_Scaffolding.md
ChloeSartorelli: Yes, oops. Fixed.

en/02_Developer_Guides/00_Model/11_Scaffolding.md
ChloeSartorelli: Discussed this in person - probably safer to not make this change for now

en/02_Developer_Guides/00_Model/11_Scaffolding.md
emteknetnz: So to confirm, you CAN'T have say:
- PartialMatchFilter be the default on general search
- AND have PhoneNumber use ExactMatchFilter in the general search

If that's the case, this seems wrong. You shouldn't have to disable general search filter just to use the ExactMatchFilter on PhoneNumber

Also, if you disable the general search filter ... presumably the general search still works? What does it then essentially fall back to an ExactMatchFilter?

ChloeSartorelli: No, you _can_ have that. Because `PartialMatchFilter` is the default for all searchable fields.

So the options are:
1. Always use a specific filter for _all_ fields when using general search. By default this is `PartialMatchFilter`
    - Whatever you configure for the specific fields will be used _only_ when using specific field search fields from the dropdown.
3. Always use the specific field filters when using general search. By default this is `PartialMatchFilter`
    - Any fields you have given a specific filter for the specific fields will apply to both the specific field search in the dropdown _and_ the search against that field in the general search.
    
    
What you cannot do is define a specific filter for two different fields, but only have one of those apply for the general search filter. We could add that as an enhancement further down the line, but I think it adds too much complexity of configuration to add that right now especially since it's an edge case.

ChloeSartorelli: > if you disable the general search filter ... presumably the general search still works?

Yes. It will work, and use whatever filters are defined for the specific fields when searching against that column in the general search.

> What does it then essentially fall back to an ExactMatchFilter?

I don't understand this question.

emteknetnz: If someone disables the general search filter via config (default set in the PR is PartialMatchFilter) and the general search still functions, then the general search has to use a filter of some variety correct? Seems like it needs to be either a PartialMatchFilter or an ExactMatchFilter.  Even if it's not actually a Silverstripe filter, the database needs some form of logic to match values?

ChloeSartorelli: > If someone disables the general search filter via config (default set in the PR is PartialMatchFilter) and the general search still functions, then the general search has to use a filter of some variety correct?

Yes.
The way it works it that it applies a filter for _each field_. By default this is always `PartialMatchFilter`. If you set this filter to a blank string in the config, it will instead check which filter is defined for that specific field (in `searchable_fields`) and use its filter. By default that will _also_ be `PartialMatchFilter`.

emteknetnz: OK thanks for clarifying.  Might be worth a bit of a wider discussion to see if we should have PartialMatchFilter as a default for the general filter then.

As things are, this is essentially an opt-out which is generally more of a breaking change (to be fair, the general search was very broken to begin with). The other option is to leave it blank and make it in opt-in to override what's already set on search fields. Things like phone numbers and SKU's might make more sense to have ExactMatchFilters.  To me the general and specific search should both respect the any configured filters

ChloeSartorelli: @maxime-rainville any views on this?

ChloeSartorelli: Note that `ExactMatchFilter` might give unexpected results with split terms, as per the documentation.

maxime-rainville: Just summarising what I understood of the argument. You can choose which filter to use for each search field.

The proposed behaviour is to have the General field default to using partial match with the option to use a different filter via YML config. You can choose to customised the advanced filters on an individual basis, but that doesn't affect the general search.

The alternative we are considering here is that the general field should be driven by the filters of each individual advanced filter. e.g. I decide that the "Last name" filter will use "ends with" matching ... I search for "bob" ... all the field use partial matching except "last name" where it uses "ends with".

Presuming, I understood this correctly. I'm going to go with the first option: the general search field should use the same match strategy across all the fields and ignore the advanced filters.

In my mind, the General search field is completely disconnected from the advanced filters. It's the dumb field where people just throw a few words in and hope for a mildly sensible list of results. If you want to further filter your results, you can then choose to define an advanced filter that will apply on top of your general search.

emteknetnz (approved):

PR #7, opened 2022-07-07
MNT Add action for building docs

GuySartorelli:
Add the action for building docs, copied over from framework.
I haven't pulled the history for the action like I did for docs, but you can see that history [here](https://github.com/silverstripe/silverstripe-framework/commits/4/.github/workflows/main.yml) if you're in the future and are curious.

## DEPENDENCY
This PR must not be merged before https://github.com/silverstripe/doc.silverstripe.org/pull/243

## Parent issue
- https://github.com/silverstripe/silverstripe-framework/issues/10215

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,16 @@
+name: Build Docs
+on:
+  push:
+    branches:
+      - '3'
+      - '4.11'
+    paths:
+      - 'en/**'
+jobs:
+  build:
+    name: build-docs
+    runs-on: ubuntu-latest
+    steps:
+      - name: Run build hook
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,16 @@
+name: Build Docs
+on:
+  push:
+    branches:
+      - '3'
+      - '4.11'
```
emteknetnz, line 6: Is the intention here to update 4.11 to 4.12 as part of the release process?

emteknetnz (changes requested):

.github/workflows/build.yml
ChloeSartorelli: Yup. Ideally that will be automated per https://github.com/silverstripe/doc.silverstripe.org/issues/175

emteknetnz (approved):

PR #8, opened 2022-07-12
DOCS update for silverstripe-framework#10134

gurucomkz:
Docs update for https://github.com/silverstripe/silverstripe-framework/pull/10134

### DIFF
--- a/en/02_Developer_Guides/01_Templates/03_Requirements.md
+++ b/en/02_Developer_Guides/01_Templates/03_Requirements.md
@@ -240,6 +240,9 @@ You can also use any of the below options in order to tweak this behaviour:
    other systems, it is usually necessary to distinguish combined files from other assets.
  * `Requirements_Backend.combine_in_dev` - By default combined files will not be combined except in test
    or live environments. Turning this on will allow for pre-combining of files in development mode.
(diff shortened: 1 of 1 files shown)


en/02_Developer_Guides/01_Templates/03_Requirements.md
```diff
@@ -256,6 +259,7 @@ SilverStripe\View\Requirements_Backend:
   combine_in_dev: true
   combine_hash_querystring: true
   default_combined_files_folder: 'combined'
+  resolve_relative_css_refs: true
```
ChloeSartorelli, line 262: I don't think this is needed since it's true by default?

GuySartorelli (changes requested):

en/02_Developer_Guides/01_Templates/03_Requirements.md
ChloeSartorelli: Actually that's a good point it probably shouldn't be true by default for BC reasons... can you please amend the docs above to say "you can set this to true to enable it"? I'll note on the other PR to change default to false.

Once it's merged, you can create a PR targetting `5` to set it to true in the next major

gurucomkz: I changed the last sentence to "Disabled by default"

GuySartorelli (approved):

Issue #44, opened 2022-07-18
CMS 5 developer doc and user help doc are set up


maxime-rainville:
# Acceptance criteria
- [x] CMS 5 branches are set up for user doc and developer doc
- [x] CMS 5 content can be seen on developer doc and user doc and API doc.
- [x] CMS 5 content is clearly marked as "pre-release"
- [x] There's a clear demarcation between CMS4 and CMS5 doc.
- [x] Until CMS5 is stable, the docs default to 4
- ~~CMS 3 content no longer shows up on doc site and API Doc.~~

# Notes 
- Will probably need some UX/design feedback

## PRs
- https://github.com/silverstripe/developer-docs/pull/47
- https://github.com/silverstripe/silverstripe-userhelp-content/pull/175
- https://github.com/silverstripe/doc.silverstripe.org/pull/244
- https://github.com/silverstripe/api.silverstripe.org/pull/102

maxime-rainville:
@silverstripe/core-team Do you have any views on removing the CMS3 doc? It's been EOL for close to a year now.

At the very least, if we keep it there, we should call attention to the fact that it's no longer supported

emteknetnz:
I think it's worth keeping because the cost of doing so is very low and there will be some projects that simply never update.  There should be a very clear UX piece such as a big red warning in the header, and make it very easy to find your way to the 5/4 docs.  

We should also ensure that search engines no longer index the content, and while we're at it we should have something in place to push live to no longer index 4 content once 5 goes stable

GuySartorelli:
Instead of simply not indexing the content, would it make more sense to only show results for the version you're currently viewing? So looking at V4 you see V4 results, looking at v5 you see v5 results etc

emteknetnz:
Not index means tell google to not index it in google search.  That way people won't land on v3 results from google search.

GuySartorelli:
Ahh, I thought you meant the search on the site itself. Yup, that's fair.

emteknetnz:
> Instead of simply not indexing the content, would it make more sense to only show results for the version you're currently viewing? So looking at V4 you see V4 results, looking at v5 you see v5 results etc

Yeah in the search on site context that probably makes sense

GuySartorelli:
Updating the userhelp doc sources assumes the following branches are created:
- https://github.com/symbiote/silverstripe-advancedworkflow
  branch: 6
- https://github.com/silverstripe/silverstripe-registry
  branch: 3          
[...]

PR #47, opened 2022-07-21
DOC Update references to CMS 3 and 4

GuySartorelli:
For CMS5 docs, we don't care about CMS3. We also want anything saying that 4 is the latest to instead refer to 5.
This PR also removes some references to specific CMS major versions because the language doesn't need to be that specific. This means future major versions won't need to update those same references.

## Parent issue:
- https://github.com/silverstripe/developer-docs/issues/44


en/02_Developer_Guides/09_Security/04_Secure_Coding.md
```diff
@@ -710,7 +710,7 @@ following in your .htaccess to ensure this behaviour is activated.
 </IfModule>
 ```
 
-As of Silverstripe CMS 4, this behaviour is on by default, and the environment variable is no longer required. For correct operation, it is necessary to always set `SS_TRUSTED_PROXY_IPS` if using a proxy.
+This behaviour is on by default; the environment variable is not required. For correct operation, it is necessary to always set `SS_TRUSTED_PROXY_IPS` if using a proxy.
```
sabina-talipova, line 713: Is this correct to have a semicolon: 
`... behaviour is on by default ;  ...`

sabina-talipova:
Just tiny question. You removed almost all sections in `03_Upgrading` folder. Don't we need them anymore?

en/02_Developer_Guides/09_Security/04_Secure_Coding.md
ChloeSartorelli: Yup.
https://www.merriam-webster.com/words-at-play/a-guide-to-using-semicolons
> A semicolon separates related independent clauses that are joined without a coordinating conjunction, such as _and_:
> > Mabel the Cat had made a big pot of goulash; it was simmering on the stove.

GuySartorelli:
> You removed almost all sections in 03_Upgrading folder. Don't we need them anymore?

- `02_Upgrading_module.md` had explicit advice for upgrading modules from v3 to v4. It's possible we'll have something like this document for CMS5, but at this stage it's cleaner to remove it, and then create a new one if we have something that is relevant explicitly for modules.
- `03_Migrating_off_CWP.md` was relevant when the CWP recipes were replaced with ccl recipes. Anyone upgrading to CMS 5 will almost certainly have already migrated off the CWP recipes.
- `04_Upgrading_project.md` has already been replaced with [`02_Upgrading_your_project.md`](https://github.com/silverstripe/developer-docs/blob/5/en/03_Upgrading/02_Upgrading_your_project.md).
- `06_Upgrading_to_GraphQL_4.md` is probably still relevant. I'll update that one to be CMS5 centric instead of removing it.

PR #51, opened 2022-07-29
Update adding_middleware.md

AntonyThorpe:
When adding GraphQL query middleware, the `callMiddleware` function in the `QueryHandler` class expects four entries when returning `$next`.  Thanks.


GuySartorelli (approved):
Thanks. Not sure where that `$params` came from...

PR #79, opened 2022-09-05
DOC Document symfony 6 dependency changes

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10389

### DIFF
--- a/en/02_Developer_Guides/08_Performance/01_Caching.md
+++ b/en/02_Developer_Guides/08_Performance/01_Caching.md
@@ -9,12 +9,12 @@ icon: tachometer-alt
 
 The framework uses caches to store infrequently changing values.
 By default, the storage mechanism chooses the most performant adapter available
-(PHP7 opcache, APC, or filesystem). Other cache backends can be configured.
+(PHP opcache, APC, or filesystem). Other cache backends can be configured.
 
 The most common caches are manifests of various resources: 
 
  * PHP class locations ([ClassManifest](api:SilverStripe\Core\Manifest\ClassManifest))
- * Configuration settings from YAML files ([ConfigManifest](api:ConfigManifest))
+ * Configuration settings from YAML files ([CachedConfigCollection](api:SilverStripe\Config\Collections\CachedConfigCollection))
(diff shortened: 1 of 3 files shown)


GuySartorelli (changes requested):
I think the docs changes from https://github.com/silverstripe/silverstripe-framework/pull/10187/files are needed here as well.

emteknetnz:
Have added changes

GuySartorelli (changes requested):
Looks like there are still other mentions of the old classes, e.g. in https://docs.silverstripe.org/en/4/developer_guides/performance/caching
> Clearing caches by namespace is dependent on the used adapter: While the `FilesystemCache` adapter clears only the namespaced cache, a `MemcachedCache` adapter will clear all caches regardless of namespace, since the underlying memcached service doesn't support this. See "Invalidation" for alternative strategies.

There's also at least one link to symfony 5 docs (https://github.com/symfony/cache/tree/5.x/Adapter)

Might pay to do a word search for any of the old cache names, and for any mention of symfony to make sure we have no outdated references.

en/02_Developer_Guides/08_Performance/01_Caching.md
```diff
@@ -255,19 +256,16 @@ with the `disable-container` argument.
...
+```
 
 ## Additional Caches
 
 Unfortunately not all caches are configurable via cache adapters.
 
  * [SSViewer](api:SilverStripe\View\SSViewer) writes compiled templates as PHP files to the filesystem
    (in order to achieve opcode caching on `include()` calls)
- * [ConfigManifest](api:SilverStripe\Core\Manifest\ConfigManifest) is hardcoded to use `FilesystemCache`
- * [ClassManifest](api:SilverStripe\Core\Manifest\ClassManifest) and [ThemeManifest](api:SilverStripe\View\ThemeManifest)
```
emteknetnz, line 271: Have removed references to Config + Class manifests because they're not hardcoded to use a particular type of cache

emteknetnz:
@GuySartorelli Have updated

GuySartorelli (approved):

Issue #74, opened 2022-08-31
Migrate update content to changelogs for CMS 5


GuySartorelli:
I started out listing breaking changes in the update docs (with the intention that it would be used to create a more how-to style document later on), and then ended up putting a more how-to write up in the changelogs because that's the kind of content we normally add there - because we normally aren't doing major releases.

I think we should be listing the explicit breaking changes (removed x, use y instead) in the changelog, and then anything that requires further explanation should be in the upgrade docs.

This issue exists both to facilitate discussion about that in case anyone disagrees, and then to actually set up the docs in whichever manner is agreed upon.

## Acceptance criteria
- [x] It is clear for CMS 5 which content goes into the update docs and what content goes into the changelog
- [x] CMS5 upgrade guide should be folded back into changelog
- [x] Update existing PR that adds entry to the upgrade guide to target the changelog instead (BLOCKED pending the attached PR being merged)

## PRs
- https://github.com/silverstripe/developer-docs/pull/84

maxime-rainville:
We've decided to put everything in the CMS 5 changelog initially. As we got closer to CMS 5 Beta 1, we will re-evaluate the need for a dedicated upgrade guide.

GuySartorelli:
Note that I can't update the existing PRs until the attached PR is merged.

GuySartorelli:
Done

PR #80, opened 2022-09-07
DOC Document SecurityAdmin extending ModelAdmin

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1350

### DIFF
--- a/en/04_Changelogs/5.0.0.md
+++ b/en/04_Changelogs/5.0.0.md
@@ -82,6 +82,11 @@ This is a major release and as a result there are a number of breaking API chang
 
 ### General changes {#api-general}
 
+- [SecurityAdmin](api:SilverStripe\Admin\SecurityAdmin) now extends [ModelAdmin](api:SilverStripe\Admin\ModelAdmin) instead of [LeftAndMain](api:SilverStripe\Admin\LeftAndMain) so that it can make use of standard API. This was done to remove duplicate code.
+  - `MemberImportForm.php` has been removed
+  - `GroupImportForm.php` has been removed
+  - Previous extension hook implementations of `SecurityAdmin::updateEditForm($form)` should still work as the extension hook is called in `ModelAdmin::getEditForm()`
+  - The Users, Groups and Roles tabs now no longer share the /admin/security path and instead have their own dedicated paths e.g. /admin/security/users
 - `isDev` and `isTest` querystring arguments have been removed due to security concerns (see [ss-2018-005](https://www.silverstripe.org/download/security-releases/ss-2018-005/)).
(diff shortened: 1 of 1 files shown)


en/04_Changelogs/5.0.0.md
```diff
@@ -15,6 +15,8 @@ title: 5.0.0 (unreleased)
 
 This is a major release and as a result there are a number of breaking API changes. For a full list of these see [upgrading your project](/upgrading/upgrading_your_project). Some specific details about a few of them are below.
 
+- SecurityAdmin now extends ModelAdmin instead of LeftAndMain. This was done to remove duplicate code.
```
ChloeSartorelli: Probably want more details in here than that - e.g. some paths will be changing, extension hooks are likely to have changed, etc.

GuySartorelli (changes requested):

GuySartorelli:
Rebased on `5` as a part of https://github.com/silverstripe/developer-docs/issues/74 - but the review still stands.

en/04_Changelogs/5.0.0.md
emteknetnz: Updated.  I don't think the changes in path are probably too much details as they're more of a UX thing.

ChloeSartorelli: The changes in paths may have implications for customisations that call custom actions in extensions, for example. It's worth highlighting IMO and there's no harm in doing so.

emteknetnz: Updated

en/04_Changelogs/5.0.0.md
```diff
@@ -82,6 +82,11 @@ This is a major release and as a result there are a number of breaking API chang
 
 ### General changes {#api-general}
 
+- [SecurityAdmin](api:SilverStripe\Admin\SecurityAdmin) now extends [ModelAdmin](api:SilverStripe\Admin\ModelAdmin) instead of [LeftAndMain](api:SilverStripe\Admin\LeftAndMain) so that it can make use of standard API. This was done to remove duplicate code.
+  - `MemberImportForm.php` has been removed
+  - `GroupImportForm.php` has been removed
+  - Previous extension hook implementations of `SecurityAdmin::updateEditForm($form)` should still work as the extension hook is called in `ModelAdmin::getEditForm()`
```
ChloeSartorelli, line 88: The actual form is slightly different so the extensions may need to be reworked slightly to work with the way modeladmin's form is produced.... but we can work out the exact wording for that closer to release. For now this is good enough.

GuySartorelli (approved):

PR #87, opened 2022-09-14
DOC Document making models previewable via extensions

GuySartorelli:
Documents changes made in https://github.com/silverstripe/silverstripe-admin/pull/1363 and https://github.com/silverstripe/silverstripe-cms/pull/2779

## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1362


sabina-talipova (approved):

sabina-talipova (commented):
@GuySartorelli , could we provide a small example of SiteConfig extension to use the following SiteConfig improvements https://github.com/silverstripe/silverstripe-siteconfig/pull/127

GuySartorelli:
Making `SiteConfig` previewable is no different from making any other vendor `DataObject` subclass previewable - if someone follows the existing documentation including the notes about `Extension`s then they will be able to apply it to `SiteConfig`. I don't think it's valuable to have an example explicitly for `SiteConfig` itself.

PR #100, opened 2022-10-05
Remove duplicate docs around PHPUnit upgrade

scott1702:


GuySartorelli (approved):
Good catch. Thanks for contributing this change.

PR #101, opened 2022-10-13
Use X-Status header for status messages

scott1702:
Issue https://github.com/silverstripe/developer-docs/issues/103

HTTP/2 protocol doesn't support status messages so we need to use the X-Status header to display toast notifications correctly


GuySartorelli (commented):
Have you actually tested if the current documented way works? afaik the core code currently does it the documented way and it works.

scott1702:
It doesn't work if using HTTP/2. It does work on regular HTTP though

GuySartorelli:
I seem to have been misremembering - I did a search and couldn't find any usage of `setStatusCode` in core which used the second parameter.

GuySartorelli (approved):

PR #112, opened 2022-11-15
DOC Make 4.12 changelog more explicit

GuySartorelli:
I have seen two separate individuals get confused about the change to entwine, seemingly thinking the `onadd` and `onmatch` event handlers should simply no longer be used because they are in some way broken. Talking to one of these individuals, it became clear that an example would be useful to explain exactly what scenario would be affected and an option for resolving it.

Also fixes some typos and a link.

### DIFF
--- a/en/04_Changelogs/4.12.0.md
+++ b/en/04_Changelogs/4.12.0.md
@@ -58,7 +58,67 @@ jQuery has made some breaking behavioural changes over time. If you are experien
 
 The only change to behaviour we've had to make to accommodate jQuery's breaking changes was with entwine's `onadd` and `onmatch` handlers. jQuery has made substantial changes to its internal functionality, especially in regards to how it adds elements to the DOM, so we had to change with it. Entwine used to change jQuery's internal functionality to fire events synchronously when adding elements to the DOM. It now uses a `MutationObserver` to detect elements being added to the DOM, which means the events are fired asynchronously.
 
-This should only affect your project in the unlikely scenario that you are adding elements to the DOM, then immediately performing some action that relies on something having happened in the `onadd` or `onmatch` handler for those elements. Across all of the core and supported modules we only found one instance of this occurring.
+This should only affect your project in the unlikely scenario that you are adding elements to the DOM, then immediately performing some action that relies on something having happened in the `onadd` or `onmatch` handler for those elements. Across all of the core and supported modules we only found one instance of this occurring. If you do have a scenario like this, you could defer the action like so:
+
+**Before:**
+```js
+$('.some-class').entwine({
+  // Initialise something in the onmatch or onadd handler
+  onmatch: function() {
+    this.data('some-data', 'some-value');
+  }
+});
+$('.some-other-class').entwine({
+  // Add an instance of the above element into the DOM and then try to use the initialised data
+  onclick: function() {
+    this.parent(parentSelector).append('<div class="some-class"></div>');
+    // Expects the data to have been set in onmatch instantly because the event used to be synchronous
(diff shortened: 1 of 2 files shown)


en/04_Changelogs/4.12.0.md
```diff
@@ -58,7 +58,67 @@ jQuery has made some breaking behavioural changes over time. If you are experien
...
+      newElement.setDeferredInitilisationCallback(function() {
+        const initiliasedData = newElement.data('some-data');
+        console.log(initiliasedData); // this will print "some-data" to the console, because it will be called after initialisation has finished
+      });
+    }
+  }
+});
+```
+
+Note that if you are not trying to used the initialised data/functionality _instantly_ after adding the element to the DOM, this kind of change is unnecessary (e.g. click event handlers which rely on _the element the user is clicking_ having been initialised do not need to defer their actions like this, because the user is extremely unlikely to click the element in the time it takes for the `MutationObserver` to fire its events). It is very much exclusive to scenarios where code which adds or replaces elements in the DOM then immediately expects some initialisation to have occurred.
```
michalkleiner:
```suggestion
Note that if you are not trying to use the initialised data/functionality _immediately_ after adding the element to the DOM, this kind of change is not necessary (i.e. click event handlers which rely on _the element the user is clicking_ having been initialised do not need to defer their actions like this, because the user is extremely unlikely to click the element in the time it takes for the `MutationObserver` to fire its events). It is very much exclusive to scenarios where code which adds or replaces elements in the DOM then immediately expects some initialisation to have occurred.
```

en/04_Changelogs/beta/4.12.0-beta1.md
```diff
@@ -133,7 +133,67 @@ jQuery has made some breaking behavioural changes over time. If you are experien
...
+      newElement.setDeferredInitilisationCallback(function() {
+        const initiliasedData = newElement.data('some-data');
+        console.log(initiliasedData); // this will print "some-data" to the console, because it will be called after initialisation has finished
+      });
+    }
+  }
+});
+```
+
+Note that if you are not trying to used the initialised data/functionality _instantly_ after adding the element to the DOM, this kind of change is unnecessary (e.g. click event handlers which rely on _the element the user is clicking_ having been initialised do not need to defer their actions like this, because the user is extremely unlikely to click the element in the time it takes for the `MutationObserver` to fire its events). It is very much exclusive to scenarios where code which adds or replaces elements in the DOM then immediately expects some initialisation to have occurred.
```
michalkleiner:
```suggestion
Note that if you are not trying to use the initialised data/functionality _immediately_ after adding the element to the DOM, this kind of change is not necessary (i.e. click event handlers which rely on _the element the user is clicking_ having been initialised do not need to defer their actions like this, because the user is extremely unlikely to click the element in the time it takes for the `MutationObserver` to fire its events). It is very much exclusive to scenarios where code which adds or replaces elements in the DOM then immediately expects some initialisation to have occurred.
```

michalkleiner (approved):
Minor stuff, otherwise ok.

en/04_Changelogs/4.12.0.md
ChloeSartorelli: `e.g.` here is correct - it's an example, not a restatement/clarification (see https://www.merriam-webster.com/words-at-play/ie-vs-eg-abbreviation-meaning-usage-difference)

michalkleiner: I think it can be either here... that kind of change is not necessary, _for example_ this and that doesn't need to defer, as well as that change is not necessary, _which means_ this and that doesn't need to defer.

All good.

PR #132, opened 2023-01-16
DOC List which classes had ReturnTypeWillChange annotation removed

maxime-rainville:
## Parent issue
- https://github.com/silverstripeltd/product-issues/issues/645


GuySartorelli (approved):
We probably want to be more explicit than this - but that will all be handled under https://github.com/silverstripeltd/product-issues/issues/627 anyway so this is fine for now.

PR #136, opened 2023-01-17
DOC Document changes to trailing slash in URLs

GuySartorelli:
## Parent issue
- https://github.com/silverstripe/silverstripe-cms/issues/2780


en/04_Changelogs/5.0.0.md
```diff
@@ -203,6 +204,47 @@ their aliases (applied by setting a string key for the array) were being ignored
...
+In PHP you can use methods like [Controller::join_links()](api:SilverStripe\Control\Controller::join_links()) or [Controller::normaliseTrailingSlash()](api:SilverStripe\Control\Controller::normaliseTrailingSlash()). In JavaScript in your CMS customisations, we recommend using using the `joinUrlPaths()` utility function from [@silverstripe/webpack-config](https://www.npmjs.com/package/@silverstripe/webpack-config). In your templates, you should use appropriate methods from your controller or model such as [SiteTree::Link()](api:SilverStripe\CMS\Model\SiteTree::Link()) which uses `Controller::join_links()` under the hood.
+[/warning]
+
+The [CanonicalURLMiddleware](api:SilverStripe\Control\Middleware\CanonicalURLMiddleware) will also, by default, redirect traffic to include or omit the trailing slash according to the above configuration. By default, this means that traffic directed to `/about-us/` will be redirected to `/about-us`. You can configure this with the following YML configuration:
+
+```yml
+SilverStripe\Core\Injector\Injector:
+  SilverStripe\Control\Middleware\CanonicalURLMiddleware:
+    properties:
+      forceTrailingSlash: false
```
emteknetnz: I've recommended we rename the property because it's very confusing - https://github.com/silverstripe/silverstripe-framework/pull/10538/files#r1080596063

en/04_Changelogs/5.0.0.md
```diff
@@ -203,6 +204,47 @@ their aliases (applied by setting a string key for the array) were being ignored
...
+
+```yml
+SilverStripe\Control\Controller:
+    add_trailing_slash: true
+```
+
+[warning]
+Because this can be controlled with configuration, it is best practice to avoid explicitly expecting a trailing slash to either be present or be omitted.
+
+In PHP you can use methods like [Controller::join_links()](api:SilverStripe\Control\Controller::join_links()) or [Controller::normaliseTrailingSlash()](api:SilverStripe\Control\Controller::normaliseTrailingSlash()). In JavaScript in your CMS customisations, we recommend using using the `joinUrlPaths()` utility function from [@silverstripe/webpack-config](https://www.npmjs.com/package/@silverstripe/webpack-config). In your templates, you should use appropriate methods from your controller or model such as [SiteTree::Link()](api:SilverStripe\CMS\Model\SiteTree::Link()) which uses `Controller::join_links()` under the hood.
```
emteknetnz: joinUrlPaths() is in admin, not webpack-config https://github.com/silverstripe/silverstripe-admin/pull/1438/files#diff-3f5e793718f99fa848e4676d4d27c8f4a5cf62ca3b6f291168d66facefcd49f5R1

emteknetnz (changes requested):

en/04_Changelogs/5.0.0.md
ChloeSartorelli: Yes. But it cannot be used without webpack-config. Saying it's in admin may lead to people thinking they don't need webpack-config, which is incorrect.

emteknetnz: > `joinUrlPaths()` utility function from [@silverstripe/webpack-config]

Well, it's not FROM webpack-config, so we'll need to change this sentence to something else.  Is webpack-config even relevant? Wouldn't webpack-config just be available for anyone working in the cms since it's in node_modules.  Isn't it just a case of importing something that's always there?

ChloeSartorelli: It is technically exposed as a global, so you could go `ssUrlLib.joinUrlPaths()` - but the way that it's exposed is just an implementation detail of webpack expose-loader and webpack externals, and the name `ssUrlLib` is somewhat arbitrary, so I wouldn't want to encourage people to access it that way.
Instead, we should encourage people to use `webpack-config` when they want to use our exposed libraries or components. I'll change to wording to be clearer.

emteknetnz: All you need is `import { joinUrlPaths } from 'lib/urls';` right?  That's all people need to know in order to use this on their module?  Or do you explicitly need webpack-config as a devDependency?

ChloeSartorelli: `lib/urls` comes from admin, and unless you explicitly set your webpack to grab from admin or tell webpack how to get it as an external, it will not resolve. The supported and recommended way of doing this is to use webpack-config.

ChloeSartorelli: I've made it clear it's part of `silverstripe/admin`, but retained the recommendation of using `@silverstripe/webpack-config`. I've also indicated _how_ to import it, which I hadn't realised I'd omitted.

en/04_Changelogs/5.0.0.md
ChloeSartorelli: done

en/02_Developer_Guides/15_Customising_the_Admin_Interface/02_CMS_Architecture.md
```diff
@@ -76,27 +76,43 @@ SilverStripe\Admin\AdminRootController:
...
+<%-- This is actually a method that can take an argument: --%>
+$AdminURL('more/path/here')
 ```
 
 And in JavaScript, this is available through the `ss` namespace
 
-
 ```js
 ss.config.adminUrl
+// To add to the path you can use joinUrlPaths from @silverstripe/webpack-config
```
emteknetnz:
```suggestion
// You can use this if you use @silverstripe/webpack-config
```

en/04_Changelogs/5.0.0.md
```diff
@@ -13,6 +13,7 @@ title: 5.0.0 (unreleased)
   - [ORM](#api-orm)
 - [Dependency changes](#dependency-changes)
   - [General changes](#dep-general)
+  - [URLs generated by Silverstripe CMS don't include a trailing slash by default](#trailing-slash)
```
emteknetnz: This isn't a "Dependency change", instead it's a functional change

en/04_Changelogs/5.0.0.md
```diff
@@ -203,6 +204,47 @@ their aliases (applied by setting a string key for the array) were being ignored
...
+
+In PHP you can use methods like [Controller::join_links()](api:SilverStripe\Control\Controller::join_links()) or [Controller::normaliseTrailingSlash()](api:SilverStripe\Control\Controller::normaliseTrailingSlash()). In JavaScript in your CMS customisations, we recommend using using the `joinUrlPaths()` utility function `silverstripe/admin` which you can access with `import { joinUrlPaths } from 'lib/urls;` if your project uses [@silverstripe/webpack-config](https://www.npmjs.com/package/@silverstripe/webpack-config). In your templates, you should use appropriate methods from your controller or model such as [SiteTree::Link()](api:SilverStripe\CMS\Model\SiteTree::Link()) which uses `Controller::join_links()` under the hood.
+[/warning]
+
+The [CanonicalURLMiddleware](api:SilverStripe\Control\Middleware\CanonicalURLMiddleware) will also, by default, redirect traffic to include or omit the trailing slash according to the above configuration. By default, this means that traffic directed to `/about-us/` will be redirected to `/about-us`. You can configure this with the following YML configuration:
```
emteknetnz:
```suggestion
The [CanonicalURLMiddleware](api:SilverStripe\Control\Middleware\CanonicalURLMiddleware) will also, by default, redirect traffic to include or omit the trailing slash according to the above configuration. By default, this means that traffic directed to `/about-us/` will be redirected to `/about-us`. You can disable this behavior with the following YML configuration:
```

emteknetnz (changes requested):

en/04_Changelogs/5.0.0.md
ChloeSartorelli: Good point. I'll slap it in the "other breaking changes" section which is also being added in https://github.com/silverstripe/developer-docs/pull/137

en/02_Developer_Guides/15_Customising_the_Admin_Interface/02_CMS_Architecture.md
ChloeSartorelli: done

en/04_Changelogs/5.0.0.md
ChloeSartorelli: done

en/04_Changelogs/5.0.0.md
ChloeSartorelli: done

emteknetnz (approved):

PR #141, opened 2023-01-22
DOC: improve consistency of SilverStripe spelling

freezernick:
As per the docs:

> "SilverStripe" should always appear without a space with both "S"s capitalised.


freezernick:
Ah, just saw a99856af441c94157800aab32d6f836efffe7fcb

Soo, contributing/documentation needs an update 🌚

PR #148, opened 2023-02-02
DOC Clarify that we could upgrade a fixed dependency to add support for a newer PHP release

maxime-rainville:
## Parent issue
- https://github.com/silverstripeltd/product-issues/issues/670


maxime-rainville:
This came up while we were establishing the list of fixed dependencies for CMS5. 

I don't think we need to go all the way and clarify  whether we would due dual support or force a straight upgrade or something else.

emteknetnz (approved):

PR #161, opened 2023-02-16
DOC StaticPublishQueue is now opt-in

emteknetnz:
Issue https://github.com/silverstripeltd/product-issues/issues/676


GuySartorelli (approved):
Wrong heading level but so's the one above it, we'll fix that closer to release.

PR #165, opened 2023-02-21
DOC Document new changes to deprecation output

GuySartorelli:
Documents changes in https://github.com/silverstripe/silverstripe-framework/pull/10697 and https://github.com/lekoala/silverstripe-debugbar/pull/148

This document was using a mix of "deprecation notice" and "deprecation warning" so I've normalised it to always say "deprecation warning" given that we've told monolog to treat them as warnings.

## Do not squash
This has two separate doc commits, which are related to the two commits in https://github.com/silverstripe/silverstripe-framework/pull/10697

## Parent issue
- https://github.com/silverstripe/silverstripe-framework/issues/10608

### DIFF
--- a/en/03_Upgrading/07_Deprecations.md
+++ b/en/03_Upgrading/07_Deprecations.md
@@ -4,27 +4,112 @@ title: Deprecations
 
 # Deprecations
 
-## Updating your codebase to use avoid using deprecated API in Silverstripe CMS
-
 You'll need to go through your codebase to remove references to deprecated APIs and update your project logic.
 
 Major releases of Silverstripe CMS introduce many API changes. They include a deprecation message that will either tell you:
 - to simply use a different yet equivalent API instead, or
 - the API in question has no replacement. This is usually for API that was mostly irrelevant.
 
-To see deprecation warnings in your project and in your project's CI, add `SS_DEPRECATION_ENABLED=true` to your project's `.env` file.
+## Enabling deprecation warnings
 
-Alternatively, add the following line to you project's `app/_config.php`.
-```php
-Deprecation::enable()
+To enable deprecation warnings, set the `SS_DEPRECATION_ENABLED` environment variable in your project's `.env` file.
+
+[info]
+If the `SS_DEPRECATION_ENABLED` environment variable is set, this takes precedence over use of the `Deprecation::enable()` static method.
+[info]
+
+```env
+SS_DEPRECATION_ENABLED=true
 ```
 
-Deprecation notices will only ever show if your `SS_ENVIRONMENT_TYPE` is set to `dev`.
+Alternatively, add the following line to your project's `app/_config.php`.
 
-Your project will now start emitting deprecation warnings on the frontend if you are calling deprecated code, along with the different API you should now use instead, if a replacement exists.
+```php
+Deprecation::enable();
+```
 
-Not all API that gets deprecated will have an equivalent replacement API in that same major version; some of the API is only available from the next major release.  A good example of this is the upgrade for what powers the `SilverStripe\Control\Email\Email` class from `swiftmailer` in CMS 4 to `symfony/mailer` in CMS 5. In these cases, you'll need to upgrade to the new major version before you can access the replacement API.
+[info]
+Deprecation warnings will only ever show if your `SS_ENVIRONMENT_TYPE` is set to `dev`.
+[/info]
 
-Your project will emit deprecation notices for these, although there may be nothing you can do to stop calling that code until you upgrade to the new major version. In this case you should use the deprecation notice to construct a list of what you may need to change after upgrading.
+Once you have resolved all of the deprecation warnings you can, it is recommended to turn off deprecation warnings again.
 
-Once you have resolved all of the deprecation notices you can, it is recommended to turn off deprecation notices again.
+Not all API that gets deprecated will have an equivalent replacement API in that same major version; some of the API is only available from the next major release. A good example of this is the upgrade for what powers the `SilverStripe\Control\Email\Email` class from `swiftmailer` in CMS 4 to `symfony/mailer` in CMS 5. In these cases, you'll need to upgrade to the new major version before you can access the replacement API.
+
+Some code that has been deprecated with no immediate replacement will not emit deprecation notices by default. If you wish to also see notices for deprecated code with no immediate replacement, add the following line to you project's `app/_config.php` file. Note that this will also emit deprecation notices for usages of the deprecated code inside core modules.
+
+```php
+Deprecation::enable(true);
+```
+
+## How to view deprecation warnings
+
+By default, deprecation warnings will be emitted to the error logger, and will be output at the end of CLI responses. They will not be included in HTTP responses by default.
+
+### Viewing deprecation warnings in the logs
+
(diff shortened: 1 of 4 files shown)


en/03_Upgrading/07_Deprecations.md
```diff
@@ -4,27 +4,49 @@ title: Deprecations
...
+      - "../silverstripe.log" # an absolute path, or a path relative to the index.php file (usually inside the public/ directory)
+      - "warning" # warning is the level deprecation warnings are logged as
+  Psr\Log\LoggerInterface.errorhandler:
+    calls:
+      ErrorLogFileHandler: [ pushHandler, [ '%$ErrorLogFileHandler' ] ]
+```
+
+See [Configuring error logging](/developer_guides/debugging/error_handling/#configuring-error-logging) to learn about other ways you can handle error logs.
+
+Additionally, if you want to view the warnings in your browser, you can install [lekoala/silverstripe-debugbar](https://github.com/lekoala/silverstripe-debugbar) as a dev dependency. Deprecation warnings will be logged in the "messages" tab of the debugbar. Be aware however that it will _not_ show you any deprecation warnings that are triggered from XHR requests or which are triggered after the middleware has finished generating the debugbar for the response.
```
emteknetnz: Given this it's only showing a subset of deprecation warnings, it seems like using debugbar here is actually a liability and not something we should encourage anyone to do.  It will give people doing upgrades a false sense of security that they fixed everything, when they haven't.

We're better off simply saying "you have to look at the log file which is the single source of truth" rather than giving only half working implementations.

emteknetnz (changes requested):

en/03_Upgrading/07_Deprecations.md
ChloeSartorelli: @maxime-rainville thoughts?

maxime-rainville: I would create a sub header that says "Showing deprecation warning in your browser". I would put the "Be aware however ..." bit in a warning block to minimise the risk that people miss it.

ChloeSartorelli: Done - the warning is in a warning block.

en/03_Upgrading/07_Deprecations.md
```diff
@@ -4,27 +4,108 @@ title: Deprecations
...
+### Viewing deprecation warnings in the logs
+
+Deprecation warnings are output to the same error logger as all other warnings and errors. You will need to make sure you have a logging handler attached to the default `Psr\Log\LoggerInterface` or `Psr\Log\LoggerInterface.errorhandler` singletons. For example, to log to a file you can add this to your yml configuration:
+
+```yml
+SilverStripe\Core\Injector\Injector:
+  ErrorLogFileHandler:
+    class: Monolog\Handler\StreamHandler
+    constructor:
+      - "../silverstripe.log" # an absolute path, or a path relative to the index.php file (usually inside the public/ directory)
```
emteknetnz, line 58:
```suggestion
      - "./silverstripe.log" # an absolute path, or a path relative to the index.php file (usually inside the public/ directory)
```
Otherwise you'll get a `PHP Fatal error:  Uncaught UnexpectedValueException: The stream or file "/var/www/../silverstripe.log" could not be opened in append mode: Failed to open stream: Permission denied in /var/www/vendor/monolog/monolog/src/Monolog/Handler/StreamHandler.php:115` if you copy paste this config

en/03_Upgrading/07_Deprecations.md
```diff
@@ -4,27 +4,108 @@ title: Deprecations
...
+Deprecation warnings are output for CLI responses by default (assuming they're enabled in general). The warnings are always output at the _end_ of the request, so you don't have to go looking through the output for them. You might want to disable outputting these warnings in CLI responses if, for example, you need to validate the output via code and don't want to add special cases for deprecation warnings.
+
+You can suppress deprecation warnings from CLI output by setting `SS_DEPRECATION_SHOW_CLI` to a falsy value in your .env file.
+
+[info]
+If the `SS_DEPRECATION_SHOW_CLI` environment variable is set, this takes precedence over use of the `Deprecation::SS_DEPRECATION_SHOW_CLI()` static method.
+[info]
+
+```env
+SS_DEPRECATION_SHOW_CLI=0
```
emteknetnz: use either false here or 0 above for SS_DEPRECATION_SHOW_HTTP so they're consistent

emteknetnz (changes requested):

en/03_Upgrading/07_Deprecations.md
ChloeSartorelli: Permissions are the responsibility of the developer. I copied this from the error log section of the docs directly.

emteknetnz: Yes permissions are the developers responsibility, however these instructions are just flat out bad?

If I copy paste the above with `../silverstripe.log` into `app/_config/mysite.yml` and dev/build flush=1 - I get a PHP Fatal error

It's trying to create the log file in a directory ABOVE where my project is.  The code comment `or a path relative to the index.php file (usually inside the public/ directory)` is just wrong? It's trying to create it relative to the project root.  Seems like the intention was for it to create the log file in the project root, but somewhere along the lines something changed

If I instead use `./silverstripe.log` things work fine

emteknetnz (changes requested):

en/03_Upgrading/07_Deprecations.md
ChloeSartorelli: > I get a PHP Fatal error

Because you haven't set the correct permissions.
I'll add a note about permissions in a `[notice]` block. But adding the log to `public/silverstripe.log` is an objectively bad instruction, so I'm not going to do that.

ChloeSartorelli: > The code comment or a path relative to the index.php file (usually inside the public/ directory) is just wrong? It's trying to create it relative to the project root.

I'll validate that - if you're right, I'll update the docs in both places about this and use the correct path.

ChloeSartorelli: I've just tested this. `./silverstripe.log` does indeed use the file in `public/silverstripe.log`, so the current instructions are correct. I'll add a note about file permissions.

ChloeSartorelli: done

en/03_Upgrading/07_Deprecations.md
ChloeSartorelli: done - opted for `true` and `false` as that seems more intuitive.

en/03_Upgrading/07_Deprecations.md
emteknetnz: Replicated issue locally

Setup
- Copy paste yml instructions above to app/_config/mylogging.yml
- Construct a deprecated class in PageController::init() to trigger error
- Call Deprecation::enable() in _config.php

When running dev/build flush=1 from CLI:
- ./silverstripe.log will be created in root dir
- ../silverstripe.log will say permission error, trys to create /var/www/../silverstripe.log

When requesting page via browser
- ./silverstripe.log will be created in public dir
- ../silverstripe.log will be created in root dir

ChloeSartorelli: I've added acceptance criteria to https://github.com/silverstripe/silverstripe-framework/issues/10551 and bumped its impact level. Whatever docs changes are needed re: file paths for logging will be done as a part of that issue.

emteknetnz (approved):

emteknetnz:
@GuySartorelli Merge conflicts

GuySartorelli:
Rebased and resolved conflicts

emteknetnz (approved):

PR #166, opened 2023-02-21
DOC DBComposite dynamic data

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10684

### DIFF
--- a/en/04_Changelogs/5.0.0.md
+++ b/en/04_Changelogs/5.0.0.md
@@ -35,6 +35,7 @@ This changelog provides a list of changes between Silverstripe CMS 4 and 5.
   - [General changes](#api-general)
   - [Email](#api-email)
   - [ORM](#api-orm)
+  - [Dynamic properties](#dynamic-properties)
   - [Removed and changed API (by module, alphabetically)](#api-removed-and-changed)
 
 ## Change to supported modules {#supported-modules}
@@ -490,6 +491,7 @@ This is a major release and contains many breaking API changes. Deprecation warn
   - If your `composer.json` file already does not have an `extra.resources-dir` key and you want to keep your resources in the `resources` directory, you can set `extra.resources-dir` to `resources`.
   - If your `composer.json` file already does not have an `extra.resources-dir` key and you want to use the new default `_resources` directory, you may need to check that your code and templates don't assume the directory name for those resources. In your templates it is preferred to [use `$resourePath()` or `$resourceURL()`](/developer_guides/templates/requirements/#direct-resource-urls) to get paths for resources.
 - Removed the `HTMLValue` injection "shorthand", use the fully qualified [`HTMLValue`](api:SilverStripe\View\Parsers\HTMLValue) instead.
+- In `silverstripe/staticpublishqueue` the class [`SilverStripe\StaticPublishQueue\Dev\StaticPublisherState`](api:SilverStripe\StaticPublishQueue\Dev\StaticPublisherState) is no longer enabled by default and can be enabled via opt-in. There are opt-in instructions in the `README.md` of the module.
 
 ### Email {#api-email}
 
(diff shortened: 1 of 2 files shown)


GuySartorelli (changes requested):
Please also (and arguably more importantly) update the _main_ changelog.
Please include a larger section than just this, since it doesn't only affect `DBComposite` - it potentially affects any subclass of `ViewableData` in someone's project or module which overriddes any of these methods:
- `__get()`
- `__set()`
- `hasField()`
- `setField()`
- `getField()`

This probably warrants a full heading about dynamic properties and what we've changed to keep things working and what people may need to do in their own projects if they've overridden any of the above.

emteknetnz:
Main changelog being 5.0.0.md? I've just put it in the 5.0.0-rc with the intention that we'll copy over whatever's in the RC changelog to the main changelog in one go once we do the stable

I'm not sure this warrants much more detail in the changelog? All we've changed is DBComposite::setField()? We've added some new API to ViewableData, though that's really there as a "it's there if you really need it" option rather than something you should be using. Ideally people would just not use dynamic properties at all.

In terms of the overloading - these comments "This should be overloaded in child classes." on ViewableData:get/set/hasField() - they're already there in CMS 4, not something new we've done here

GuySartorelli:
> Main changelog being 5.0.0.md?

Yes. We don't tend to copy things from beta/rc down up the the stable changelog, we usually copy from the stable changelog down to those. We already are going to need to copy stuff from the stable changelog to the RC so it'd be weird to now have to copy in _both_ directions during the release, especially since we don't usually do that.

> I'm not sure this warrants much more detail in the changelog?

I think it does. Currently, anyone who has overridden any of the above-mentioned methods may be expecting dynamic properties to just keep working, but they may now find that because they don't have _true_ dynamic properties, they run into the exact same problem that prompted this PR in the first place.
The problem was identified for `DBComposite` - but people may need to make changes in their own projects to work with this new dynamic data API and if we don't call that out how will they even know that new API exists?

> that's really there as a "it's there if you really need it" option

Sure. But people need to know it's there to use.

> these comments "This should be overloaded in child classes." on ViewableData:get/set/hasField() - they're already there in CMS 4, not something new we've done here

I know. But that just means we need to be able to support people who _have_ overloaded them, and give those developers a clear upgrade path.

michalkleiner:
I would support having a separate section, there might be devs who use this a lot.
Looks like CMS 5 upgrades from 4 are starting to become fun like CMS 4 upgrades from 3.

emteknetnz:
Have added section to 5.0.0 - removed mentioned of DBComposite::setField() specifically since that's basically wrapped in the larger section about dynamic properties

I also moved an unrelated thing about staticpublishqueue that was in 5.0.0-rc1.md to 5.0.0.md

en/04_Changelogs/5.0.0.md
```diff
@@ -35,6 +35,7 @@ This changelog provides a list of changes between Silverstripe CMS 4 and 5.
   - [General changes](#api-general)
   - [Email](#api-email)
   - [ORM](#api-orm)
+  - [Dynamic properities](#dynamic-properties)
```
ChloeSartorelli:
```suggestion
  - [Dynamic properties](#dynamic-properties)
```

en/04_Changelogs/5.0.0.md
```diff
@@ -523,6 +525,13 @@ This is a major release and contains many breaking API changes. Deprecation warn
   - In Silverstripe CMS 4, calling [`SQLSelect::setLimit()`](api:SilverStripe\ORM\Queries\SQLSelect::setLimit()) with `0` as argument would unset the limit. In Silverstripe CMS 5, it sets a limit of `0` causing the query to return no results. Call `setLimit()` with `null` to unset the limit.
 - [`ArrayList::limit()`](api:SilverStripe\ORM\ArrayList::limit()) in Silverstripe CMS 5 throws an `InvalidArgumentException` when called with a negative `$length` or `$offset`.
 
+### Dynamic properites {#dynamic-properties}
```
ChloeSartorelli:
```suggestion
### Dynamic properties
```
No need to add an explicit anchor if it's just the same as the default

en/04_Changelogs/5.0.0.md
```diff
@@ -523,6 +525,13 @@ This is a major release and contains many breaking API changes. Deprecation warn
...
+
+- PHP Dynamic properities were deprecated in PHP 8.2 to minimize the confusion caused by having API that was not explicitly declared.
```
ChloeSartorelli:
```suggestion
- PHP Dynamic properties were deprecated in PHP 8.2 to minimise the confusion caused by having API that was not explicitly declared.
```
"minimize" with a z is american spelling - though I won't hold up the PR just for this if you choose not to change that.

en/04_Changelogs/5.0.0.md
```diff
@@ -523,6 +525,13 @@ This is a major release and contains many breaking API changes. Deprecation warn
...
+- To support backwards compatibility with project and module code, the magic methods `__get()` and `__set()` on [`SilverStripe\View\ViewableData`](api:SilverStripe\View\ViewableData) will simulate dynamic properties by writing values to an array stored as a private property.
+- This functionality may not work if subclasses override any of the following methods: `__get()`, `__set()`, `getField()`, `setField()`, `hasField()`.
+- If these method overrides break this functionality, the private property array that stores the values can be accessed with via the following methods: [`SilverStripe\View\ViewableData::getDynamicData()`](api:SilverStripe\View\ViewableData::getDynamicData()), [`SilverStripe\View\ViewableData::setDynamicData()`](api:SilverStripe\View\ViewableData::setDynamicData()), [`SilverStripe\View\ViewableData::hasDynamicData()`](api:SilverStripe\View\ViewableData::hasDynamicData())
```
ChloeSartorelli:
```suggestion
- If these method overrides don't call the parent methods, they may need access to the dynamic data. The private property array that stores the values can be accessed with via the following methods: [`SilverStripe\View\ViewableData::getDynamicData()`](api:SilverStripe\View\ViewableData::getDynamicData()), [`SilverStripe\View\ViewableData::setDynamicData()`](api:SilverStripe\View\ViewableData::setDynamicData()), [`SilverStripe\View\ViewableData::hasDynamicData()`](api:SilverStripe\View\ViewableData::hasDynamicData())
```

GuySartorelli (changes requested):

en/04_Changelogs/5.0.0.md
emteknetnz: Updated

en/04_Changelogs/5.0.0.md
emteknetnz: Updated

en/04_Changelogs/5.0.0.md
emteknetnz: Updated

en/04_Changelogs/5.0.0.md
emteknetnz: Updated

GuySartorelli (approved):

PR #158, opened 2023-02-09
DOC Rearrange 5.0.0-beta1 changelog

emteknetnz:
Issue https://github.com/silverstripeltd/product-issues/issues/679

PR #171, opened 2023-02-24
DOC Document some behaviours that aren't supported anymore

GuySartorelli:
I've got three commits here. The first two aren't technically related to the issue at hand but I found them while doing this documentation and they're super simple so shouldn't add any hassle to the reviewer.

Commit 1: Remove 4.x alpha changelogs. The beta, rc, and stable ones were already removed - we just missed
these ones.
Commit 2: Move "templates" section to where it should be. This was missed when rearranging the changelog in https://github.com/silverstripe/developer-docs/pull/158
Commit 3: Document the behaviours which are deprecated in https://github.com/silverstripe/silverstripe-framework/pull/10702 which weren't already documented in the changelog

## Parent issue
- https://github.com/silverstripe/silverstripe-framework/issues/10692


en/04_Changelogs/5.0.0.md
```diff
@@ -36,6 +36,7 @@ This changelog provides a list of changes between Silverstripe CMS 4 and 5.
   - [Email](#api-email)
   - [ORM](#api-orm)
   - [Dynamic properties](#dynamic-properties)
+  - [Templates](#templates)
```
michalkleiner, line 39: Someone said somewhere we don't need the specific anchor when it's the same as the one produced from the title. Though I actually prefer having it even when it's the same, for consistency, and for the cases when the title extends, we don't need to change it in several places.

en/04_Changelogs/5.0.0.md
```diff
@@ -532,6 +523,17 @@ This is a major release and contains many breaking API changes. Deprecation warn
 - To support backwards compatibility with project and module code, the magic methods `__get()` and `__set()` on [`SilverStripe\View\ViewableData`](api:SilverStripe\View\ViewableData) will simulate dynamic properties by writing values to an array stored as a private property.
 - This functionality may not work if subclasses override any of the following methods: `__get()`, `__set()`, `getField()`, `setField()`, `hasField()`.
 - If these method overrides don't call the parent methods, they may need access to the dynamic data. The private property array that stores the values can be accessed with via the following methods: [`SilverStripe\View\ViewableData::getDynamicData()`](api:SilverStripe\View\ViewableData::getDynamicData()), [`SilverStripe\View\ViewableData::setDynamicData()`](api:SilverStripe\View\ViewableData::setDynamicData()), [`SilverStripe\View\ViewableData::hasDynamicData()`](api:SilverStripe\View\ViewableData::hasDynamicData())
+### Templates
```
michalkleiner, line 527:
```suggestion

### Templates
```

en/04_Changelogs/5.0.0.md
ChloeSartorelli: Is this meant as a comment on the _heading_ (where it's not needed but we could have a separate card to discuss what our standard should be for that - maybe on the linting docs RFC)? It's always needed for the link, because you have to tell it where to link to.

michalkleiner: Ah, right, makes sense, I mixed those two together. All good.

en/04_Changelogs/5.0.0.md
ChloeSartorelli: Oops! Good catch.
Done.

emteknetnz (approved):

PR #175, opened 2023-03-07
DOC Add upgrade guide for tractorcow/silverstripe-fluent

GuySartorelli:
Description from parent issue (since it's private):
> Acceptance criteria
> - The Guide should explain how to upgrade from Fluent v4 to Fluent v6/v7.
> - The Guide should explain how to upgrade from Fluent v5 to Fluent v6/v7.
>
> Notes
>
> - If you are on Fluent v4, we are not sure if the best course of action is to skip 5 and go straight to 6. The person who picks up this card should spend some time analysing what the best approach is on in this case and draft the guide accordingly.

I've opted to make a single upgrade guide that doesn't make assumptions about what version you're upgrading _from_ nor _to_ - though it does state the idea is to adapt your code to be compatible with the _latest version_.
This means we don't have to duplicate instructions or advice for scenarios where v4 and v5 behave the same way, and both need the same actions taken to upgrade to v6.

There is only _one_ additional consideration for upgrading to v7, which I'll cover in a PR for the CMS 5 changelog. The change is just that `FluentExtension::getLinkingMode()` and `FluentExtension::LocaleLink()` are removed in v7. The upgrade guide already notes that they're deprecated and what to use instead, so I think it's fine to not mention v7 here.
I'm explicitly avoiding mention of v7 as that's a CMS 5 consideration, and this upgrade guide is for CMS 4.

## After merging
After merging this PR, immediately merge it up to `4` and `5.0`. On merge-up to `5.0`, delete it, as it isn't applicable for CMS 5 projects (this is the same treatment given to the graphql v3->v4 and phpunit 5=>9 upgrade guides)

## Parent issue 
- https://github.com/silverstripeltd/product-issues/issues/662


GuySartorelli:
@mfendeksilverstripe can you please give this a once-over and let me know if I've missed anything or got anything wrong?
CC also @chrispenny and @tractorcow

en/03_Upgrading/07_Upgrading_Fluent.md
```diff
@@ -0,0 +1,306 @@
...
+
+## Fluent state
+
+Fluent state is the mechanism responsible for maintaining the global current locale state. Any time you swap between locales, the fluent state handles this change.
+
+In fluent 5 the way this state affects ORM actions changed.
+
+|              | **Fluent 4** | **Fluent 5** | **Fluent 6** |
+|--------------|--------------|--------------|--------------|
+| Fluent state | Local        | Global       | Global       |
```
chrispenny: I'd just like to double check: Was it version 5 where it switched to `Global`? I remembered it being version 6 (but I could certainly be wrong).

en/03_Upgrading/07_Upgrading_Fluent.md
```diff
@@ -0,0 +1,306 @@
...
+```php
+$titles = FluentState::singleton()->withState(function (FluentState $state): array {
+    $state->setLocale('my_locale_code');
+
+    // This returns results for the 'my_locale_code' locale.
+    return MyModel::get()->column('Title');
+});
+// If you returned the DataList above, here it would give results for the locale in the global scope, not for 'my_locale_code'.
+```
+
```
chrispenny: An example of how writing would work might also be handy.

#### Fluent v4 {#fluent-state-example-v4}

```php
// Set your current Locale state to locale_one before fetching the model
FluentState::singleton()->setLocale('locale_one');

// Fetching the model here will persist the Locale in the local state of the model
$model = MyModel::get()->byID(1);
$model->Title = 'Updated Title';

// Switching Locales here changes the global state, but not the local state of the model
FluentState::singleton()->setLocale('locale_two');

$model->write();
// The result is that "Updated Title" is saved to 'locale_one'
```

#### Fluent v5 and v6 {#fluent-state-example-v5-v6}

```php
// Set your current Locale state to locale_one before fetching the model
FluentState::singleton()->setLocale('locale_one');

// Fetching the model here sets the Locale to the local state of the model, but it is not persisted
$model = MyModel::get()->byID(1);
$model->Title = 'Updated Title';

// Switching Locales here changes both the global state, and the local state of the model
FluentState::singleton()->setLocale('locale_two');

$model->write();
// The result is that "Updated Title" is saved to 'locale_two'
```

en/03_Upgrading/07_Upgrading_Fluent.md
mfendeksilverstripe: I think it was `5` as Fluent `6` was mostly focused on versioned history stuff.

mfendeksilverstripe (approved):
Very nice guide 👍

en/03_Upgrading/07_Upgrading_Fluent.md
ChloeSartorelli: _would_ you temporarily set and revert the locale like that? It seems to me like that should be done inside `withState()` as well?
The guide explicitly tells people to avoid this pattern, as I assumed it was an antipattern.

en/03_Upgrading/07_Upgrading_Fluent.md
ChloeSartorelli: I took this from a guide Mojmir created originally, but I decided to run the code examples to validate the results.

In fluent 6.x-dev with global state locale "en_US", with no explicit fallbacks defined and default config:
```php
$result = FluentState::singleton()->withState(function (FluentState $state) {
    $state->setLocale('en_NZ');
    return [
        'titles' => Page::get()->column('Title'),
        'list' => Page::get(),
    ];
});

var_dump([
    $result['titles'], // includes titles from the en_NZ locale
    $result['list']->column('Title'), // includes titles from the en_NZ locale
    Page::get()->column('Title') // includes titles from the en_US locale
]);
```

Did I do something wrong there? I would have expected only the top result to use `en_NZ` and the bottom two to use `en_US`, right?

en/03_Upgrading/07_Upgrading_Fluent.md
chrispenny: Whether it's wrapped in `withState()` or not, the point I'm trying to get across is simply that the behaviour of the `write()` has changed.

`write()` behaviour was a significant change for some of our early projects, especially since the "patterns" that we now have were yet to be discovered/decided.

en/03_Upgrading/07_Upgrading_Fluent.md
chrispenny: Interesting... Yes, I think I would expect the bottom two to use `en_US` if we're saying that there is no more local state.

ChloeSartorelli: @mfendeksilverstripe do you have any insight here?

en/03_Upgrading/07_Upgrading_Fluent.md
ChloeSartorelli: Yup, that's a fair point. I'll add an example of writing once we've sorted out what has actually changed in https://github.com/silverstripe/developer-docs/pull/175#discussion_r1128327995

en/03_Upgrading/07_Upgrading_Fluent.md
mfendeksilverstripe: This is a bit unexpected - maybe Fluent 6 had some changes that I'm not aware of. I remember that Damian was the one who suggested to split some changes off as Fluent 6 from 5 and made the separation. Maybe there were some other changes added into the package along the way.

I think the way forward is to write unit tests for all these cases and document the behaviour.

ChloeSartorelli: If that's the case, are we effectively saying "Whatever the behaviour currently is, that's the correct behaviour"? I'm happy to write those unit tests if that's the case - but I don't have enough background with fluent if the intention is to find out what the behaviour is _and then make decisions about what is expected and what's a bug_

mfendeksilverstripe: I think we need to regroup on that once that once the unit tests are in place but my general recommendation would be to go with `Global` approach as it's a safer and more consistent. I recognise that this may be debatable but preserving local state is additional complexity.

ChloeSartorelli: It sounds like there's uncertainty as to what the expected behaviour is around this section - I'm going to remove it from the upgrade guide for now (as it appears the behaviour hasn't changed).

I'll open a separate card in `tractorcow/silverstripe-fluent` for working out
a) what the current behaviour _is_,
b) what it _should_ be,
c) if the answers to those questions warrant any changes to this guide.

mfendeksilverstripe: Sounds good, thanks 👍

ChloeSartorelli: https://github.com/tractorcow-farm/silverstripe-fluent/issues/787

ChloeSartorelli: Removed this section.

en/03_Upgrading/07_Upgrading_Fluent.md
ChloeSartorelli: Removed this section as per https://github.com/silverstripe/developer-docs/pull/175#discussion_r1128717103

en/03_Upgrading/07_Upgrading_Fluent.md
tractorcow: >    $result['list']->column('Title'), // includes titles from the en_NZ locale

The query state was set to `Fluent.locale = en_NZ` when the query was created (i.e. inside the callback). This state persists after the callback is returned. Read queries are always sticky to the locale when they were created, not when they are executed.

tractorcow: > Interesting... Yes, I think I would expect the bottom two to use en_US if we're saying that there is no more local state.

This should apply to fluent 7 on silverstripe 5; I'll leave this to you to figure out as I'm not using ss 5 :)

tractorcow: Still local on fluent 6.

https://github.com/tractorcow-farm/silverstripe-fluent/blob/6/src/Extension/Traits/FluentObjectTrait.php#L75-L83

tractorcow: FYI write queries are NOT sticky, and are always in the global locale, and always have been. :)

en/03_Upgrading/07_Upgrading_Fluent.md
```diff
@@ -0,0 +1,252 @@
+---
+title: Upgrading Fluent
+summary: Upgrade your Silverstripe CMS project to use the latest version of tractorcow/silverstripe-fluent
+---
+
+Silverstripe only commercially supports version 4.x of [tractorcow/silverstripe-fluent](https://github.com/tractorcow-farm/silverstripe-fluent) with the 4.x major release line of Silverstripe CMS - however, the fluent module also has versions 5.x and 6.x which are compatible with Silverstripe CMS 4.x.
+
+This documents some of the breaking and functional changes across these three major release lines of the fluent module, as well as some some key things to look out for and general guidelines on how to adapt your code to be compatible with the latest version.
```
emteknetnz:
```suggestion
This documents some of the breaking and functional changes across these three major release lines of the fluent module, as well as some key things to look out for and general guidelines on how to adapt your code to be compatible with the latest version.
```

en/03_Upgrading/07_Upgrading_Fluent.md
```diff
@@ -0,0 +1,252 @@
...
+|---------------------------|--------------|--------------|--------------|
+| Locale permission support | No           | Yes          | Yes          |
+
+[hint]
+If you're upgrading from Fluent v4, review your requirements for user permissions in the context of locales and update the permissions as needed. For example you may have content authors who should only be allowed to edit content for specific locales.
+[/hint]
+
+### Enhanced CMS UI
+
+Fluent 5 introduced some new configurable UI actions, including a new `Localisation` tab in the edit form for localised models where you see and manage the localisation status of each record. The `Localisation` tab is referred in documentation as the "localisation manager".
```
emteknetnz:
```suggestion
Fluent 5 introduced some new configurable UI actions, including a new `Localisation` tab in the edit form for localised models where you see and manage the localisation status of each record. The `Localisation` tab is referred to in documentation as the "localisation manager".
```

en/03_Upgrading/07_Upgrading_Fluent.md
```diff
@@ -0,0 +1,252 @@
...
+
+|               | `localise_actions_enabled` | `batch_actions_enabled` | `copy_from_locale_enabled` | `copy_to_locale_enabled` |
+|---------------|--------|--------|--------|--------|
+| Default value | `true` | `true` | `true` | `true` |
+| Functionality | Enable localise actions for records which are not available in the current locale (copy to draft, copy & publish and Localise actions) | Enable "Localisation" actions in the edit form. More about this below. | Enable "copy from locale" action in the localisation manager | Enable "copy to locale" action in the localisation manager |
+
+As mentioned in the table above, the `batch_actions_enabled` configuration value controls whether a new "Localisation" action group appears in the edit form. This is not to be confused with "batch actions" menu in the site tree.
```
emteknetnz:
```suggestion
As mentioned in the table above, the `batch_actions_enabled` configuration value controls whether a new "Localisation" action group appears in the edit form. This is not to be confused with the "batch actions" menu in the site tree.
```

emteknetnz (changes requested):

en/03_Upgrading/07_Upgrading_Fluent.md
ChloeSartorelli: Done

en/03_Upgrading/07_Upgrading_Fluent.md
ChloeSartorelli: Done

en/03_Upgrading/07_Upgrading_Fluent.md
ChloeSartorelli: Done

emteknetnz (approved):

PR #178, opened 2023-03-08
DOC Clarify when we will consider changing a module name or namespace

maxime-rainville:
Slight tweak to major release policy to clarify when we consider OK to chnage a module name or namespace.

## Parent issue
- https://github.com/silverstripe/developer-docs/issues/179

### DIFF
--- a/en/06_Project_Governance/05_Major_release_policy.md
+++ b/en/06_Project_Governance/05_Major_release_policy.md
@@ -138,7 +138,7 @@ Commercially supported modules are reviewed with each new major release. At laun
 
 New modules may be added to the commercially supported modules list during a Silverstripe CMS major release line's lifespan. Modules are not withdrawn from the commercially supported modules list for a Silverstripe CMS major releases line unless external factors make it impractical to continue support. e.g. The module requires a discontinued third party service to work as intended.
 
-Modules that are supported for one major release line may not be supported for follow up release lines. Supported modules can be rename or their namespaces can be changed when a new Silverstripe CMS major release is published.
+Modules that are supported for one major release line may not be supported for follow-up release lines. Supported modules can be renamed or their namespaces can be changed when a new Silverstripe CMS major release is published.
 
 ## Expected Silverstripe CMS major release timeline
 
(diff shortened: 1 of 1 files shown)


en/06_Project_Governance/05_Major_release_policy.md
```diff
@@ -138,7 +138,7 @@ Commercially supported modules are reviewed with each new major release. At laun
 
 New modules may be added to the commercially supported modules list during a Silverstripe CMS major release line's lifespan. Modules are not withdrawn from the commercially supported modules list for a Silverstripe CMS major releases line unless external factors make it impractical to continue support. e.g. The module requires a discontinued third party service to work as intended.
 
-Modules that are supported for one major release line may not be supported for follow up release lines.
+Modules that are supported for one major release line may not be supported for follow up release lines. Supported modules can be rename or their namespaces can be changed when a new Silverstripe CMS major release is published.
```
michalkleiner:
```suggestion
Modules that are supported for one major release line may not be supported for follow-up release lines. Supported modules can be renamed or their namespaces can be changed when a new Silverstripe CMS major release is published.
```

michalkleiner:
Is it only during CMS major versions? I would expect a module can be renamed/re-namespaced during its own major version change. Do we actually take a module's name as a public API of the CMS itself we need to keep semver commitment for?

maxime-rainville:
@michalkleiner I'm hopeful that we won't be doing too many out-of-sync major releases in CMS5. There were a couple major release of the GraphQL and Fluent modules during the lifetime of CMS4. (Technically, we only supported Fluent 4 officially.)

The shorter lifespan of CMS5 will negate the need for individual module majors. Basically, by the time we might consider doing a substantial architectural change to a module that would require a new major, we'll most likely be close enough to the next CMS major release that we decide it's not worth the trouble to do an out-of-sync major right now.

I think the current policy as written allows us to do a major release of a single module as long as we keep supporting the version that shipped with the new major at launch. For example, if we find that for some reason we need to tag v3 of `silverstripe/assets`, we could decide not to officially support v3 and to keep supporting v2 for the lifetime of CMS 5.

It's difficult to imagine a scenario where we would consider doing an individual module major right now. Given this, I'm inclined to just leave this scenario unmentioned in the policy and deal with it if/when it becomes a potential problem.

GuySartorelli:
> Do we actually take a module's name as a public API of the CMS itself we need to keep semver commitment for?

Because we have things like `moduleexists` conditions in configuration, requirements API syntax that uses module names, etc, I think we'd have to have a _really_ good reason to change a module name outside of a major release.

GuySartorelli (approved):

Issue #62, opened 2022-08-09
Update docs from graphql 3 to graphql 4 compatability


GuySartorelli:
There are several sections in the docs outside the explicit graphql section that refer to graphql, e.g. https://docs.silverstripe.org/en/4/developer_guides/model/versioning/#expose-graphql-scaffolding

These currently are written with graphql 3 compatability. They should be updated for graphql 4 compatability.

## Acceptance criteria
- [x] All graphql v3 examples in the CMS 4 docs are updated
    - [x] to ensure they still actually work
    - [x] to _also_ include graphql v4 examples
- [x] All graphql v3 examples in the CMS 5 docs are removed (leaving the graphql v4 examples)
    - [x] There are no mentions of "graphql v4"

## OLD CMS 4 PRs
Once these are all merged, reassign to me so I can update the CMS 5 docs
- https://github.com/silverstripe/developer-docs/pull/184
- https://github.com/silverstripe/silverstripe-admin/pull/1473
- https://github.com/silverstripe/developer-docs/pull/187
- https://github.com/silverstripe/developer-docs/pull/197

## OLD CMS 5 PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1481

## CMS 4 PRs
Once these are all merged, reassign to me so I can update the CMS 5 docs and code if necessary
- https://github.com/silverstripe/developer-docs/pull/211
- https://github.com/silverstripe/developer-docs/pull/215
- https://github.com/silverstripe/silverstripe-admin/pull/1498

## CMS 5 PRs
- https://github.com/silverstripe/developer-docs/pull/196
- https://github.com/silverstripe/developer-docs/pull/247
- https://github.com/silverstripe/silverstripe-admin/pull/1500

rafaeldsousa:
Can confirm the above. 
While following the guide, could see that queries present variables in wrong casing and that the queries do not include namespaces when created, as suggested at the bottom of the docs. 
**E.g**
```
readOneMyVersionedObject(
    Versioning: {
      Mode: LATEST
    },
    ID: $id
  ) {
    ID
    Versions (limit: $limit, offset: $offset) {
      pageInfo {
        totalCount
      }
      edges {
        node {
          Version
          Author {
            FirstName
            Surname
          }
          Publisher {
            FirstName
            Surname
          }
          Published
          LiveVersion
          LatestDraftVersion
          LastEdited
        }
      }
    }
  }
```

Not the case anymore, no part of namespace is included in the queries name.
```
[...]

PR #184, opened 2023-03-13
DOC Update history viewer docs with gql4 example

GuySartorelli:
This documention was just flat out broken. The GQL3 docs had old webpack config, a bunch of unnecessary stuff, and some of the doc just let to completely incorrect behaviour (e.g. `Mode: LATEST` - but we need _all_ of the versions).

I've updated it to be graphql4 focused primarily, and just shown what needs to change whenever something's different for graphql3. This caters better to our recommended approach (i.e. use gql4 unless you have to use 3 for some reason) and will make it a lot easier to pull out the v3 stuff for CMS 5.

Once this is merged, I'll merge it up to the `5.0` docs, and create a PR targetting that to pull out the graphql 3 stuff from it.

## Issue
- https://github.com/silverstripe/developer-docs/issues/62

### DIFF
--- a/en/02_Developer_Guides/00_Model/10_Versioning.md
+++ b/en/02_Developer_Guides/00_Model/10_Versioning.md
@@ -865,6 +865,12 @@ Since Silverstripe CMS 4.3 you can use the React and GraphQL driven history view
 comparisons for a versioned DataObject. This is automatically enabled for SiteTree objects and content blocks in
 [dnadesign/silverstripe-elemental](https://github.com/dnadesign/silverstripe-elemental).
 
+[warning]
+Because of the lack of specificity in the `HistoryViewer.Form_ItemEditForm` scope used when injecting the history viewer to the DOM, only one model can have a working history panel at a time, with exception to `SiteTree` which has its own history viewer scope. For example, if you already have `dnadesign/silverstripe-elemental` installed, the custom history viewer instance injected as a part of this documentation will _break_ the one provided by the elemental module.
+
+There are ways you can get around this limitation. You may wish to put some conditional logic in `app/client/src/boot/index.js` below to only perform the transformations if the current location is within a specific model admin, for example.
+[/warning]
+
 If you want to enable the history viewer for a custom versioned DataObject, you will need to:
 
 * Expose GraphQL scaffolding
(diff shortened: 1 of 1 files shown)


en/02_Developer_Guides/00_Model/10_Versioning.md
```diff
@@ -894,12 +910,19 @@ class MyVersionedObject extends DataObject
...
+packages to be built in order to enable history viewer functionality. This section includes a very basic webpack configuration which uses [@silverstripe/webpack-config](https://www.npmjs.com/package/@silverstripe/webpack-config).
+
+[hint]
+If you have this configured for your project already, ensure you have the `react-apollo` and `graphql-tag` libraries in your `package.json`
+requirements (with the appropriate version constraints from below), and skip this section.
+[/hint]
+
+[notice]
+Using `@silverstripe/webpack-config` will keep your transpiled bundle size smaller and ensure you are using the correct versions of `react-apollo` and `graphql-tag`, as these will automatically be added as [webpack externals](https://webpack.js.org/configuration/externals/). If you are not using that npm package, it is very important you use the correct versions of those dependencies.
+[/notice]
```
sabina-talipova, line 926: Perhaps we could clarify that the libraries should be installed as `dependencies` somewhere in "hint" or "notice" section. Just in case if the user skips the `package.json` code snippet.

ChloeSartorelli: Ahhh, but they don't have to be - that's the beauty of webpack externals.

en/02_Developer_Guides/00_Model/10_Versioning.md
```diff
@@ -1065,71 +1082,77 @@ For this we need one query and one mutation:
...
-          Published
-          LiveVersion
-          LatestDraftVersion
-          LastEdited
+          deleted
+          draft
+          published
+          liveVersion
+          latestDraftVersion
+          lastEdited
```
sabina-talipova, line 1120: Probably, I missed this point somewhere. Is it the rule to use lower case variables name in GQL4? If yes, could we mention this fact somewhere, please.

sabina-talipova (commented):
Looks good for me. Just small question to clarify.

en/02_Developer_Guides/00_Model/10_Versioning.md
ChloeSartorelli: It's not a rule but it is best practice and it's the default behaviour. See [the main graphql documentation](https://docs.silverstripe.org/en/4/developer_guides/graphql/working_with_dataobjects/adding_dataobjects_to_the_schema/#adding-more-fields) which mentions this.

sabina-talipova (approved):
LGFM

PR #188, opened 2023-03-16
DOC Review documentation for common errors

GuySartorelli:
- Remove unnecessary references to CMS 2, 3, and 4
- Change links to use https
- Update broken links (if I happen to find them)
- Replace "localhost" and "yoursite" with "www.example.com" where appropriate
- Remove outdated "how to do a release" doc (with Max's approval)
- Escape any rogue HTML tags
- Update veracity of `00_Server_Requirements.md` and `02_Composer.md`
    - I should have done these separately to make it easier to review but it'll be too hard to split that out now. Sorry about that.

Any changes to formatting/grammar/spelling were done simply because I noticed it while I was there - please don't recommend additional formatting/grammar/spelling changes for parts of the doc I haven't touched.

## Issue
- https://github.com/silverstripe/developer-docs/issues/164

Issue #164, opened 2023-02-20
CMS 5 Doc review


maxime-rainville:
Our CMS 5 doc should not be telling lies.

## Acceptance criteria
- All CMS 3 and CMS 4 examples are removed/updated for CMS 5 doc
- CMS 5 doc is reviewed with a focus on area with substantial change
    - including validate javascript examples still work
- ~~All GraphQL v3 logic is removed and replaced with GraphQL v4/v5 equivalent.~~ handled separately

## Related card 
 - https://github.com/silverstripe/developer-docs/issues/62

## Out-of-scope
- Lessons
- Not having namespaces
- Userhelp

## PRs
- (CMS4!) https://github.com/silverstripe/developer-docs/pull/190
- https://github.com/silverstripe/developer-docs/pull/188
- https://github.com/silverstripe/developer-docs/pull/189
- https://github.com/silverstripe/developer-docs/pull/198
- https://github.com/silverstripe/developer-docs/pull/206

PR #189, opened 2023-03-20
DOC Update Getting Started and ORM sections

GuySartorelli:
A (relatively quick) sweep over the getting started and ORM sections of the docs to update any false or outdated information.

Note: This should be reviewed _after_ https://github.com/silverstripe/developer-docs/pull/188

## Scope
The primary purpose of this PR is to ensure that:
1. code examples actually work in CMS 5
2. written documentation is factually correct

In many cases I have also augmented the existing documentation to make it clearer or to add information I felt was missing.

There are also cases where I have updated code examples or written documentation to better reflect best practices - but that is _not_ the main purpose of this PR. If there are areas where I have not updated examples or docs to reflect best practices (but those examples work and the docs are factually correct) then they are _out of scope_ for this PR.

Formatting that I haven't explicitly touched is definitely out of scope.

## Issue
- https://github.com/silverstripe/developer-docs/issues/164

### DIFF
--- a/en/00_Getting_Started/03_Environment_Management.md
+++ b/en/00_Getting_Started/03_Environment_Management.md
@@ -35,9 +35,14 @@ not be detected automatically. If you wish to load environment variables from a 
 need to do so manually. See the [Including an extra `.env` file](#including-an-extra-env-file) section below for more
 information.
 
-## Managing environment variables with Apache
+## Other ways to manage environment variables
 
-You can set "real" environment variables using Apache. Please
+Silverstripe CMS will respect environment variables provided by the server.
+
+If you are using a docker compose setup, you can set environment variables in your `docker-compose.yml` file.
+[See the docker compose docs for more information](https://docs.docker.com/compose/compose-file/#environment).
+
+You can set environment variables using Apache. Please
 [see the Apache docs for more information](https://httpd.apache.org/docs/current/env.html).
 
 ## How to access the environment variables
@@ -48,6 +53,12 @@ Accessing the environment variables should be done via the [`Environment::getEnv
 use SilverStripe\Core\Environment;
 Environment::getEnv('SS_DATABASE_CLASS');
 ```
+
+[hint]
+The `Environment::getEnv()` method will return `false` both if there was no value set for a variable or if
+the variable was explicitly set as `false`. You can determine whether a variable has been set by calling
+[`Environment::hasEnv()`](api:SilverStripe\Core\Environment::hasEnv()).
+[/hint]
 
 Individual settings can be assigned via [`Environment::setEnv()`](api:SilverStripe\Core\Environment::setEnv()) or [`Environment::putEnv()`](api:SilverStripe\Core\Environment::putEnv()) methods.
 
@@ -74,16 +85,16 @@ SilverStripe\Core\Injector\Injector:
             ThisWillNotSubstitute: 'lorem `REGULAR_TEXT` ipsum'
 ```
 
-[info]
+[warning]
 Environment variables cannot be used outside of Injector config.
-[/info]
+[/warning]
 
 ## Including an extra .env file
 
 Sometimes it may be useful to include an extra `.env` file - on a shared local development environment where all
 database credentials could be the same. To do this, you can add this snippet to your `app/_config.php` file:
 
-Note that by default variables cannot be overloaded from this file; Existing values will be preferred over values in
+Note that by default variables cannot be overridden from this file; Existing values will be preferred over values in
 this file.
 
 ```php
@@ -93,40 +104,42 @@ $loader = new EnvironmentLoader();
 $loader->loadFile($env);
 ```
 
+[warning]
+Note that because `_config.php` is processed after yaml configuration, variables set in these extra `.env` files cannot be used inside yaml config.
+[/warning]
+
 ## Core environment variables
 
 Silverstripe core environment variables are listed here, though you're free to define any you need for your application.
 
 | Name  | Description |
 | ----  | ----------- |
-| `SS_DATABASE_CLASS` | The database class to use, MySQLDatabase, MSSQLDatabase, etc. defaults to MySQLDatabase.|
-| `SS_DATABASE_SERVER`| The database server to use, defaulting to localhost.|
+| `SS_DATABASE_CLASS` | The database class to use. Only `MySQLDatabase` is included by default, but other values are available in optional modules such as [`PostgreSQLDatabase`](https://github.com/silverstripe/silverstripe-postgresql). Defaults to `MySQLDatabase`.|
+| `SS_DATABASE_SERVER`| The database server to use. Defaults to `localhost`.|
 | `SS_DATABASE_USERNAME`| The database username (mandatory).|
 | `SS_DATABASE_PASSWORD`| The database password (mandatory).|
 | `SS_DATABASE_PORT`|     The database port.|
 | `SS_DATABASE_SUFFIX`|   A suffix to add to the database name.|
 | `SS_DATABASE_PREFIX`|   A prefix to add to the database name.|
+| `SS_DATABASE_NAME` | The database name. If not set, `SS_DATABASE_CHOOSE_NAME` must be set instead. |
+| `SS_DATABASE_CHOOSE_NAME`| Boolean/Int. If defined, then the system will choose a default database name for you. The database name will be "SS_" followed by the name of the folder into which you have installed Silverstripe.<br />If `SS_DATABASE_CHOOSE_NAME` is an integer greater than one, then an ancestor folder will be used for the  database name. This is handy for a site that's hosted from `/sites/examplesite/www` or `/buildbot/allmodules-2.3/build`. If it's `2`, the parent folder will be chosen; if it's `3` the grandparent, and so on.<br /><br />Ignored if `SS_DATABASE_NAME` is set.|
 | `SS_DATABASE_TIMEZONE`| Set the database timezone to something other than the system timezone.
-| `SS_DATABASE_NAME` | Set the database name. Assumes the `$database` global variable in your config is missing or empty. |
-| `SS_DATABASE_CHOOSE_NAME`| Boolean/Int. If defined, then the system will choose a default database name for you if one isn't give in the $database variable. The database name will be "SS_" followed by the name of the folder into which you have installed Silverstripe. If this is enabled, it means that the phpinstaller will work out of the box without the installer needing to alter any files. This helps prevent accidental changes to the environment. If `SS_DATABASE_CHOOSE_NAME` is an integer greater than one, then an ancestor folder will be used for the  database name. This is handy for a site that's hosted from /sites/examplesite/www or /buildbot/allmodules-2.3/build. If it's 2, the parent folder will be chosen; if it's 3 the grandparent, and so on.|
-| `SS_DEPRECATION_ENABLED` | Enable deprecation notices for this environment. `SS_ENVIRONMENT_TYPE` must be set to `dev` for deprecation notices to show. |
-| `SS_ENVIRONMENT_TYPE`| The environment type: dev, test or live.|
-| `SS_DEFAULT_ADMIN_USERNAME`| The username of the default admin. This is a user with administrative privileges.|
-| `SS_DEFAULT_ADMIN_PASSWORD`| The password of the default admin. This will not be stored in the database.|
-| `SS_USE_BASIC_AUTH`| Baseline protection for requests handled by Silverstripe. Usually requires additional security measures for comprehensive protection. See [Environment Types](/developer_guides/debugging/environment_types) for caveats.|
-| `SS_SEND_ALL_EMAILS_TO`| If you define this constant, all emails will be redirected to this address.|
-| `SS_SEND_ALL_EMAILS_FROM`| If you define this constant, all emails will be sent from this address.|
-| `SS_ERROR_LOG` | Relative path to the log file. |
-| `SS_PROTECTED_ASSETS_PATH` | Path to secured assets - defaults to ASSETS_PATH/.protected |
-| `SS_DATABASE_MEMORY` | Used for SQLite3 DBs |
-| `SS_TRUSTED_PROXY_IPS` | IP address or CIDR range to trust proxy headers from. If left blank no proxy headers are trusted. Can be set to 'none' (trust none) or '*' (trust all) |
-| `SS_ALLOWED_HOSTS` | A comma deliminated list of hostnames the site is allowed to respond to |
-| `SS_MANIFESTCACHE` | The manifest cache to use (defaults to file based caching). Must be a `Psr\Cache\CacheItemPoolInterface`, `Psr\SimpleCache\CacheInterface`, or `SilverStripe\Core\Cache\CacheFactory` class implementation |
-| `SS_IGNORE_DOT_ENV` | If set the .env file will be ignored. This is good for live to mitigate any performance implications of loading the .env file |
-| `SS_BASE_URL` | The url to use when it isn't determinable by other means (eg: for CLI commands) |
-| `SS_DATABASE_SSL_KEY` | Absolute path to SSL key file |
-| `SS_DATABASE_SSL_CERT` | Absolute path to SSL certificate file |
-| `SS_DATABASE_SSL_CA` | Absolute path to SSL Certificate Authority bundle file |
-| `SS_DATABASE_SSL_CIPHER` | Optional setting for custom SSL cipher |
-| `SS_FLUSH_ON_DEPLOY` | Try to detect deployments through file system modifications and flush on the first request after every deploy. Does not run "dev/build", but only "flush". Possible values are `true` (check for a framework PHP file modification time), `false` (no checks, skip deploy detection) or a path to a specific file or folder to be checked. See [DeployFlushDiscoverer](api:SilverStripe\Core\Startup\DeployFlushDiscoverer) for more details.<br /><br />False by default. |
-| `SS_TEMP_PATH` | File storage used for the default cache adapters in [Manifests](/developer_guides/execution_pipeline/manifests), [Object Caching](/developer_guides/performance/caching) and [Partial Template Caching](/developer_guides/templates/partial_template_caching). It defaults to creating a sub-directory of PHP's built-in `sys_get_temp_dir()`. Can be an absolute path (with a leading `/`), or a path relative to the project root. |
+| `SS_DEPRECATION_ENABLED` | Enable deprecation notices for this environment. `SS_ENVIRONMENT_TYPE` must be set to `dev` for deprecation notices to show. See [Deprecations](/upgrading/deprecations/) for more information. |
+| `SS_DEPRECATION_SHOW_HTTP` | Include deprecation warnings in HTTP responses if `SS_ENVIRONMENT_TYPE` is true. Defaults to false. |
+| `SS_DEPRECATION_SHOW_CLI` | Include deprecation warnings in CLI responses if `SS_ENVIRONMENT_TYPE` is true. Defaults to true. |
+| `SS_ENVIRONMENT_TYPE`| The environment type. Should be one of `dev`, `test`, or `live`. See [Environment Types](/debugging/environment_types/) for more information. |
+| `SS_DEFAULT_ADMIN_USERNAME`| The username of the default admin. This is a user with administrative privileges. |
+| `SS_DEFAULT_ADMIN_PASSWORD`| The password of the default admin. This will not be stored in the database. |
+| `SS_USE_BASIC_AUTH`| Baseline protection for requests handled by Silverstripe. Usually requires additional security measures for comprehensive protection. Set this to the name of a permission which users must have to be able to access the site (e.g. "ADMIN"). See [Environment Types](/developer_guides/debugging/environment_types) for caveats. |
+| `SS_SEND_ALL_EMAILS_TO`| If you define this constant all emails will be redirected to this address, overriding the original address(es). |
+| `SS_SEND_ALL_EMAILS_FROM`| If you define this constant all emails will be sent from this address, overriding the original address. |
+| `SS_ERROR_LOG` | Path to a file for logging errors, relative to the project root. See [Logging and Error Handling](/developer_guides/debugging/error_handling/) for more information about error logging. |
+| `SS_PROTECTED_ASSETS_PATH` | Path to secured assets - defaults to `ASSETS_PATH/.protected`. See [Protected file paths](/developer_guides/files/file_storage/#protected-file-paths) for more information. |
+| `SS_DATABASE_MEMORY` | Used for [SQLite3](https://github.com/silverstripe/silverstripe-sqlite3) database connectors. |
+| `SS_TRUSTED_PROXY_IPS` | IP address or CIDR range to trust proxy headers from. If left blank no proxy headers are trusted. Can be set to 'none' (trust none) or '*' (trust all). See [Request hostname forgery](/developer_guides/security/secure_coding/#request-hostname-forgery) for more information. |
+| `SS_ALLOWED_HOSTS` | A comma deliminated list of hostnames the site is allowed to respond to. See [Request hostname forgery](/developer_guides/security/secure_coding/#request-hostname-forgery) for more information. |
+| `SS_MANIFESTCACHE` | The manifest cache to use (defaults to file based caching). Must be a `Psr\Cache\CacheItemPoolInterface`, `Psr\SimpleCache\CacheInterface`, or `SilverStripe\Core\Cache\CacheFactory` class implementation. |
+| `SS_IGNORE_DOT_ENV` | If set, the `.env` file will be ignored. This is good for live to mitigate any performance implications of loading the `.env` file. |
+| `SS_BASE_URL` | The url to use when it isn't determinable by other means (eg: for CLI commands). Should either start with a protocol (e.g. `https://www.example.com`) or with a double forward slash (e.g. `//www.example.com`). |
+| `SS_FLUSH_ON_DEPLOY` | Try to detect deployments through file system modifications and flush on the first request after every deploy. Does not run "dev/build" - only "flush". Possible values are `true` (check for a framework PHP file modification time), `false` (no checks, skip deploy detection) or a path to a specific file or folder to be checked. See [DeployFlushDiscoverer](api:SilverStripe\Core\Startup\DeployFlushDiscoverer) for more details.<br /><br />False by default. |
+| `SS_TEMP_PATH` | File storage used for the default cache adapters in [Manifests](/developer_guides/execution_pipeline/manifests), [Object Caching](/developer_guides/performance/caching) and [Partial Template Caching](/developer_guides/templates/partial_template_caching). Can be an absolute path (with a leading `/`), or a path relative to the project root. Defaults to creating a sub-directory of PHP's built-in `sys_get_temp_dir()` or using the `silverstripe-cache` directory relative to the project root if one is present. |
--- a/en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
+++ b/en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
@@ -6,7 +6,7 @@ icon: database
 
 # Introduction to the Data Model and ORM
 
-Silverstripe uses an [object-relational mapping](https://en.wikipedia.org/wiki/Object-relational_mapping) to represent its
+Silverstripe uses an [object-relational mapping](https://en.wikipedia.org/wiki/Object-relational_mapping) (aka "ORM") to represent its
 information.
 
 *  Each database table maps to a PHP class.
@@ -27,11 +27,14 @@ use SilverStripe\ORM\DataObject;
 
 class Player extends DataObject 
 {
+    private static $table_name = 'Player';
+
     private static $db = [
         'PlayerNumber' => 'Int',
         'FirstName' => 'Varchar(255)',
         'LastName' => 'Text',
-        'Birthday' => 'Date'
+        'Birthday' => 'Date',
+        'Status' => 'Varchar(255)',
     ];
 }
 ```
@@ -39,14 +42,20 @@ class Player extends DataObject
 This `Player` class definition will create a database table `Player` with columns for `PlayerNumber`, `FirstName` and
 so on. After writing this class, we need to regenerate the database schema.
 
+[hint]
+You can technically omit the `table_name` property, and a default table name will be created based on the fully qualified class name - but this can result in table names that are too long for the database engine to handle. We recommend that you _always_ explicitly declare a table name for your models.
+
+See more in [Mapping classes to tables with `DataObjectSchema`](#mapping-classes-to-tables) below.
+[/hint]
+
 ## Generating the Database Schema
 
 After adding, modifying or removing `DataObject` subclasses, make sure to rebuild your Silverstripe CMS database. The
-database schema is generated automatically by visiting the `/dev/build` (e.g. `https://www.example.com/dev/build`) in your browser
-while authenticated as an administrator.
+database schema is generated automatically by visiting `/dev/build` (e.g. `https://www.example.com/dev/build`) in your browser
+while authenticated as an administrator, or by running `sake dev/build` on the command line (see [Command Line Interface](/developer_guides/cli/) to learn more about `sake`).
 
 [info]
-In dev mode, you do not need to be authenticated to build the database by visiting the `/dev/build` route.
+In "dev" mode, you do not need to be authenticated to run `/dev/build`. See [Environment Types](/developer_guides/debugging/environment_types) for more information.
 [/info]
 
 This script will analyze the existing schema, compare it to what's required by your data classes, and alter the schema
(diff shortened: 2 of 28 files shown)


GuySartorelli:
A note on overloading vs overriding:
There are a few times where the docs used to say "overload" and I've changed this to "override".
This is because the way the word "overload" was being used fits neither the [generally accepted definition of overloading in software development](https://en.wikipedia.org/wiki/Function_overloading) (i.e. declaring multiple methods in the same class with different signatures, which is not supported in PHP) nor the [definition of overloading for PHP specifically](https://www.php.net/manual/en/language.oop5.overloading.php) (i.e. usage of a specific subset of PHP's magic methods).

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
```diff
@@ -32,53 +32,43 @@ class Player extends DataObject
 
 ## Available Types
```
ChloeSartorelli, line 33: I used https://github.com/silverstripe/silverstripe-framework/blob/5/_config/model.yml as my reference here

en/02_Developer_Guides/00_Model/05_Extending_DataObjects.md
ChloeSartorelli, line 1: Not only did the examples here not work as advertised, they also promoted _terrible_ practices, so they had to be completely rewritten.
I couldn't think of a specific valid use-case for using `onBeforeDelete()` so I left it open-ended - which is fine, because the important part of the example is showing how to implement the extension hook in the first place.

en/00_Getting_Started/03_Environment_Management.md
```diff
@@ -35,7 +35,12 @@ not be detected automatically. If you wish to load environment variables from a
 need to do so manually. See the [Including an extra `.env` file](#including-an-extra-env-file) section below for more
 information.
 
-## Managing environment variables with Apache
+## Other ways to manage environment variables
+
+Silverstripe CMS will respect "real" environment variables provided by the server.
```
emteknetnz:
```suggestion
Silverstripe CMS will respect environment variables provided by the server.
```

emteknetnz: Also remove the mention to "real" environment variables in the apache section below

en/00_Getting_Started/03_Environment_Management.md
```diff
@@ -49,6 +54,12 @@ use SilverStripe\Core\Environment;
 Environment::getEnv('SS_DATABASE_CLASS');
 ```
 
+[hint]
+The `Environment::getEnv()` method will return `false` both when there was no value set for a variable and when
```
emteknetnz:
```suggestion
The `Environment::getEnv()` method will return `false` both if there was no value set for a variable or if
```
Just make it a little clear they're mutually exclusive reasons

en/00_Getting_Started/03_Environment_Management.md
```diff
@@ -93,40 +104,42 @@ $loader = new EnvironmentLoader();
 $loader->loadFile($env);
 ```
 
+[warning]
+Note that because `_config.php` is processed after yaml configuration, variables set in these extra `.env` files cannot be used inside yaml config.
```
emteknetnz, line 108: They can?

e.g. https://github.com/silverstripe/silverstripe-framework/blob/5.0/_config/mailer.yml#L28

en/00_Getting_Started/03_Environment_Management.md
```diff
@@ -93,40 +104,42 @@ $loader = new EnvironmentLoader();
...
+| `SS_FLUSH_ON_DEPLOY` | Try to detect deployments through file system modifications and flush on the first request after every deploy. Does not run "dev/build" - only "flush". Possible values are `true` (check for a framework PHP file modification time), `false` (no checks, skip deploy detection) or a path to a specific file or folder to be checked. See [DeployFlushDiscoverer](api:SilverStripe\Core\Startup\DeployFlushDiscoverer) for more details.<br /><br />False by default. |
+| `SS_TEMP_PATH` | File storage used for the default cache adapters in [Manifests](/developer_guides/execution_pipeline/manifests), [Object Caching](/developer_guides/performance/caching) and [Partial Template Caching](/developer_guides/templates/partial_template_caching). Can be an absolute path (with a leading `/`), or a path relative to the project root. Defaults to creating a sub-directory of PHP's built-in `sys_get_temp_dir()` or using the `silverstripe-cache` directory relative to the project root if one is present. |
```
emteknetnz, line 145: Just wondering if it's worth alpha sorting all of these

en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
```diff
@@ -27,72 +27,64 @@ use SilverStripe\ORM\DataObject;
...
+You can technically omit the `table_name` property, and a default table name will be created based on the fully qualified class name - but this can result in table names that are too long for the database engine to handle. We recommend that you _always_ explicitly declare a table name for your models.
+
+See more in [Mapping classes to tables with `DataObjectSchema`](#mapping-classes-to-tables) below.
+[/hint]
+
 ## Generating the Database Schema
 
 After adding, modifying or removing `DataObject` subclasses, make sure to rebuild your Silverstripe CMS database. The
-database schema is generated automatically by visiting the URL http://www.yoursite.com/dev/build while authenticated as an administrator.
+database schema is generated automatically by visiting the URL http://www.yoursite.com/dev/build while authenticated as an administrator, or by running `sake dev/build` on the command line (see [Command Line Interface](/developer_guides/cli/) to learn more about `sake`).
```
emteknetnz: Does `sake dev/build` ever actually work?  Just wondering if it's worth updating any references to `sake` to `vendor/bin/sake`?

en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
```diff
@@ -112,24 +105,26 @@ CREATE TABLE `Player` (
...
 
-A new instance of a [DataObject](api:SilverStripe\ORM\DataObject) can be created using the `new` syntax.
+A new instance of a [DataObject](api:SilverStripe\ORM\DataObject) can be created using the `new` keyword.
 
 ```php
 $player = new Player();
 ```
 
-Or, a better way is to use the `create` method.
+But a better way is to use the `create()` method.
```
emteknetnz:
```suggestion
However a better way is to use the `create()` method.
```
"But" negates the previous statement. It's not incorrect to use `new`

en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
```diff
@@ -153,48 +148,75 @@ $id = $player->write();
...
-// returns the players 'ID' column value
+[hint]
+All of the above methods to get a single record will return `null` if there is no record to return.
+[/hint]
 
-echo $player->dbObject('LastEdited')->Ago();
-// calls the `Ago` method on the `LastEdited` property.
-```
+[info]
+`DataObject::get()->byID()` and `DataObject::get_by_id()` achieve similar results, but the object returned by `DataObject::get_by_id()` is cached against a `static` property within `DataObject`.
```
emteknetnz:
```suggestion
`DataObject::get()->byID()` and `DataObject::get_by_id()` achieve similar results, though the object returned by `DataObject::get_by_id()` is cached against a `static` property within `DataObject`.
```

en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
```diff
@@ -153,48 +148,75 @@ $id = $player->write();
...
-
-// returns a `DataList` containing all the `Player` records that have the `FirstName` of 'Sam'
-
 ```
 
+There's a lot more to filtering and sorting, so make sure to keep reading.
+
 [info]
-Provided `filter` values are automatically escaped and do not require any escaping.
+Values passed in to the `filter()` and `sort()` methods are automatically escaped and do not require any additional escaping. This makes it easy to safely filter/sort records by user input.
```
emteknetnz, line 188: Is there any mention in the ORM docs for the raw-sql where() method vs the non-raw-sql sort() method?

en/02_Developer_Guides/00_Model/02_Relations.md
```diff
@@ -126,18 +132,51 @@ be necessary. E.g. Additional parent classes for each respective relationship, o
...
+When defining a `has_many` relation, you _must_ specify a `has_one` relationship on the related class as well. To add a `has_one` relation on core classes, yml config settings can be used:
+
 ```yml
 SilverStripe\Assets\Image:
   has_one:
-    MyDataObject: MyDataObject
+    Team: App\Model\Team
 ```
+
+Note that in some cases you may be better off using a `many_many` relation instead. Carefully consider whether you are defining a "1-to-many" or a "many-to-many" relationship.
```
emteknetnz: Wondering if we keep the somewhat more obvious though kind of weird looking 1-to-many and 1-to-1 notation or change it probably more grammatically correct to one-to-many and one-one

en/02_Developer_Guides/00_Model/02_Relations.md
```diff
@@ -651,6 +748,64 @@ As these lists are not backed by the database, most of the filtering methods on
...
+
+By default, this service only inspects relations for classes which have either no namespace or a namespace beginning with `App\`. You can declare your own namespaces by setting the `allow_rules` configuration:
+
+```yaml
+SilverStripe\Dev\Validation\RelationValidationService:
+  allow_rules:
+    # using the "app" key you can override the default "App" namespace
+    app: 'MyApp'
+    # you can add any namespace declarations with-or-without keys
+    - 'AnotherApp'
```
emteknetnz: This mixed key/no-key notation looks really weird.  And having 'AnotherApp' as an example namespace is also weird because why would you have 2x "apps" in your project?

en/02_Developer_Guides/00_Model/02_Relations.md
```diff
@@ -651,6 +748,64 @@ As these lists are not backed by the database, most of the filtering methods on
...
+    # you can declare explicitly sub-namespaces
+    - 'MyOrg\MyModule'
```
emteknetnz: What's a sub-namespace? Isn't that just a namespace?

en/02_Developer_Guides/00_Model/03_Lists.md
```diff
@@ -27,65 +29,62 @@ foreach($members as $member) {
...
+//    'sig@example.com',
+//    'will@example.com'
 // ];
+$emailAddresses = Member::get()->column('Email');
 ```
 
 ## Iterating over a large list {#chunkedFetch}
 
-When iterating over a DataList, all DataObjects in the list will be loaded in memory. This can consume a lot of memory when working with a large data set.
+When iterating over a `DataList`, the ORM will create a [`Generator`](https://www.php.net/manual/en/language.generators.overview.php). This means we don't have all of the `DataObject` records in the list instantiated in memory, but the ORM _does_ fetch all of the data about those records and loads that data in memory. This can consume a lot of memory when working with a large data set.
```
emteknetnz, line 82: This clarification actually makes this more confusing - It's not clear where boundaries of the ORM ends, I would have though that DataList resides in the ORM.  So it seems like you're contradicting yourself saying that there's a Generator to save on memory, but it does nothing cos it's all loaded into memory

Is the in inner workings of DataList creating a generator relevant to the end user?

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
```diff
@@ -157,81 +204,70 @@ class Player extends DataObject
...
-        return '1982-01-01';
-    }
-}
-
-$obj = new MyObject;
-$obj->getMyDate(); // returns string
-$obj->MyDate; // returns string
-$obj->obj('MyDate'); // returns object
-$obj->obj('MyDate')->InPast(); // returns boolean
+$player = Player::get()->byId(1);
```
emteknetnz, line 238:
```suggestion
$obj = Player::get()->byId(1);
```
?

en/02_Developer_Guides/00_Model/05_Extending_DataObjects.md
```diff
@@ -5,81 +5,79 @@ summary: Modify the data model without using subclasses.
...
+        if ($this->isChanged('TeamID') && $this->Team()->exists()) {
+            // If the player's number is already used by someone else on this team
+            if (in_array($this->Number, $this->Team()->Players()->exclude('ID', $this->ID)->column('Number'))) {
+                // Assign a new player number
+                $this->Number = $this->Team()->getAvailablePlayerNumber();
+            }
         }
         
-        // CAUTION: You are required to call the parent-function, otherwise
+        // CAUTION: You are required to call parent::onBeforeWrite(), otherwise
```
emteknetnz, line 57: We should probably have `parent::onBeforeWrite();` at the top of the method in our examples. I know it's fine to have it at the bottom, even in the middle of a method, though I think it's far more standard to call the parent method before the child method.

en/02_Developer_Guides/00_Model/05_Extending_DataObjects.md
```diff
@@ -5,81 +5,79 @@ summary: Modify the data model without using subclasses.
...
     {
-        if(!Permission::check('PLAYER_DELETE')) {
-            Security::permissionFailure($this);
-            exit();
-        }
-        
+        /* Do some cleanup here relevant to your project before deleting the actual database record */
+
+        // CAUTION: You are required to call parent::onBeforeDelete(), otherwise
+        // SilverStripe will not execute the request.
```
emteknetnz, line 80: Should move this to top of the method

en/02_Developer_Guides/00_Model/06_SearchFilters.md
```diff
@@ -18,59 +18,66 @@ you can put on field names to change this behavior. These are represented as `Se
...
-// to fetch any player that's name contains the letter 'z'
+// fetch any player whose name contains the letter 'z'
 $players = Player::get()->filterAny([
     'FirstName:PartialMatch' => 'z',
     'LastName:PartialMatch' => 'z'
 ]);
 ```
 
+[hint]
+Notice the syntax - to invoke a `SearchFilter` in a `DataList`'s `filter*()` or `exclude*()` methods, you add a colon after the field name, followed by the name of the filter (excluding the actual word "filter"). e.g. for a `StartsWithFilter`: `'FieldName:StartsWith'`
```
emteknetnz: Where are there a `*` for `filter*()` or `exclude*()`?

en/02_Developer_Guides/00_Model/07_Permissions.md
```diff
@@ -47,12 +53,80 @@ class MyDataObject extends DataObject
...
+use SilverStripe\Security\Permission;
+
+class MyDataObject extends SomeParentObject 
+{
+    public function canView($member = null) 
+    {
+        // If any extension returns false, the result will be false
+        // otherwise if any extension returns true, the result will be true
+        $extended = $this->extendedCan(__FUNCTION__, $member);
+        // You could also explicitly check for false values, and ignore any true values, if you don't
```
emteknetnz: This statement doesn't seem to relate to the code directly below it.  The comment above indicates that "false returns have precedence", so ... This whole thing is confusing

en/02_Developer_Guides/00_Model/07_Permissions.md
```diff
@@ -47,12 +53,80 @@ class MyDataObject extends DataObject
...
+        }
+
+        // If no extensions return true or false, check for a specific permission here.
+        return Permission::check('CMS_ACCESS_CMSMain', 'any', $member);
+    }
+
+    public function canEdit($member = null) 
+    {
+        // If the parent class says the member can't perform this action, don't let them do it.
+        // Be careful though - if the parent class doesn't implement canEdit(), you will end up
```
emteknetnz: The parent class will always implements canEdit(), because the can* methods are defined on DataObject

en/02_Developer_Guides/00_Model/08_SQL_Select.md
```diff
@@ -56,20 +61,51 @@ how to properly prepare user input and variables for use in queries
...
+* [`sqlColumnForField()`](api:SilverStripe\ORM\DataObjectSchema::sqlColumnForField()): Get the ANSI-quoted table and column name for a given `DataObject` field (in `"Table"."Field"` format)
+* [`tableForField()`](api:SilverStripe\ORM\DataObjectSchema::tableForField()): Get the table name in the class hierarchy which contains a given field column.
+* [`tableName()`](api:SilverStripe\ORM\DataObjectSchema::tableName()): Get table name for the given class. Note that this does not confirm a table actually exists (or should exist), but returns the name that would be used if this table did exist. Male sure to call `classHasTable()` before using this table name in a query.
+
+[hint]
+While the default database connector will work fine without explicitly ANSI-quoting table names in queries, it is good practice to make sure they are quoted (especially if you're writing these queries in a module that will be publicly shared) to ensure your queries will work on other database connectors such as [`PostgreSQLDatabase`](https://github.com/silverstripe/silverstripe-postgresql) which explicitly require ANSI quoted table names.
```
emteknetnz, line 75:
```suggestion
While the default database connector will work fine without explicitly ANSI-quoting table names in queries, it is good practice to make sure they are quoted if you're writing these queries in a module that will be publicly shared to ensure your queries will work on other database connectors such as [`PostgreSQLDatabase`](https://github.com/silverstripe/silverstripe-postgresql) which explicitly require ANSI quoted table names.
```
Wouldn't recommend people do this 99% of time when they're making a website that uses mysql, it'll just be unnecessary code spam.

en/02_Developer_Guides/00_Model/10_Versioning.md
```diff
@@ -24,26 +24,29 @@ You can check them out on github:
...
 ### Stages
 
-In most cases, you'll want to have one polished version of a `Page` visible to the general public while your editors might be working off a draft version. Silverstripe CMS handles this through the concept of _stage_.
+In most cases, you'll want to have one polished version of a `Page` visible to the general public while your editors might be making additional changes on a draft version. Silverstripe CMS handles this through the concept of _stages_.
 
-By default, adding the `Versioned` extension to a DataObject will create 2 stages:
-* "Stage" for tracking draft content
-* "Live" for tracking content publicly visible.
+By default, adding the `Versioned` extension to a `DataObject` will create 2 stages:
+* "Stage" for tracking draft content (aka "draft")
```
emteknetnz, line 40: Might be worth referencing that we can there is the Versioned::DRAFT = 'Stage'; constant.  Or maybe it's just not helpful in this context

en/02_Developer_Guides/00_Model/10_Versioning.md
```diff
@@ -156,16 +174,201 @@ Versioning only works if you are adding the extension to the base class. That is
...
+
+These methods accept an optional `Member` argument. If not provided, they will assume you want to check the permission against the current `Member`. When performing a version operation on behalf of a `Member`, you'll probably want to use these methods to confirm they are authorised.
+
+[warning]
+Like with the base `can` permission checks, these checks are _not_ performed automatically when invoking the associated action via PHP. i.e. if you call `publishSingle()` on a record in your own code, Silverstripe CMS will _not_ check if the currently authenticated user has permission to publish the record. Make sure you are performing permission checks by calling these `can` methods before invoking the associated actions.
+[/warning]
+
+```php
+use SilverStripe\Security\Security;
+
```
emteknetnz:
```suggestion
```

en/02_Developer_Guides/00_Model/10_Versioning.md
```diff
@@ -156,16 +174,201 @@ Versioning only works if you are adding the extension to the base class. That is
...
+The `canViewStage()` method can be used to check if a Member can view a specific stage of a record. It should be invoked by user code to check if a record is visible in the given stage.
+
+```php
+use SilverStripe\Versioned\Versioned;
+
+// Check if `$member` can view the Live version of $record.
+$record->canViewStage(Versioned::LIVE, $member);
+
+// Check if `$member` can view the Stage version of $record.
```
emteknetnz:
```suggestion
// Check if `$member` can view the Stage/Draft version of $record.
```
Wondering if it's worth using stage/draft for everything - would that reduce confusion?

en/02_Developer_Guides/00_Model/10_Versioning.md
```diff
@@ -156,16 +174,201 @@ Versioning only works if you are adding the extension to the base class. That is
...
+$record->canViewStage();
+```
+
+For the `can` methods that all `DataObject` models have, see [Model-Level Permissions](permissions).
+
+#### Customising permissions for a versioned DataObject
+
+`Versioned` record visibility can be customised in one of the following ways by editing your code:
+
+ * Override the `canViewVersioned()` method in your `DataObject` subclass. Make sure that this returns `true`, or
```
emteknetnz, line 282: Should this be Override `canViewStage` instead since we're saying `view this object in the current viewing mode.` here?  

Regardless, this should probbly be below the canView bullet point below because sure this is second option, where overriding canView should be the first option?

en/02_Developer_Guides/00_Model/10_Versioning.md
```diff
@@ -176,28 +379,32 @@ class MyPage extends Page
...
     ];
+
     private static $owns = [
         'Image'
     ];
 }
 ```
 
-If a `MyPage` gets published, all its related `Banners` will also be published, which will cause all `Image` DataObjects to be published.
+If a `MyPage` record gets published, all its related `Banners` will also be published, which will cause all related `Image` records to be published.
```
emteknetnz, line 407: Will $owns  also cascade unpublish/archive as well? If so should mention that here.

en/02_Developer_Guides/00_Model/10_Versioning.md
```diff
@@ -314,90 +537,126 @@ class Page extends SiteTree
...
+
+// Don't forget to set the mode back afterwards!
+Versioned::set_reading_mode($oldMode);
+```
+
+[info]
+`Versioned::set_stage(Versioned::LIVE)` is the equivalent of `Versioned::set_reading_mode('Stage.' . Versioned::LIVE)`.
+[/info]
+
+That said, it is better practice to use [`Versioned::withVersionedMode()`](api:SilverStripe\Versioned\Versioned::withVersionedMode()) because this removes the possibility of forgetting to set the stage back to what it was afterward.
```
emteknetnz: Should provide the better alternative as the first example. Then you can provide the old style example as a second example, if at all.

emteknetnz (changes requested):

en/00_Getting_Started/03_Environment_Management.md
ChloeSartorelli: Note this is _explicitly referring to extra .env files_ - not the normal `.env` file that we use by default.

en/00_Getting_Started/03_Environment_Management.md
ChloeSartorelli: That's out of scope for this PR.

en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
ChloeSartorelli: The word "but" doesn't automatically mean everything written before it is incorrect or untrue.... but I'll make this change to avoid PR ping pong since it doesn't materially change anything.

en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
ChloeSartorelli: I don't understand the question.... why would `where()` be compared with `sort()` at all?
Do you mean comparing `filter()` vs `where()`  and `sort()` vs `orderBy()`?

en/02_Developer_Guides/00_Model/02_Relations.md
ChloeSartorelli: > This mixed key/no-key notation looks really weird

That's fine, it's allowed to look weird, so long as it's correct and is accurately demonstrating what can be done.
In this case, it's demonstrating that you can choose to have keys or to omit them. Which is correct. This PR is not about making the docs show best practice, as mentioned in the description.

> having 'AnotherApp' as an example namespace is also weird because why would you have 2x "apps" in your project?

You wouldn't have two apps in the same project, but you also wouldn't declare "App" twice. This seemed like the lesser of two evils. But again, best practice isn't what this PR is about anyway.

en/02_Developer_Guides/00_Model/02_Relations.md
ChloeSartorelli: Partial namespace would probably be clearer. I'll use that wording instead.
Basically, you can declare `SilverStripe\ORM` and it'll include every FQCN which has that at the start of its class name.

en/02_Developer_Guides/00_Model/03_Lists.md
ChloeSartorelli: I..... don't know what you mean. But yes I think it is relevant to say that the _actual `DataObject` instances_ are not loaded into memory.

> it seems like you're contradicting yourself saying that there's a Generator to save on memory, but it does nothing cos it's all loaded into memory

The _raw data_ is loaded into memory. This data is not then wrapped in instantiated `DataObject` objects, which means less memory is used than would have been.

en/02_Developer_Guides/00_Model/05_Extending_DataObjects.md
ChloeSartorelli: It doesn't matter where in the method it gets called, so I'm inclined to put this in the "out of scope" bucket for this PR.

en/02_Developer_Guides/00_Model/05_Extending_DataObjects.md
ChloeSartorelli: It doesn't matter where in the method it gets called, so I'm inclined to put this in the "out of scope" bucket for this PR.

en/02_Developer_Guides/00_Model/06_SearchFilters.md
ChloeSartorelli: Because it applies to all of the methods which start with `filter` or `exclude` - I can explicitly mention those methods here instead if you'd like.

en/02_Developer_Guides/00_Model/07_Permissions.md
ChloeSartorelli: `DataObject` implements `canEdit()`. The parent class may not be `DataObject`, and may not implement the method itself. It will _inherit_ the method, but may not implement it. I'll add the word "explicitly" before the word "implement" to make this clearer.

en/02_Developer_Guides/00_Model/08_SQL_Select.md
ChloeSartorelli: > it is good practice to make sure they are quoted if you're writing these queries in a module that will be publicly shared

It doesn't say "if you are making your own website code is not going to be shared" - it's recommending this _only_ for modules, and _only_ if they're intended to be shared.

en/02_Developer_Guides/00_Model/10_Versioning.md
ChloeSartorelli: > Might be worth referencing that we can there is the Versioned::DRAFT = 'Stage'; constant.

Can you please reword this? I think you may have made a typo in there somewhere, and it makes it hard to know what you meant.

en/02_Developer_Guides/00_Model/10_Versioning.md
ChloeSartorelli: Depends on the reader. It would be just as easy to read that as "the Stage version or the Draft version"
I should have put quotes around "Stage" there, though. I'll fix that.

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
ChloeSartorelli: Oops, copy/paste error. Fixed. I swapped them all to `$player` to avoid any possible confusion that `$obj->obj()` might cause.

en/02_Developer_Guides/00_Model/10_Versioning.md
ChloeSartorelli: I think you just meant "it might be helpful to mention the `DRAFT` and `LIVE` constants - in which case, I agree! Updated.

en/02_Developer_Guides/00_Model/10_Versioning.md
ChloeSartorelli: Done

en/02_Developer_Guides/00_Model/10_Versioning.md
ChloeSartorelli: Done

en/02_Developer_Guides/00_Model/10_Versioning.md
ChloeSartorelli: The current doc is correct.
`canViewStage()` is only ever invoked from `VersionedResolver::resolveVersionList()`, whereas `canViewVersioned()` is invoked every time you call `canView()` on any versioned record. I've updated the list in "Version specific _can_ methods" to explicitly state this.

ChloeSartorelli: As for whether it's the "first" or "second" option, I'll leave it as is - this section is specifically about versioning, so versioned solutions come first - and also frankly it doesn't matter either way.

en/02_Developer_Guides/00_Model/10_Versioning.md
ChloeSartorelli: Nope - `cascade_deletes` is used for cascading those operations. I'll mention that here.

en/02_Developer_Guides/00_Model/10_Versioning.md
ChloeSartorelli: Done

en/02_Developer_Guides/00_Model/07_Permissions.md
ChloeSartorelli: Done

en/02_Developer_Guides/00_Model/07_Permissions.md
ChloeSartorelli: It does relate to that code - it's saying, in effect "The code below is checking that there is any value other than null, but depending on your use case you may want to explicitly check for a `false` value instead."
I'll use that wording instead, if it makes it clearer.

en/02_Developer_Guides/00_Model/06_SearchFilters.md
ChloeSartorelli: Done

en/02_Developer_Guides/00_Model/02_Relations.md
ChloeSartorelli: I think my thinking at the time was "the examples above use only the first part of a namespace, but you can include more of the namespace if you want"
I'll make it clearer.

en/02_Developer_Guides/00_Model/02_Relations.md
ChloeSartorelli: I've changed `AnotherApp` to `MyOrg`

en/02_Developer_Guides/00_Model/02_Relations.md
ChloeSartorelli: Changed "1-to-1", "many-to-1`" and "1-to-many" to say "one" instead.

en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
ChloeSartorelli: Done

en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
ChloeSartorelli: Done

en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
ChloeSartorelli: The section linked to mentions putting `sake` into `/usr/bin/sake` which would make it work globally. I'm not inclined to go through and change everywhere sake is mentioned to include the path as part of this PR.

en/00_Getting_Started/03_Environment_Management.md
ChloeSartorelli: Done

en/00_Getting_Started/03_Environment_Management.md
ChloeSartorelli: That wording was pre-existing but I have no problem removing it. Done.

en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
```diff
@@ -444,74 +458,78 @@ $teams = Team::get()->filter('Players.Avg(PointsScored):GreaterThan', 15);
...
 // SELECT * FROM Player WHERE FirstName != 'Sam' AND LastName != 'Minnée'
+$players = Player::get()->exclude('FirstName', 'Sam')->exclude('Surname', 'Minnée');
+$players = Player::get()->excludeAny([
+    'FirstName' => 'Sam',
+    'Surname' => 'Minnée',
+]);
 ```
 
-And removing Sig and Sam with that are either age 17 or 43.
+And removing andone named "Sig" or "Sam" with that are either age 17 or 43.
```
emteknetnz:
```suggestion
And removing one named "Sig" or "Sam" with that are either age 17 or 43.
```

en/02_Developer_Guides/00_Model/03_Lists.md
emteknetnz: I think the confusion is that the first paragraph was originally what the ORM does by default (no generator), however now the word 'generator' has been added to the first paragraph.  By deafult the ORM will not use a generator? Only chunkfetched uses the generator?

I thin the original paragraph was better.  Maybe just revert to that and improve it by add "by default" to the first sentence i.e. "When iterating over a DataList, by default all DataObjects in the list will be loaded in memory"

emteknetnz (changes requested):

emteknetnz:
Needs a rebase

en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
ChloeSartorelli: Should have been "removing any" - if there were more than one it'd remove more than one.
Done.

en/02_Developer_Guides/00_Model/03_Lists.md
ChloeSartorelli: As discussed in person, the ORM _does_ use generators by default in CMS 5. Do you still feel like some change is needed here, given that context?

GuySartorelli:
Rebased

en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
ChloeSartorelli: Can you please explain further? Or is this not a problem afterall?

emteknetnz (approved):

PR #191, opened 2023-03-21
Fix html entities in notices

edwilde:
It looks like the `&amp;` is not being converted to `&` when compiled, as it is being compiled into a `<code>` tag because of the `\``

Screen Shot 2023-03-21 at 4 34 48 PM

Should be safe to just include the `&nbsp;` raw.


GuySartorelli (approved):
Nice catch! Thanks.

PR #200, opened 2023-03-29
DOC Deprecated asyncphp/doorman methods

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/11


GuySartorelli (changes requested):
`asyncphp/doorman` isn't a supported module, so we shouldn't add this here.

PR #205, opened 2023-04-04
Suggest adding extension point in docs when instantiation a custom FieldList for CMS fields

LiamKearn:


michalkleiner:
Good idea, though I would actually rather suggest getting the initial fieldset via `$fields = parent::getCMSFields();` and then adding to it, which for DataObject descendants is pretty easy as it doesn't contain much (unlike pages). That takes care of the extension point for you.

Alternatively, we can promote the use of `beforeUpdateCMSFields` callback which is run after the default fields are scaffolded and before the update hook, as the name suggests, instead of after the default scaffolding and after the extension point when you get the fields via the parent call.

```
public function getCMSFields()
{
    $this->beforeUpdateCMSFields(function (FieldList $fields) {
        $fields->addFieldToTab(
            'Root.Main',
            TextField::create(
                'MyFieldName',
                'My field title'
            )
        );
    });

    return parent::getCMSFields();
}
```

LiamKearn:
The surrounding documentation is around explicitly **not scaffolding fields** (that's my interpretation of "fully" since the other options are previously covered).

I agree that both of the options you provided are more sound for enforcing the use of the hook. I didn't even know about `::beforeUpdateCMSFields` so thanks for sharing that, might be good to add that as a suggestion as well. However I have reservations around scaffolding fields in larger projects in which I've seen that using the scaffolder leads to ridiculously imperative (and slow) form systems.

For example `$fields->fieldByName('ExternalMyName')->setTitle('My Name')` is silly in my opinion. Yet I see it all too often instead of just `TextField::create('ExternalMyName', 'My Name')` which is going to be far easier for the next developer to both grok and change.

Don't get me wrong I completely understand that a lot of the SS community is not going to be interested in manual field creation but I think it's worth keeping a no-scaffolding section in the docs with the added best practices of offering extensibility (which is arguably also going lead to the imperative forms which I dislike 😄).

GuySartorelli:
> For example $fields->fieldByName('ExternalMyName')->setTitle('My Name') is silly in my opinion

Agreed - in that scenario, field labels configuration should probably be preferred. That's neither here-nor-there though, imo.

This PR makes a change that is very similar to the change made for [the CMS 5 docs](https://docs.silverstripe.org/en/5/developer_guides/model/scaffolding/#form-fields). Given that context, @michalkleiner do you think this needs to be clarified further? I agree with @LiamKearn that this part is for _not_ using the scaffolded fields. We could add some wording in there that makes this more explicitly clear if needed.

LiamKearn:
Oh CMS5 docs are absolutely what I was after here. Nice stuff!

That works for me, Happy to close :)

PR #207, opened 2023-04-09
DOC Remove former core committers

maxime-rainville:
Remove Dan and Chris from list of core committers. Chris hasn't been a core committer in ages. Dan stepped down recently.

## Parent issue
https://github.com/silverstripeltd/product-issues/issues/719


michalkleiner:
Thank you @flamerohr and @dhensby for all your work, without your contributions Silverstripe CMS wouldn't be where it is today. Kia ora!

PR #209, opened 2023-04-11
DOC Explicitly set log files using absolute file paths

GuySartorelli:
## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/10551

### DIFF
--- a/en/02_Developer_Guides/07_Debugging/01_Error_Handling.md
+++ b/en/02_Developer_Guides/07_Debugging/01_Error_Handling.md
@@ -163,12 +163,19 @@ SilverStripe\Core\Injector\Injector:
   LogFileHandler:
     class: Monolog\Handler\StreamHandler
     constructor:
-      - "../silverstripe.log"
+      - "/var/www/silverstripe.log"
       - "info"
 ```
 
-The log file will be relative to the main index.php file path (default: inside public/), so "../silverstripe.log" will
-create a file in your project root.
+[warning]
+The log file path must be an absolute file path, as relative paths may behave differently between CLI and HTTP requests. If you want to use a _relative_ path, you can use the `SS_ERROR_LOG` environment variable to declare a file path that is relative to your project root:
+
+```sh
+SS_ERROR_LOG="./silverstripe.log"
+```
+
+You don't need any of the yaml configuration above if you are using the `SS_ERROR_LOG` environment variable - but you can use a combination of the environment variable and yaml configuration if you want to configure multiple error log files.
+[/warning]
 
 The `info` argument provides the minimum level to start logging at.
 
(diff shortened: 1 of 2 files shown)


en/02_Developer_Guides/07_Debugging/01_Error_Handling.md
```diff
@@ -163,12 +163,17 @@ SilverStripe\Core\Injector\Injector:
...
 
-The log file will be relative to the main index.php file path (default: inside public/), so "../silverstripe.log" will
-create a file in your project root.
+[warning]
+The log file path must be an absolute file path, as relative paths may behave differently between CLI and HTTP requests. If you want to use a _relative_ path, you can use the `SS_ERROR_LOG` environment variable to declare a file path that is relative to your project root:
+
+```sh
+SS_ERROR_LOG="./silverstripe.log"
+```
+[/warning]
```
michalkleiner: So when using the env variable, do we still need the whole yaml config? Or some defaults apply in that case? Or is it a complementary setting just for the log path and needs to be done together with the yaml? I think we need a bit more explanation around how those two things work together or whether they are exclusive to each other etc.

ChloeSartorelli: The `SS_ERROR_LOG` is separate to the yaml config. You can have one, the other, or both. I'll try to think of a succinct way to explain that in the doc...

ChloeSartorelli: Done

en/02_Developer_Guides/07_Debugging/01_Error_Handling.md
```diff
@@ -163,12 +163,19 @@ SilverStripe\Core\Injector\Injector:
   LogFileHandler:
     class: Monolog\Handler\StreamHandler
     constructor:
-      - "../silverstripe.log"
+      - "/var/www/silverstripe.log"
```
sabina-talipova, line 166: `var/www/...` path depends on environment settings and it is relevant only if you use Linux system, but for Windows users (without Docker) it might be different. Could we mention about possible differences.

ChloeSartorelli: https://github.com/silverstripe/developer-docs/pull/210#discussion_r1163443803

sabina-talipova:
@michalkleiner if you are happy with making changes, then I'll approve and merge.

michalkleiner (approved):
I would probably still reword some of it since the path doesn't have to be absolute (as we're using the word _must_), a relative one works as well, but has the discovered issues based on the request (CLI vs HTTP) which might be too tedious to explain in full correctness.

This mitigates it well enough, so happy with it. Merge when suits.

sabina-talipova (approved):
Looks good!

PR #210, opened 2023-04-11
DOC Explicitly set deprecation log files using absolute file paths

GuySartorelli:
## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/10551

### DIFF
--- a/en/03_Upgrading/07_Deprecations.md
+++ b/en/03_Upgrading/07_Deprecations.md
@@ -55,7 +55,7 @@ SilverStripe\Core\Injector\Injector:
   ErrorLogFileHandler:
     class: Monolog\Handler\StreamHandler
     constructor:
-      - "../silverstripe.log" # an absolute path, or a path relative to the index.php file (usually inside the public/ directory)
+      - "var/www/silverstripe.log"
       - "warning" # warning is the level deprecation warnings are logged as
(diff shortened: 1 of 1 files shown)


en/03_Upgrading/07_Deprecations.md
```diff
@@ -55,15 +55,23 @@ SilverStripe\Core\Injector\Injector:
   ErrorLogFileHandler:
     class: Monolog\Handler\StreamHandler
     constructor:
-      - "../silverstripe.log" # an absolute path, or a path relative to the index.php file (usually inside the public/ directory)
+      - "var/www/silverstripe.log"
```
sabina-talipova, line 58: `var/www/...` path depends on environment settings and it is relevant only if you use Linux system, but for Windows users (without Docker) it might be different. Could we mention about possible differences.

ChloeSartorelli: Developers should know what _their_ absolute path is. This is meant to be an example - just like all of the rest of the docs, developers will need to adjust it as approrpriate for their own environmental conditions.

It's just like how we have "mysite" or "app" or "yoursite.com" in various places - those might be different for specific projects, but the developers who have those differences should be aware of them.

sabina-talipova (approved):
Looks good!

PR #197, opened 2023-03-27
DOC Fix some small issues with gql examples

GuySartorelli:
Following on from https://github.com/silverstripe/developer-docs/pull/184 and https://github.com/silverstripe/developer-docs/pull/187, fixes a few small issues I found while updating for CMS 5.

## Issue
- https://github.com/silverstripe/developer-docs/issues/62

PR #187, opened 2023-03-16
DOC Update react injector docs to actually work

GuySartorelli:
These docs weren't updated after some changes were made in admin which override default graphql3 behaviour, which broke basically everything documented here - mostly due to UpperCamelCase fields becoming lowerCamelCase. It also _never_ worked with graphql 4.

Once this is merged, I'll merge it up to the `5.0` docs, and create a PR targetting that to pull out the graphql 3 stuff from it.

## Note
Requires https://github.com/silverstripe/silverstripe-admin/pull/1473 to be merged for any of this to work. Without that PR, neither the originally documented implementation _nor_ the new docs will work.

You can use this documentation to help test/review that PR.

## Issue
- https://github.com/silverstripe/developer-docs/issues/62

Issue #129, opened 2022-12-20
Review all front-end-related documentation


GuySartorelli:
Our front-end documentation is fractured, outdated, and in some cases potentially misleading. It could do with a review.

## Some things to consider:
- https://docs.silverstripe.org/en/4/developer_guides/customising_the_admin_interface/reactjs_redux_and_graphql/#introduction-to-the-react-layer says the react API isn't part of our semver commitment, but that seems to be the only place that mentions it. It certainly isn't mentioned in our major release policy, which is the place where most people will look for that sort of information.
- It would be nice if there was a clearer structure to the doc - currently it seems to find information about javascript development in the CMS we have to traverse all over the place. These should be collected in a single section, ideally (see https://github.com/silverstripe/developer-docs/issues/41)
  - In particular, the react injector documentation is very difficult to parse. It needs to be split into multiple pages. At the very least, there should be one page that focuses on _the API_ of the react injector, and another that provides concrete step-by-step examples of how to use it in a specific narrow-scope project. At the moment those are mixed together, which ultimately means the step-by-step guide is hard to follow at best, and the documentation about the API isn't clear nor complete. See https://github.com/silverstripe/developer-docs/pull/215#issuecomment-1507784574
- A lot of our graphql js stuff is outdated (see https://github.com/silverstripe/developer-docs/issues/62)

PR #215, opened 2023-04-12
DOC Update react injector docs to work

GuySartorelli:
Reapplies the changes to graphql injector docs that were (meant to be) [reverted](https://github.com/silverstripe/developer-docs/pull/199) (along with any new changes I found that were required)

## NOTE
The revert commit was not completed correctly. It was meant to revert the entirety of https://github.com/silverstripe/developer-docs/pull/187 and https://github.com/silverstripe/developer-docs/pull/197, but failed to do so!
That's okay because this commit would have partially reinstated those PRs anyway.

Make sure you test and review _the entirety_ of the graphql/javascript code from `## Using Injector to customise GraphQL queries` to the end of the file. That includes both the graphql 3 _and_ graphql 4 sections! _ALL_ of that is in scope for this PR.

## Requires extra admin PR also
- https://github.com/silverstripe/silverstripe-admin/pull/1498

## Issue
- https://github.com/silverstripe/developer-docs/issues/62

### DIFF
--- a/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md
+++ b/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md
@@ -905,7 +905,6 @@ SilverStripe\GraphQL\Manager:
               read:
                 paginate: false
                 name: readNotes
-              create: true
 ```
 
 [hint]
@@ -920,15 +919,16 @@ Finally, let's make a really simple container app which holds a header and our n
 
 ```js
 import React from 'react';
-import { inject } from 'lib/Injector';
 import Notes from './components/Notes';
 
-const App = ({ ListComponent }) => (
+const App = () => (
   <div>
     <h3>Notes</h3>
     <Notes />
   </div>
 );
+
+export default App;
 ```
 
 **my-module/client/src/index.js**
@@ -955,8 +955,8 @@ Injector.ready(() => {
 
     onunmatch: function() {
       ReactDOM.unmountComponentAtNode(this[0]);
-    }
-  })
+    },
+  });
 });
 ```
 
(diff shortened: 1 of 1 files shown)


sabina-talipova:
Could we also fix this small spelling issue `cient` to `client`.
https://github.com/silverstripe/developer-docs/blob/1a9525dec9596d59090a4f2606e0ee3da0d19b57/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md?plain=1#L1412

And small question: why we use somewhere `app/client` and in another places `my-module/client`?

GuySartorelli:
> why we use somewhere `app/client` and in another places `my-module/client`?

It's to distinguish between the two contexts this documentation is bouncing between - there's a module which provides the notes components and the actual `DataObject` implementation, and then there's some app project which is customising that component by adding the priority field.

Really this whole section of docs needs to be reconsidered, probably being split into a more traditional "here is the API and how it works" and then these examples should be moved into a "how to" section and fleshed out a bit more. But that's not in scope for this PR where I'm just trying to make sure the examples work at all.

sabina-talipova:
> Really this whole section of docs needs to be reconsidered, probably being split into a more traditional "here is the API and how it works" and then these examples should be moved into a "how to" section and fleshed out a bit more. But that's not in scope for this PR where I'm just trying to make sure the examples work at all.

We should, probably, update this section. Since information presented in `## Using Injector to customise GraphQL queries` section is not enough to use it as an example for testing.

sabina-talipova (changes requested):
Could you fix spelling mistake `cient` to `client `
https://github.com/silverstripe/developer-docs/blob/1a9525dec9596d59090a4f2606e0ee3da0d19b57/en/02_Developer_Guides/15_Customising_the_Admin_Interface/07_ReactJS_Redux_and_GraphQL.md?plain=1#L1412

GuySartorelli:
> We should, probably, update this section. Since information presented in `## Using Injector to customise GraphQL queries` section is not enough to use it as an example for testing.

I've made a note about this in https://github.com/silverstripe/developer-docs/issues/129, which is about reviewing the structure and general content of front-end related documentation. It falls outside the scope of _this_ PR, which is really just about making the graphql code examples work - not necessarily about making them clear and easy to understand.

In other words, we'll do this in stages:
- Stage one is getting the code example to work at all. That's this PR
- Stage two is restructuring the doc so that the code example and API documentation are separate, and independently easy to follow. That's https://github.com/silverstripe/developer-docs/issues/129

For the purposes of testing _this_ PR, I'll send you the project I used to do so with information about how to use it for testing this PR.

GuySartorelli:
I've fixed the typo.

sabina-talipova (approved):
LGFM

PR #260, opened 2023-05-04
DOCS: Add note regarding the split of the activedirectory module

madmatt:
Has been split into the saml and ldap modules. Also note that the saml module is not commercially supported.


GuySartorelli (approved):

GuySartorelli:
Thanks for that extra context.

PR #266, opened 2023-05-15
DOC filterAny() many-many aggregate conjunction

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10744

### DIFF
--- a/en/04_Changelogs/5.1.0.md
+++ b/en/04_Changelogs/5.1.0.md
@@ -13,12 +13,15 @@ title: 5.1.0 (unreleased)
 ## Features and enhancements
 
 ### Other new features
+
 - You can now exclude specific `DataObject` models from the check and repair step of `dev/build` - see [ORM Performance](/developer_guides/performance/orm/#skip-check-and-repair) for more information.
 - You can change what `SearchFilter` the `TreeDropdownField` uses with yaml configuration - see [ORM Performance](/developer_guides/performance/orm/#treedropdownfield) for more information.
 - The [`i18nTextCollector`](api:SilverStripe\i18n\TextCollection\i18nTextCollector) now collects strings for ORM properties (e.g. `$db` fields) in `DataObject` and `Extension` classes, and from themes. See [i18n - collecting text](/developer_guides/i18n/#collecting-text) for more details.
 - Extensions which modify permissions for [`Group`](api:SilverStripe\Security\Group) records which return `true` will be respected, the same as when modifying permissions for any other `DataObject` record.
 
 ## Bug fixes
+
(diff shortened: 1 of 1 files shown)


en/04_Changelogs/5.1.0.md
```diff
@@ -13,13 +13,16 @@ title: 5.1.0 (unreleased)
...
 ### Other new features
+
 - You can now exclude specific `DataObject` models from the check and repair step of `dev/build` - see [ORM Performance](/developer_guides/performance/orm/#skip-check-and-repair) for more information.
 - You can change what `SearchFilter` the `TreeDropdownField` uses with yaml configuration - see [ORM Performance](/developer_guides/performance/orm/#treedropdownfield) for more information.
 - The [`i18nTextCollector`](api:SilverStripe\i18n\TextCollection\i18nTextCollector) now collects strings for ORM properties (e.g. `$db` fields) in `DataObject` and `Extension` classes, and from themes. See [i18n - collecting text](/developer_guides/i18n/#collecting-text) for more details.
 - Extensions which modify permissions for [`Group`](api:SilverStripe\Security\Group) records which return `true` will be respected, the same as when modifying permissions for any other `DataObject` record.
 
 ## Bug fixes
 
+- `SilverStripe\ORM\DataList::filterAny()` queries on many-many relations that use an aggregate `HAVING` clause now correctly use an `OR` conjunction rather than an incorrect `AND` conjunction.
```
ChloeSartorelli:
```suggestion
- [`DataList::filterAny()`](api:SilverStripe\ORM\DataList::filterAny()) queries on many-many relations that use an aggregate `HAVING` clause now correctly use an `OR` conjunction rather than an incorrect `AND` conjunction.
```

GuySartorelli (changes requested):

en/04_Changelogs/5.1.0.md
emteknetnz: Updated

GuySartorelli (commented):
Is it worth calling out the fact that you can now choose whether a subquery uses `WHERE` or `HAVING`?

emteknetnz:
> Is it worth calling out the fact that you can now choose whether a subquery uses WHERE or HAVING?

I don't think so, seems like the subquery is a little piece of the ORM that isn't intended for anyone else to really interact with. It's just a think that's used to construct complex SQL statements behind the scenes

GuySartorelli (approved):

GuySartorelli:
I don't necessarily agree that it's not intended to be interacted with - I've used it in the past on various projects that needed deep ORM interactions. But I take the point that it's not a common use case. I'm okay with leaving that out.

Issue #216, opened 2023-04-12
CMS 5 Doc Review - Configuration


maxime-rainville:
Review tho [Forms section](https://docs.silverstripe.org/en/5/developer_guides/configuration/)

## Acceptance criteria
- All CMS 3 and CMS 4 examples are removed/updated for CMS 5 doc
- CMS 5 doc is reviewed with a focus on area with substantial change
  - including validate javascript examples still work

## Parent EPIC
https://github.com/silverstripe/developer-docs/issues/212

## PRs
- https://github.com/silverstripe/developer-docs/pull/268

Issue #212, opened 2023-04-12
EPIC Review CMS5 doc for truthfulness


maxime-rainville:
This was [initially planned as a single card](https://github.com/silverstripe/developer-docs/issues/164), but it turned out to be a bit toot much to chew in one go.

## Acceptance criteria
- All CMS 3 and CMS 4 examples are removed/updated for CMS 5 doc
- CMS 5 doc is reviewed with a focus on area with substantial change
  - including validate all code examples still work

## SCOPE
The primary purpose of this PR is to ensure that:
- code examples actually work in CMS 5
- written documentation is factually correct

It is okay to also augment the existing documentation to make it clearer or to add information you feel is missing, but don't focus on that.

It is also okay to updated code examples or written documentation to better reflect best practices - but that is not the main purpose of this PR. Don't explicitly look for it, but it's okay to fix it where you happen to notice it, if you want to.

Peer review should not suggest changes which focus on best practices, grammatical/spelling problems, or unclear language except where the content they're making the suggestion for is _new_ in the PR. There will be separate best-practices and grammar/spelling reviews of the docs as a whole following this work.

### EXPLICITLY OUT OF SCOPE
- Lessons (references to lesson will be removed)
- Not having namespaces (unless adding these is necessary for code to be clear or correct)
- Userhelp docs
- Docs not following best practices, but still factually correct (except where the practice being promoted is so bad it could introduce security issues - I saw some of those lingering around)
- Bad formatting

GuySartorelli:
We didn't get around to these and aren't able to spend time reviewing the entirety of docs anymore. Will close all of these.

PR #268, opened 2023-05-16
DOCS Update configuration docs

GuySartorelli:
## Issue
- https://github.com/silverstripe/developer-docs/issues/216

### DIFF
--- a/en/02_Developer_Guides/04_Configuration/00_Configuration.md
+++ b/en/02_Developer_Guides/04_Configuration/00_Configuration.md
@@ -24,175 +24,163 @@ For providing content editors or CMS users a place to manage configuration see t
 
 ## Configuration Properties
 
-Configuration values are static properties on any Silverstripe CMS class. These should be at the top of the class and 
-marked with a `@config` docblock. The API documentation will also list the static properties for the class. They should
-be marked `private static` and follow the `lower_case_with_underscores` structure.
+Configuration values are private static properties on any PHP class. These should be at the top of the class.
+The typical naming convention for configuration properties is `lower_case_with_underscores`.
+
+Any class defining configuration properties should use the [`Configurable`](api:SilverStripe\Core\Config\Configurable) trait.
 
 **app/src/MyClass.php**
-
-
-```php
-class MyClass extends Page 
-{
-
-    /**
-     * @config
-     */
-    private static $option_one = true;
-
-    /**
-     * @config
-     */
-    private static $option_two = [];
-}
-
-```
-
-## Accessing and Setting Configuration Properties
-
-This can be done by calling the static method [Config::inst()](api:SilverStripe\Core\Config\Config::inst()), like so:
-
-
-```php
-$config = Config::inst()->get('MyClass', 'property');
-```
-
-Or through the `config()` method on the class.
-
-```php
-$config = $this->config()->get('property');
-```
-
-You may need to apply the [Configurable](api:SilverStripe\Core\Config\Configurable) trait in order to access the `config()` method.
-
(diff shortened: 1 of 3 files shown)


en/02_Developer_Guides/04_Configuration/00_Configuration.md
```diff
@@ -24,175 +24,163 @@ For providing content editors or CMS users a place to manage configuration see t
...
-Note that by default `Config::inst()` returns only an immutable version of config. Use `Config::modify()`
+Note that by default `Config::inst()` returns only an immutable version of config. Use [`Config::modify()`](api:SilverStripe\Core\Config\Config::modify())
 if it's necessary to alter class config. This is generally undesirable in most applications, as modification
 of the config can immediately have performance implications, so this should be used sparingly, or
 during testing to modify state.
 
-Note that while both objects have similar methods the APIs differ slightly. The below actions are equivalent:
+If you do need to modify configuration at run time, you can do so using the following API:
+
+* `Config::modify()->merge(MyClass::class, 'property', 'newvalue');` or `MyClass::config()->merge('property', 'newvalue')` - merges the new configuration value in with any pre-existing values for this property
```
emteknetnz, line 97: I don't think we should be including both ways of doing them when they're equivalent. 

While it's OK to mention at the start there are 2x different way so to do it, `MyClass::config()->merge('property', 'newvalue')` is probably the better of the two since it's shorter, and reinforce s the fact you need to have the `Configurable` trait applied to the class in order for config to work.

Therefore could you please remove any references to `Config::modify()->merge(MyClass::class, 'property', 'newvalue');` style setting here and in the rest of this doc

en/02_Developer_Guides/04_Configuration/00_Configuration.md
```diff
@@ -24,175 +24,163 @@ For providing content editors or CMS users a place to manage configuration see t
...
+    Name: arrayreset
+    ---
+    Class\With\Array\Config:
+    an_array: null
+    ---
+    Name: array
+    ---
+    Class\With\Array\Config:
+    an_array: ['value_a', 'value_b']
+    ```
```
emteknetnz:
```suggestion
    ```yml
    ---
    Name: arrayreset
    ---
    Class\With\Array\Config:
      an_array: null
    ---
    Name: array
    ---
    Class\With\Array\Config:
      an_array: ['value_a', 'value_b']
    ```
```

emteknetnz (changes requested):

en/02_Developer_Guides/04_Configuration/00_Configuration.md
ChloeSartorelli: This goes into "best practices" which will be covered in https://github.com/silverstripe/developer-docs/issues/269 later on. For now, both of these are valid ways to access the config and both work just fine. The scope of this PR and the issue it relates to is primarily "are the docs factually correct?"

Given that these examples already existed and I've just moved them and clarified them slightly, this suggestion goes out of scope for this PR.

See https://github.com/silverstripe/developer-docs/issues/212 regarding scope.

ChloeSartorelli: Also, config _does_ work for classes that don't have the `Configurable` trait. It's _recommended_ to have that trait, but it's not a requirement.

en/02_Developer_Guides/04_Configuration/00_Configuration.md
ChloeSartorelli: Done

emteknetnz (approved):

PR #270, opened 2023-05-17
DOC fix partial caching config example

jonom:
Just realised that I have actually **not** been increasing the default cache lifetime on the sites I've built over the past several years 🙃

### DIFF
--- a/en/02_Developer_Guides/08_Performance/01_Caching.md
+++ b/en/02_Developer_Guides/08_Performance/01_Caching.md
@@ -154,6 +154,10 @@ you need to be careful with resources here (e.g. filesystem space).
 
 
 ```yml
+---
+Name: my-project-cache
+After: '#corecache'
+---
 SilverStripe\Core\Injector\Injector:
   Psr\SimpleCache\CacheInterface.cacheblock:
     constructor:


en/02_Developer_Guides/08_Performance/01_Caching.md
```diff
@@ -154,6 +154,10 @@ you need to be careful with resources here (e.g. filesystem space).
 
 
 ```yml
+---
+Name: my-project-cache
+After: '*#corecache'
```
ChloeSartorelli:
```suggestion
After: '#corecache'
```
The wildcard is implicit - no need to include it in the example as it may lead people to think there's more going on than there actually is.

GuySartorelli (changes requested):
Thank you, good catch! Just one small change.

en/02_Developer_Guides/08_Performance/01_Caching.md
jonom: Thanks. TBH I didn't know what the asterisk or `#` did, I copied the example [from here](https://forum.silverstripe.org/t/increasing-partial-cache-lifetime/1331/4?u=jonom). Just as a sanity check - these are all equivalent?

```
After:
  - 'corecache'
```
```
After: 'corecache'
```
```
After: '#corecache'
```
```
After: 'framework#corecache'
```

ChloeSartorelli: Technically that last one would only be before a config section named `corecache` in a module named framework, but otherwise yes.

GuySartorelli (approved):

PR #287, opened 2023-06-16
Local module instructions

catharsisjelly:
I wanted to make a module and it was not clear to me how to include a local module while I am developing it. I got some help from the Slack channel and thought I would include it in the docs.


emteknetnz:
This seems like it's documenting how to use composer rather than anything Silverstripe specific

PR #293, opened 2023-07-03
DOC Update i18n

AntonyThorpe:
Removed references to Zend_Translate now that the symfony/translate library is being used


GuySartorelli (approved):
Thanks, good spotting

Issue #278, opened 2023-06-06
Move CMS 4 supported module list to developer doc to align with CMS 5


maxime-rainville:
We still have a list of supported module for CMS 4 here: https://www.silverstripe.org/software/addons/silverstripe-commercially-supported-module-list/

## Acceptance criteria
- An equivalent page is created in the CMS 4 doc.
- The current DotORG page redirects to the doc (don't include a version number in the redirection link - that will ensure it directs to the most recent version).
- The CMS 4 and CMS 5 supported modules doc pages link to each other and make explicit that module support will vary from version to version.

## PRs
- https://github.com/silverstripe/developer-docs/pull/296
- https://github.com/silverstripe/developer-docs/pull/297

GuySartorelli:
Doc merged, no remaining ACs

GuySartorelli:
Oops, there is a remaining AC. Reassigning to @emteknetnz to change page on .org to redirect to docs

emteknetnz:
Have changed https://www.silverstripe.org/software/addons/silverstripe-commercially-supported-module-list to a redirector page

GuySartorelli:
Perfect. _now_ all ACs are done. Closing.

PR #297, opened 2023-07-05
DOC Add link to CMS 4 supported modules

emteknetnz:
Issue https://github.com/silverstripe/developer-docs/issues/278


GuySartorelli:
I'm going to close this PR on the basis that we don't tend to say "this is the CMS 5 docs, click here for CMS 4" on any page. There's already a banner mechanism for that.

emteknetnz:
There's an AC for this

The CMS 4 and CMS 5 supported modules doc pages link to each other and **make explicit that module support will vary from version to version.**

PR #309, opened 2023-07-25
update PHP requirements

xini:
SS is compatible with PHP 8.2, right?


GuySartorelli:
Hi, thanks for taking the time to submit this change.

`<= 8.2` already denotes that it's php 8.2 compatible (i.e. less than _or equal to_ 8.2) - and it also explicitly means when the next php version comes out we aren't saying it's compatible until we have validated that it is compatible. We'd prefer to keep being explicit about the compatibility, so I'll close this pr.

PR #310, opened 2023-07-30
[CVE-2023-32302] Require password field to be non-empty

emteknetnz:


GuySartorelli (approved):
Matches original PR. Feels a bit weird merging this in now - but we're releasing this fix and patching it today so it makes sense to merge this now, and it'll appear in the actual per-commit changelog so probably still makes sense to include this. We can change wording later to reflect the fact it was already patched before the 5.1 release.

PR #311, opened 2023-07-31
DOC Warnings about searchable_fields content

sabina-talipova:
## Parent issue
- https://github.com/silverstripe/silverstripe-framework/issues/10761

### DIFF
--- a/en/02_Developer_Guides/00_Model/11_Scaffolding.md
+++ b/en/02_Developer_Guides/00_Model/11_Scaffolding.md
@@ -70,9 +70,12 @@ The `$searchable_fields` property uses a mixed array format that can be used to 
 system. The default is a set of array values listing the fields.
 
 [info]
-`$searchable_fields` will default to use the [`$summary_fields` config](#summary-fields) if not defined. This works fine unless 
-your `$summary_fields` config specifies fields that are not stored in the database.
+`$searchable_fields` will default to use the [`$summary_fields` config](#summary-fields), excluding anything that isn't a database field (such as method calls) if not explicitly defined.
 [/info]
+
+[warning]
+If you define a `searchable_fields` configuration, _do not_ specify fields that are not stored in the database (such as methods), as this will cause an error.
+[/warning]
 
 ```php
 use SilverStripe\ORM\DataObject;


en/02_Developer_Guides/00_Model/11_Scaffolding.md
```diff
@@ -70,8 +70,7 @@ The `$searchable_fields` property uses a mixed array format that can be used to
 system. The default is a set of array values listing the fields.
 
 [info]
-`$searchable_fields` will default to use the [`$summary_fields` config](#summary-fields) if not defined. This works fine unless 
-your `$summary_fields` config specifies fields that are not stored in the database.
+`$searchable_fields` will default to use the [`$summary_fields` config](#summary-fields) if not defined. Do not specify fields that are not stored in the database in `$searchable_fields` since it can invoke error.
```
ChloeSartorelli: I think this is actually more confusing than the old message.

The old message warns that `searchable_fields` will fall back to `summary_fields` - and that if there are non-database-fields in `summary_fields`, that _fallback_ scenario will fail.

The new message warns that `searchable_fields` will fall back to `summary_fields`, but then tells developers not to put non-database-fields in `searchable_fields`.... which they're _not_ doing if they're relying on this fallback behaviour.

It feels like maybe the intention is to keep the old message as it is, but to also add a _separate_ warning against putting non-database-fields in `searchable_fields`?

GuySartorelli (changes requested):

en/02_Developer_Guides/00_Model/11_Scaffolding.md
sabina-talipova: Since the goal of this PR is to exclude from `searchableFields` those fields that are in `summary_fields` that pass the `hasMethod` check, but do not pass the check for the presence of fields in the DB. Following this logic, now it will not matter whether auxiliary methods like `CMSThumbnail` or `Count` are specified in `summary_fields`, as they will be excluded from `searchable_fields`. Perhaps we could add an appropriate check for `searchable_fields` too to avoid errors. But in any case, the user should be warned.

sabina-talipova: Updated. Try to make it more clear.

en/02_Developer_Guides/00_Model/11_Scaffolding.md
```diff
@@ -70,8 +70,7 @@ The `$searchable_fields` property uses a mixed array format that can be used to
 system. The default is a set of array values listing the fields.
 
 [info]
-`$searchable_fields` will default to use the [`$summary_fields` config](#summary-fields) if not defined. This works fine unless 
-your `$summary_fields` config specifies fields that are not stored in the database.
+- `$searchable_fields` will default to use the [`$summary_fields` config](#summary-fields) if not defined. If you define a `searchable_fields` configuration do not specify fields that are not stored in the database, as this will cause an error.
```
ChloeSartorelli:
```suggestion
`$searchable_fields` will default to use the [`$summary_fields` config](#summary-fields), excluding anything that isn't a database field (such as method calls) if not explicitly defined.
```
Lets keep this info block just about the fallback behaviour, and create a new _warning_ block (below) about not including db fields in `searchable_fields`.

en/02_Developer_Guides/00_Model/11_Scaffolding.md
```diff
@@ -70,8 +70,7 @@ The `$searchable_fields` property uses a mixed array format that can be used to
...
 [/info]
```
ChloeSartorelli, line 75:
```suggestion

[warning]
If you define a `searchable_fields` configuration, _do not_ specify fields that are not stored in the database (such as methods), as this will cause an error.
[/warning]

```
Lets move this out into its own dedicated _warning_ block - it's more important that people pay attention to this, than to the information about which is just about the fallback behaviour.

GuySartorelli (changes requested):

en/02_Developer_Guides/00_Model/11_Scaffolding.md
sabina-talipova: Updated

en/02_Developer_Guides/00_Model/11_Scaffolding.md
sabina-talipova: Updated

ChloeSartorelli: You've still left the `-` at the start -  see https://github.com/silverstripe/developer-docs/pull/311#discussion_r1282592164

en/02_Developer_Guides/00_Model/11_Scaffolding.md
```diff
@@ -70,10 +70,13 @@ The `$searchable_fields` property uses a mixed array format that can be used to
 system. The default is a set of array values listing the fields.
 
 [info]
-`$searchable_fields` will default to use the [`$summary_fields` config](#summary-fields) if not defined. This works fine unless 
-your `$summary_fields` config specifies fields that are not stored in the database.
+- `$searchable_fields` will default to use the [`$summary_fields` config](#summary-fields), excluding anything that isn't a database field (such as method calls) if not explicitly defined.
```
ChloeSartorelli:
```suggestion
`$searchable_fields` will default to use the [`$summary_fields` config](#summary-fields), excluding anything that isn't a database field (such as method calls) if not explicitly defined.
```

GuySartorelli (changes requested):

en/02_Developer_Guides/00_Model/11_Scaffolding.md
sabina-talipova: Updated.

GuySartorelli (approved):

PR #332, opened 2023-08-23
DOC Correct hyperlink

n8-dev:
Specifically on `/getting_started/environment_management/`.

Was 404ing


GuySartorelli (approved):
Thanks! Good catch

Issue #338, opened 2023-08-30
Add silverstripe/dynamodb as a supported module for CMS 5


GuySartorelli:
[silverstripe/dynamodb](https://github.com/silverstripe/silverstripe-dynamodb/) is required for at least one of silverstripe cloud and silverstripe cloud public sector.
It should be commercially supported - and it's already in kitchen sink.

## Acceptance criteria
- silverstripe/dynamodb is listed as commercially supported in the docs for cms 5
- silverstripe/dynamodb is added to the supported-modules repo if it's not there already
- Validate that silverstripe/dynamodb is in the sink and gets picked up when we tag releases.
 
## PRs
- https://github.com/silverstripe/supported-modules/pull/25
- https://github.com/silverstripe/supported-modules/pull/26
- https://github.com/silverstripe/developer-docs/pull/342
- https://github.com/silverstripe/developer-docs/pull/343

emteknetnz:
Have confirmed that dynamodb will show up in a new cow plan

![image](https://github.com/silverstripe/developer-docs/assets/4809037/25d3855e-83d0-489f-982e-50acebd0340d)

GuySartorelli:
PRs merged

PR #342, opened 2023-09-04
DOC Add dynamodb

emteknetnz:
Issue https://github.com/silverstripe/developer-docs/issues/338


GuySartorelli (approved):
Will have merge conflicts with the CMS 4 PR, but it does need to be different, and this repo has merge conflicts already so fine to merge now and we can sort out all the merge conflicts at the same time.

LGTM

PR #344, opened 2023-09-04
DOCS Add missing word "guide" to "User Help guide"

GuySartorelli:
This is how it is described on the first page of the user help guide, and is a clearer name than simply "User Help".


GuySartorelli:
This was really just for demonstrative purposes and isn't important enough to leave open.

PR #345, opened 2023-09-07
Updated with database column type override info

tiller1010:
Added information to override database column types with the yml configuration syntax.

### DIFF
--- a/en/02_Developer_Guides/05_Extending/01_Extensions.md
+++ b/en/02_Developer_Guides/05_Extending/01_Extensions.md
@@ -149,8 +149,14 @@ App\Extensions\MyDataClassConfigExtension:
 ```
 
 [notice]
-Note that the value for `key1` in the `my_configuration_property` array way _not_ overridden by the extension class. Configuration declared in an extension class is merged into the base class as a lower priority than the base class itself. Where there is any collision between configuration declared on the base class and on the extension class, the base class configuration is used.
+Note that the value for `key1` in the `my_configuration_property` array was _not_ overridden by the extension class. Configuration declared in an extension class is merged into the base class as a lower priority than the base class itself. Where there is any collision between the configuration declared on the base class and on the extension class, the base class configuration is used. To workaround this, use the .yml configuration because $db, $has_one, etc are ultimately just configuration properties.
 [/notice]
+
+```yml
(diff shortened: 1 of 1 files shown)


en/02_Developer_Guides/05_Extending/01_Extensions.md
```diff
@@ -149,9 +149,15 @@ App\Extensions\MyDataClassConfigExtension:
...
 [notice]
-Note that the value for `key1` in the `my_configuration_property` array way _not_ overridden by the extension class. Configuration declared in an extension class is merged into the base class as a lower priority than the base class itself. Where there is any collision between configuration declared on the base class and on the extension class, the base class configuration is used.
+Note that the value for `key1` in the `my_configuration_property` array was _not_ overridden by the extension class. Configuration declared in an extension class is merged into the base class as a lower priority than the base class itself. Where there is any collision between the configuration declared on the base class and on the extension class, the base class configuration is used. To workaround this, use the .yml configuration because $db, $has_one, etc are ultimately just configuration properties.
 [/notice]
 
+```yml
+App\Data\MyDataClass:
+  my_configuration_property:
+    key1: 'is overridden'
+```
```
ChloeSartorelli:
````suggestion
Note that the value for `key1` in the `my_configuration_property` array was _not_ overridden by the extension class. Configuration declared in an extension class is merged into the base class as a lower priority than the base class itself. Where there is any collision between the configuration declared on the base class and on the extension class, the base class configuration is used.

If you need to override values, you should do so using the [yml configuration API](/developer_guides/configuration/configuration).

```yml
App\Data\MyDataClass:
  my_configuration_property:
    key1: 'is overridden'
```
[/notice]
````

The code example is part of the notice, so please move it inside the notice block.

Also reworded slightly - this suggestion isn't a workaround to a limitation so much as it is the correct way to override configuration values.

Also removed "because $db, $has_one, etc are ultimately just configuration properties" because that doesn't seem relevant here. The example doesn't show or reference any of those. If you want to reference those, there's probably a more appropriate place than within this notice block.

GuySartorelli (changes requested):
Thank you for submitting this. I've requested some changes below.

GuySartorelli (approved):
Thanks for this. Looks good to me.

PR #353, opened 2023-09-26
DOC Move orderBy docs to raw SQL section.

GuySartorelli:
We shouldn't encourage using `orderBy()` for random sorting - that's what the `shuffle()` abstraction is for.
Instead, don't mention `orderBy()` at all until the dedicated raw SQL section.

## Issue
- https://github.com/silverstripe/developer-docs/issues/354

### DIFF
--- a/en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
+++ b/en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
@@ -303,16 +303,11 @@ $players = Players::get()->sort([
 ]);
 ```
 
-You can also sort randomly. Using the `DB` class, you can get the random sort method per database type.
+You can also sort randomly.
 
 ```php
-$random = DB::get_conn()->random(); 
-$players = Player::get()->orderBy($random);
+$players = Player::get()->shuffle();
 ```
-
-[warning]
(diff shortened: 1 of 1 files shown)


en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
```diff
@@ -303,17 +303,12 @@ $players = Players::get()->sort([
...
 
 ```php
-$random = DB::get_conn()->random(); 
-$players = Player::get()->orderBy($random);
+$players = Player::get()->shuffle();
 ```
 
-[warning]
-Note that we've used the `orderBy()` method here. This is because `sort()` doesn't allow sorting by raw SQL, which is necessary to use a random sort. Be careful whenever you use the `orderBy()` method to ensure you don't pass in any values you can't trust.
-[/warning]
```
sabina-talipova, line 315: Small question. Should we completely remove this statement or we could also move it in "Order by" section?

ChloeSartorelli: Most of this is incorrectly saying you _need_ to use raw SQL to do a random sort, but that's what the `shuffle()` method is for.
The rest of this is saying that raw SQL is inherently risky, which is already stated in the raw SQL section.

Basically, I don't think there's anything here that we need to keep.

sabina-talipova (approved):
Looks good to me!

Issue #363, opened 2023-10-06
Small doc problems to be fixed


GuySartorelli:
I've found a few minor problems in the docs. Capturing the PRs for them all in this card.

## PRs
- https://github.com/silverstripe/developer-docs/pull/364
- https://github.com/silverstripe/developer-docs/pull/365
- https://github.com/silverstripe/developer-docs/pull/366

sabina-talipova:
PRs merged.

PR #365, opened 2023-10-06
DOC Fix API links on fluent upgrade page

GuySartorelli:
These API links aren't working because the slashes are backwards.

## Issue
- https://github.com/silverstripe/developer-docs/issues/363


sabina-talipova (approved):
Looks good to me.

PR #366, opened 2023-10-06
DOC Make error log file path absolute

GuySartorelli:
The error log file path must be absolute as per https://docs.silverstripe.org/en/developer_guides/debugging/error_handling/#logging-to-a-file

## Issue
- https://github.com/silverstripe/developer-docs/issues/363


sabina-talipova (approved):
Looks good to me.

PR #374, opened 2023-10-12
DOC Create new users with random passwords

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10940

### DIFF
--- a/en/04_Changelogs/5.2.0.md
+++ b/en/04_Changelogs/5.2.0.md
@@ -9,6 +9,7 @@ title: 5.2.0 (unreleased)
 - [Features and enhancements](#features-and-enhancements)
   - [New ORM features](#new-orm-features)
   - [ErrorPage allowed codes configuration](#errorpage-allowed-codes-configuration)
+  - [Create random passwords for new users](#create-random-passwords-for-new-users)
   - [Other new features](#other-new-features)
 - [API changes](#api-changes)
 - [Bug fixes](#bug-fixes)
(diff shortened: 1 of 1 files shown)


en/04_Changelogs/5.2.0.md
ChloeSartorelli, line 1: We should reference the fact that this changed in 5.1 for security reasons, and that now we're changing it again in a way that opens up old use cases (i.e. creating a user without knowing their password) without resurfacing the original vulnerability.

en/04_Changelogs/5.2.0.md
```diff
@@ -87,7 +88,13 @@ SilverStripe\ErrorPage\ErrorPage:
...
 ```
 
-### Other new features
+### Create new users with random passwords
+
+If a new user is created in the security section of the CMS with an empty password, then a strong random password will now be automatically assigned to the user which will not be visible to the person creating the user. The user will need to click the "I've forgotten my password" link in order to receive a password reset email.
+
+This only applies to users created through the security section of the CMS. Users created programatically with an empty password will not have a random password automatically assigned to them. This is the current behaviour and is often used for non-production purposes such as unit-testing.
+
+Note that it is not possible to log in to the CMS using an empty password with the default Silverstripe authenticator.
```
ChloeSartorelli: > Note that it is not possible to log in to the CMS using an empty password with the default Silverstripe authenticator.

I don't think this note is needed here, especially after https://github.com/silverstripe/developer-docs/pull/374#discussion_r1357600572

GuySartorelli (changes requested):

en/04_Changelogs/5.2.0.md
emteknetnz: Have reworked this and said that you could set an emtpy password just not log in with it when running < 5.1

en/04_Changelogs/5.2.0.md
emteknetnz: Updated

en/04_Changelogs/5.2.0.md
```diff
@@ -87,7 +88,13 @@ SilverStripe\ErrorPage\ErrorPage:
     - 500
 ```
 
-### Other new features
```
ChloeSartorelli, line 90: Don't remove this heading - we'll need this for other changes, and it's still referenced in the table of contents.

GuySartorelli (changes requested):

en/04_Changelogs/5.2.0.md
emteknetnz: Added back in

GuySartorelli (approved):
LGTM

Issue #11, opened 2022-07-15
Add linting for docs


GuySartorelli:
We don't currently have any automated checks to validate the quality of our documentation. This means we might potentially have now - or could easily introduce in the future - markdown formatting issues, incorrectly formatted links, etc. This is especially likely to happen with some of the custom markdown syntax such as API links.

## Proposed automated checks:

- Markdown linting
  - e.g. all code blocks have a syntax-highlighting hint (use `text` for plain text blocks)
- Ensure custom functionality is formatted correctly
  - links (e.g. api links use the `api:Namespace\Class::someMethod()` syntax instead of a hardcoded links, links to other docs sections are relative)
  - `[info]` and similar are formatted correctly and are never left unclosed
  - Custom anchors (i.e. `{#some-anchor}` in headings)
  - no malformed title, description, icons
    - e.g. backticks in these causes errors in the netlify build
  - Probably others (part of the work will be looking at what functionality docs.silverstripe.org has and adding to this list)
- Changelog marked as "unreleased" (in title and header) unless it has an actual commit changelog included
- code blocks should be standardised (e.g. allow either `yaml` or `yml` but not both)
- Standardised `namespace` declarations in php code examples
- Any broken links are identified (e.g. at a minimum this would be broken relative links to other doc pages - but could expand to include absolute URLs and api docs)
- Either enforce a character limit for lines of documentation (our guide says "keep documentation lines shorter than 120 characters" at the time of writing) OR enforce that paragraphs must not be split into multiple lines.
- Everything in our [contributing documentation](https://docs.silverstripe.org/en/contributing/documentation/) page that can be linted should be linted.
- Code inside code blocks for at least PHP and Javascript are linted

This should be run on all PRs automatically via github actions.

NOTE: Linting should respect Silverstripe's [writing style guide](https://silverstripe.atlassian.net/wiki/spaces/MAR/pages/506364107/Silverstripe+Writing+Style+Guide) (only accessible to Silverstripe employees currently)

## potentially useful tools
- https://vale.sh/ for linting docs generally
- [various suggestions](https://earthly.dev/blog/markdown-lint/) which can work in combination to fulfill various requirements
- [eslint-plugin-markdown](https://github.com/eslint/eslint-plugin-markdown) for linting js examples in markdown
- [DavidAnson/markdownlint](https://github.com/DavidAnson/markdownlint) or [markdownlint/markdownlint](https://github.com/markdownlint/markdownlint) for linting the markdown itself
[...]

PR #387, opened 2023-10-29
MNT Add linting

GuySartorelli:
## NOTES
- Some of the changes required to pass linting PHP code has involved things like adding namespaces, which might affect the context of documentation surrounding that block (e.g. yaml configuration may reference the classname, and needs the full namespace). I've done my best to catch all of those scenarios, but please make sure you're looking for those sorts of scenarios during review.
- There were inconsistencies in how we declared the filename for a code block, and the most common way failed linting. I've unified those to now be included as a comment in the first line of the codeblock itself. I've made an edit to include this as one of our documentation standards going forward.
- I have explicitly disabled the rule which limits the number of characters per line as it simply touches too much documentation to do in this pass. This will be hard enough to merge-up as it is. After this has been merged in and up, I'll create a separate PR that _only_ applies that rule.

**DO NOT SQUASH COMMITS**

## Issue
- https://github.com/silverstripe/developer-docs/issues/11


package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "docs",
+  "scripts": {
+    "lint": "yarn install && yarn lint-md && yarn lint-js && yarn lint-php && yarn lint-language",
+    "lint-fix": "yarn install && yarn lint-md --fix; yarn lint-js --fix; yarn lint-php --fix; echo 'Fixed all auto-fixable problems'",
+    "lint-md": "yarn install && markdownlint-cli2 en/**/*.md",
+    "lint-js": "yarn install && eslint en",
+    "lint-php": "composer install --prefer-dist --no-progress --ansi --no-interaction --no-scripts --no-plugins --optimize-autoloader && vendor/bin/mdphpcs en",
+    "lint-language": "yarn install && vale sync && vale en"
```
ChloeSartorelli: The individual scripts are used by CI so we can clearly see what is failing and where, without having either a massive hard-to-parse output or a failure in one script which results in skipping the rest of the linting.

The main `lint` script is useful to run locally to double check all is well before pushing changes.

The `lint-fix` script is useful to run locally to fix anything that _can_ be autofixed, without really caring what script is fixing it.

emteknetnz:
In the files change tab at the bottom there a "Unchanged files with check annotations" - any idea if that's something that can be configured off at the github actions level, or if it's a user preference thing?

composer.json
```diff
@@ -19,5 +19,14 @@
             "name": "The SilverStripe Community",
             "homepage": "http://silverstripe.org"
         }
+    ],
+    "require-dev": {
+        "silverstripe/markdown-php-codesniffer": "dev-pulls/1/new-repo"
```
emteknetnz: Will need to update this to use a non fork branch

package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "docs",
+  "scripts": {
+    "lint": "yarn install && yarn lint-md && yarn lint-js && yarn lint-php && yarn lint-language",
+    "lint-fix": "yarn install && yarn lint-md --fix; yarn lint-js --fix; yarn lint-php --fix; echo 'Fixed all auto-fixable problems'",
+    "lint-md": "yarn install && markdownlint-cli2 en/**/*.md",
+    "lint-js": "yarn install && eslint en",
+    "lint-php": "composer install --prefer-dist --no-progress --ansi --no-interaction --no-scripts --no-plugins --optimize-autoloader && vendor/bin/mdphpcs en",
```
emteknetnz:
```suggestion
    "lint-php": "vendor/bin/mdphpcs en",
```

Seems super weird running `composer install` here.  Realistically there's only a few people who will ever run this on their locally so there's no need to be "helpful"

package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "docs",
+  "scripts": {
+    "lint": "yarn install && yarn lint-md && yarn lint-js && yarn lint-php && yarn lint-language",
+    "lint-fix": "yarn install && yarn lint-md --fix; yarn lint-js --fix; yarn lint-php --fix; echo 'Fixed all auto-fixable problems'",
+    "lint-md": "yarn install && markdownlint-cli2 en/**/*.md",
+    "lint-js": "yarn install && eslint en",
```
emteknetnz:
```suggestion
    "lint": "yarn lint-md && yarn lint-js && yarn lint-php && yarn lint-language",
    "lint-fix": "yarn lint-md --fix; yarn lint-js --fix; yarn lint-php --fix; echo 'Fixed all auto-fixable problems'",
    "lint-md": "markdownlint-cli2 en/**/*.md",
    "lint-js": "eslint en",
```
We don't run `yarn install` in any of our other package.json scripts so shouldn't here either.  There's only a few people who will ever run these scripts and we all have a general understanding that you need to `yarn install` before you do things

package.json
emteknetnz:
```suggestion
    "lint-language": "vale sync && vale en"
```

emteknetnz: What does `vale sync` do?

.vale/styles/Silverstripe-cms/Repetition.yml
```diff
@@ -0,0 +1,12 @@
...
+#
+# Checks for duplicate words, like `the the` or `and and`.
+#
+# https://gitlab.com/gitlab-org/gitlab/-/blob/master/doc/.vale/gitlab/Repetition.yml
+extends: repetition
+message: '"%s" is repeated.'
+level: error
+alpha: true
+tokens:
+  - '[^\s]+'
```
emteknetnz: Add blank at end of file

.vale/styles/Silverstripe-cms/Race.yml
```diff
@@ -0,0 +1,18 @@
...
+ignorecase: true
+level: warning
+action:
+  name: replace
+swap:
+  master copy: pass key|original
+  master key: pass key|original
+  master plan: grand scheme|guiding principles
+  master the art: become skilled
+  master: main|primary|hub|reference
```
emteknetnz: We've got plenty of references to master / dev-master branches including in thing like the 4.9.0-rc1 changelog, which we shouldn't touch.  I'm not sure if we actually have any master branches left (if we do, probably not to hard to just rename to main? Or just delete them. Separate card though).

Actually on a related topic are we able to exclude any changelogs that have already gone out the door, though apply this linting to changelogs going forward?  Seems quite wrong to me to be altering changelogs after the fact since they're a snapshot.

.vale.ini
```diff
@@ -0,0 +1,11 @@
...
+
+Vocab = Silverstripe
+
+MinAlertLevel = suggestion
+
+; Don't check inside heading anchors, i.e. inside {#my-anchor}
+WordTemplate = "[^#]\b(?:%s)\b[^}]"
+
+[*.md]
+BasedOnStyles = Silverstripe-cms, openly #Vale
```
emteknetnz: Add blank line at end of file

.nvmrc
```diff
@@ -0,0 +1 @@
+18
```
emteknetnz: Add blank line at end of file

.github/workflows/ci.yml
```diff
@@ -0,0 +1,44 @@
...
+        run: npm install --global yarn
+
+      - name: Install PHP
+        if: ${{ matrix.script == 'lint-php' }}
+        uses: shivammathur/setup-php@1a18b2267f80291a81ca1d33e7c851fe09e7dfc4 # v2.22.0
+        with:
+          php-version: 8.1
+
+      - name: Run lint
+        run: yarn ${{ matrix.script }}
```
emteknetnz: Add blank line at end of file

.vale/styles/Silverstripe-cms/Terms.yml
```diff
@@ -0,0 +1,31 @@
...
+  yaml: YAML (or yml for syntax highlighting code blocks)
+  url: URL
+  urls: URLs
+  uri: URI
+  uris: URIs
+  gpu: GPU
+  typescript: TypeScript
+  'getters?': getter method(s)
+  'setters?': setter method(s)
+  'dataobjects?': 'models(s)|record(s) (or wrap it in backticks if referencing `DataObject` specifically)'
```
emteknetnz: Add blank line at end of file

.vale/styles/Silverstripe-cms/Typography.yml
```diff
@@ -0,0 +1,15 @@
...
+swap:
+  '…': '...'
+  '\(c\)': ©
+  '\(tm\)': ™
+  '\(r\)': ®
+  '×': 'x'
+  '[“”„]': '"'
+  '[‘’‚]': "'"
+  '[‹«]': "<"
+  '[›»]': ">"
```
emteknetnz: Add blank line at end of file

.vale/styles/openly/Hedging.yml
```diff
@@ -0,0 +1,29 @@
+---
+extends: existence
+message: "Consider removing '%s'"
+ignorecase: true
+level: warning
+tokens:
+  - appear to be
+  - arguably
+  - as far as I know
+  - could
```
emteknetnz: I can see rules like this just being annoying

There are 154 matches on the word 'could'

.vale/styles/openly/Spelling.yml
```diff
@@ -0,0 +1,15 @@
...
+ignore:
+  - openly/tech-terms/security.txt
+  - openly/tech-terms/blockchain.txt
+  - openly/tech-terms/devops.txt
+  - openly/tech-terms/general.txt
+  - openly/tech-terms/techwriting.txt
+  - openly/tech-terms/mobile.txt
+  - openly/tech-terms/linux.txt
+  - openly/tech-terms/monitoring.txt
+  - openly/tech-terms/apis.txt
```
emteknetnz: Add blank line at end of file

emteknetnz (changes requested):
I'm thinking we should split off the vale language linting into its own card and tackle separately from the non-language stuff like indentation first.

Once that's sorted then look at doing language second because at that point we'll get a mountain of warnings tell us that we should consider replacing words like "simple" and "could"

GuySartorelli:
> In the files change tab at the bottom there a "Unchanged files with check annotations" - any idea if that's something that can be configured off at the github actions level, or if it's a user preference thing?

Looks like that's just something GitHub is doing, which is a pain. There's (a very small amount of) discussion about it at https://github.com/actions/toolkit/issues/457

I think the only reason we don't see this on any PRs in our other repos is because we don't have any pre-existing linting issues in those other repos.

Once we've fixed them all in this PR, we won't see them anymore here either.

composer.json
ChloeSartorelli: Yup, will do that as soon as the PR for that has been merged and the package is added to packagist.

GuySartorelli:
> I'm thinking we should split off the vale language linting into its own card and tackle separately from the non-language stuff like indentation first.

After discussing this offline I agree that it will be easier and faster to review the structural linting first, and deal with the language stuff separately as that could change the meanings of what's written and therefore requires a higher level of scrutiny.

See https://github.com/silverstripe/documentation-lint/issues/14

.vale/styles/Silverstripe-cms/Repetition.yml
ChloeSartorelli: Marking resolved since this is now out of scope.

.vale/styles/Silverstripe-cms/Race.yml
ChloeSartorelli: Marking resolved since this is now out of scope.

.vale.ini
ChloeSartorelli: Marking resolved since this is now out of scope.

.nvmrc
GuySartorelli: Marking resolved since this is now out of scope.

.vale/styles/openly/Hedging.yml
ChloeSartorelli: Marking resolved since this is now out of scope.

.vale/styles/Silverstripe-cms/Typography.yml
ChloeSartorelli: Marking resolved since this is now out of scope.

.vale/styles/Silverstripe-cms/Terms.yml
ChloeSartorelli: Marking resolved since this is now out of scope.

.github/workflows/ci.yml
GuySartorelli: Marking resolved since this is now out of scope.

package.json
ChloeSartorelli: Marking resolved as this is now out of scope

.nvmrc
ChloeSartorelli: Done

.github/workflows/ci.yml
ChloeSartorelli: Done

.vale/styles/openly/Spelling.yml
ChloeSartorelli: Marking resolved since this is now out of scope.

package.json
ChloeSartorelli: Done. I'll add some documentation about how to run linting since it's no longer just "run the lint command"

package.json
ChloeSartorelli: Done. I'll add some documentation about how to run linting since it's no longer just "run the lint command"

I will note though that none of our other package.json scripts are for this purpose - so we don't _have_ to align with how they're set up. With those setups, you'll always have already run `yarn install` before linting because the reason you're linting is because you've made some js or css changes, which isn't the case here. I'm only making this change now to avoid unnecessary PR ping pong.

emteknetnz (changes requested):
Can we exclude old changelogs (but not new ones, maybe do something with `git branch -r`)? There are A LOT of linting issues in old changelogs and to me it seems wrong to update old changelogs since they're a snapshot in time, particularly once we add in the language linting.

GuySartorelli:
Discussed offline - the markdown linting especially should probably be done on changelogs. A lot of the justification for the markdown linting rules is to remove ambiguity that can result in differences between markdown renderers - so if we decide to change what renderer we use to convert the md to html, we won't have unexpected problems rendering these old docs.

We can consider whether the language linting should affect historic changelogs when we deal with the separate issue.

@maxime-rainville can you please just let us know if you're okay with us changing the old changelogs as they pertain to markdown, js, and php linting?

maxime-rainville:
@GuySartorelli I'm absolutely fine with updating old change log. I don't think of them as an "historical record". If something happens later on that alters the context of the changelog, I have no problem with pushing updates to them.

composer.json
ChloeSartorelli: Done

GuySartorelli:
@emteknetnz Looks like you didn't look at this one?

.markdownlint-cli2.mjs
```diff
@@ -0,0 +1,14 @@
+
+// const extraRules = require("extraRules");
```
emteknetnz: Remove

en/00_Getting_Started/02_Composer.md
```diff
@@ -306,7 +306,7 @@ the ["Aliases" chapter of the Composer documentation](http://getcomposer.org/doc
 
 Composer needs hints about the base package version, either by using `composer create-project`
 as described above, or by checking out the `silverstripe-installer` project directly from version control. In order to
-use Composer on archive downloads from silverstripe.org, or other unversioned sources, an advanced workaround is to set
+use Composer on archive downloads from `silverstripe.org`, or other unversioned sources, an advanced workaround is to set
```
emteknetnz: Seems like urls should be hyperlinks (that open in a new window?) rather than surrounded in backticks

en/00_Getting_Started/02_Composer.md
```diff
@@ -324,7 +324,7 @@ on [packagist.org](http://packagist.org).
 
 Follow the packagist.org advice on choosing a [unique name and vendor prefix](https://packagist.org/about). Please don't
 use the `silverstripe/<modulename>` vendor prefix, since that's reserved for modules produced by Silverstripe Ltd. In
-order to declare that your module is in fact a Silverstripe CMS module, use the "silverstripe" tag in the composer.json
+order to declare that your module is in fact a Silverstripe CMS module, use the "Silverstripe" tag in the `composer.json`
```
emteknetnz: Seems like this shouldn't be uppercased in this scenario?

en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
```diff
@@ -698,16 +694,26 @@ time.
 For example, suppose we have the following set of classes:
 
 ```php
-use SilverStripe\CMS\Model\SiteTree;
+namespace {
```
emteknetnz, line 697: These empty namespaces are pointless, can we simply not have them if there isn't one?

en/02_Developer_Guides/00_Model/02_Relations.md
```diff
@@ -448,26 +560,35 @@ class Tag extends DataObject
     /**
      * Example iterator placeholder for belongs_many_many.
      * This is a list of arbitrary types of objects
-     * @return Generator|DataObject[]
+     * @return \Generator|DataObject[]
```
emteknetnz: Is the prefixed slash on unimported global php things in our proper code linting (I don't think it is?) Is it a rule for the docs linting?

en/02_Developer_Guides/00_Model/10_Versioning.md
```diff
@@ -1058,27 +1098,26 @@ SilverStripe\GraphQL\Manager:
...
 
-`silverstripe/versioned` provides some graphql plugins we're taking advantage of here. See [Working with DataObjects - Versioned content](/developer_guides/graphql/working_with_dataobjects/versioning/) for more information.
+`silverstripe/versioned` provides some GraphQL plugins we're taking advantage of here. See [Working with DataObjects - Versioned content](/developer_guides/graphql/working_with_dataobjects/versioning/) for more information.
 
 For this we need one query and one mutation:
 
-**app/client/src/state/readOneMyVersionedObjectQuery.js**
-
 ```js
+// app/client/src/state/readOneMyVersionedObjectQuery.js
```
emteknetnz, line 1120: The more I look at these the less I like having the filename at the top of the file.  A lot of these examples you can just copy paste in, though if I did that I'd then want to delete the thing at the top of my file that has the filename since it's redundant. 

Also .json examples aren't valid because native json doesn't support comments

I'd much prefer to have the filename above the code fence in **bold** is best?

en/02_Developer_Guides/03_Forms/03_Form_Templates.md
```diff
@@ -4,75 +4,73 @@ summary: Customize the generated HTML for a FormField or an entire Form.
...
 ```php
-$field = new TextField();
+$field = TextField::create();
 
 $field->setTemplate('CustomTextField');
 // Sets the template for the <input> tag. i.e '<input $AttributesHTML />'
 
 $field->setFieldHolderTemplate('CustomTextField_Holder');
-// Sets the template for the wrapper around the text field. i.e 
+// Sets the template for the wrapper around the text field. i.e
```
emteknetnz: Should `i.e` be `i.e.` ?

Ditto `e.g` if there's any that haven't been replaced by `for example`

en/02_Developer_Guides/03_Forms/Field_types/03_HTMLEditorField.md
```diff
@@ -411,17 +414,20 @@ so this is considered advanced usage of the field.
...
+use SilverStripe\Forms\HTMLEditor\TinyMCEConfig;
+
+TinyMCEConfig::get('cms')
+    ->disablePlugins('ssbuttons')
+    ->removeButtons('sslink', 'ssmedia');
+    ->addButtonsToLine(2, 'link', 'media');
 ```
 
-### Developing a wrapper to use a different WYSIWYG editors with HTMLEditorField
+### Developing a wrapper to use a different WYSIWYG editors with hTMLEditorField
```
emteknetnz: Shouldn't be lowercased

en/02_Developer_Guides/04_Configuration/01_SiteConfig.md
```diff
@@ -25,68 +24,65 @@ $SiteConfig.Tagline
...
-$config = SiteConfig::current_site_config(); 
+$config = SiteConfig::current_site_config();
 
 echo $config->Title;
 
 // returns "Website Name"
 ```
 
-## Extending SiteConfig
+## Extending siteConfig
```
emteknetnz: Shouldn't be lowercased

en/02_Developer_Guides/05_Extending/01_Extensions.md
```diff
@@ -1,40 +1,39 @@
 ---
 title: Extensions
-summary: Extensions and DataExtensions let you modify and augment objects transparently. 
+summary: Extensions and DataExtensions let you modify and augment objects transparently.
 icon: code
 ---
 
-# Extensions and DataExtensions
+# Extensions and dataExtensions
```
emteknetnz: Shouldn't be lower cased

en/02_Developer_Guides/06_Testing/01_Functional_Testing.md
```diff
@@ -15,32 +15,33 @@ creating [HTTPRequest](api:SilverStripe\Control\HTTPRequest), receiving [HTTPRes
 $page = $this->get($url);
 ```
 
-Performs a GET request on $url and retrieves the [HTTPResponse](api:SilverStripe\Control\HTTPResponse). This also changes the current page to the value
+Performs a GET request on $URL and retrieves the [HTTPResponse](api:SilverStripe\Control\HTTPResponse). This also changes the current page to the value
```
emteknetnz: Shouldn't be capitilised, ditto for a few other examples of this in this file

en/02_Developer_Guides/06_Testing/01_Functional_Testing.md
```diff
@@ -70,68 +70,66 @@ $this->logOut();
...
+Assert that the most recently queried page contains a number of content tags specified by a CSS selector. The given CSS
+selector will be applied to the HTML of the most recent page. The content of every matching tag will be examined. The
 assertion fails if one of the expectedMatches fails to appear.
 
 [notice]
 `&nbsp;` characters are stripped from the content; make sure that your assertions take this into account.
 [/notice]
 
-### assertExactHTMLMatchBySelector
+### AssertExactHTMLMatchBySelector
```
emteknetnz: Shouldn't be capitilised

en/02_Developer_Guides/09_Security/01_Access_Control.md
```diff
@@ -50,17 +51,16 @@ the admin screens work).
 Here are my notes trying to figure this stuff out. Not really useful unless you're VERY interested in how exactly SS
 works.
 
-
 ### Loading the admin page: looking at security
 
-If you go to [your site]/admin *Director.php* maps the 'admin' URL request through a [Director](api:SilverStripe\Control\Director) rule to the
-[CMSMain](api:SilverStripe\CMS\Controllers\CMSMain) controller (see [CMSMain](api:SilverStripe\CMS\Controllers\CMSMain), with no arguments). 
+If you go to [your site]/admin *Director.PHP* maps the 'admin' URL request through a [Director](api:SilverStripe\Control\Director) rule to the
```
emteknetnz: Shouldn't be capitilised

en/02_Developer_Guides/11_Integration/00_CSV_Import.md
```diff
@@ -12,63 +12,67 @@ but this method doesn't know anything about your datamodel. In Silverstripe CMS,
 this can be handled through the a specialized CSV importer class that can
 be customised to fit your data.
 
-## The CsvBulkLoader class
+## The csvBulkLoader class
```
emteknetnz: Shouldn't be capitilised

en/02_Developer_Guides/11_Integration/00_CSV_Import.md
```diff
@@ -151,100 +168,122 @@ Sample CSV Content
...
+    public function __construct($objectClass)
+    {
+        $this->columnMap = [
+            'Number' => 'PlayerNumber',
+            'Name' => '->importFirstAndLastName',
+            'Birthday' => 'Birthday',
+            'Team' => 'Team.Title',
+        ];
+
+        $this->$duplicateChecks = [
```
emteknetnz: This doesn't look right

en/02_Developer_Guides/11_Integration/00_CSV_Import.md
```diff
@@ -151,100 +168,122 @@ Sample CSV Content
...
+            'Number' => 'PlayerNumber',
+        ];
+
+        $this->$relationCallbacks = [
```
emteknetnz: Nor this

en/02_Developer_Guides/11_Integration/02_RSSFeed.md
```diff
@@ -58,48 +57,48 @@ RSSFeed::linkToFeed($link, $title);
...
     {
-        return Page::get()->sort("LastEdited", "DESC")->limit(10);
+        return HomePage::get()->sort('LastEdited', 'DESC')->limit(10);
     }
 }
-
 ```
 
-### Rendering DataObjects in a RSSFeed
+### Rendering dataObjects in a rSSFeed
```
emteknetnz: Capitilisation

en/02_Developer_Guides/11_Integration/How_Tos/custom_csvbulkloader.md
```diff
@@ -4,116 +4,122 @@ summary: Customise your data importing
 icon: upload
 ---
 
-# How to: A custom CSVBulkLoader instance
+# How to: a custom cSVBulkLoader instance
```
emteknetnz: Capitilisation

en/02_Developer_Guides/11_Integration/How_Tos/custom_csvbulkloader.md
```diff
@@ -4,116 +4,122 @@ summary: Customise your data importing
...
+            'Name' => '->importFirstAndLastName',
+            'Geburtsdatum' => 'Birthday',
+            'Gruppe' => 'Team.Title',
+        ];
+
+        $this->$duplicateChecks = [
+            'SpielerNummer' => 'PlayerNumber',
+        ]
+
+        $this->$relationCallbacks = [
```
emteknetnz: These two look wrong again

en/02_Developer_Guides/12_Search/02_FulltextSearch.md
```diff
@@ -16,23 +16,24 @@ a high level wrapper for running advanced search services such as Solr, Lucene o
 `MySQL` search.
 [/notice]
 
-## Adding Fulltext Support to MySQLDatabase
+## Adding fulltext support to mySQLDatabase
```
emteknetnz: Capitilisation

en/02_Developer_Guides/14_Files/05_File_Migration.md
```diff
@@ -273,23 +274,23 @@ This will take all existing files under a hash path and move them to their natur
...
 
-No, you do not need to run the file migration task when upgrading to Silverstripe CMS 4.4 or later. 
+No, you do not need to run the file migration task when upgrading to Silverstripe CMS 4.4 or later.
 
 Your existing files will still resolve. Old hash paths will be redirected to the latest version of the file. As newer versions of existing files are published, they will be moved to their natural path.
 
 You can choose to run the migration task at any point.
 
-### What if I’m using `legacy_filenames`?
+### What if i’m using `legacy_filenames`?
```
emteknetnz: Capilitisation - looks like it doesn't have a regular `'` quote

emteknetnz (changes requested):
My main things
- For file names for code fences, they should be bold above code fence, not in it the code fence
- Remove empty namespaces from php files
- There's a fair number capitilisation problems, this could get really annoying if failing builds because of it and cannot merge because the linter has got it wrong

GuySartorelli:
>  For file names for code fences, they should be bold above code fence, not in it the code fence

I'll leave discussion about this in https://github.com/silverstripe/developer-docs/pull/387#discussion_r1391731248

> Remove empty namespaces from php files

Do you mean there are some `namespace;` somewhere? I can't find any? Please specify.

> There's a fair number capitilisation problems, this could get really annoying if failing builds because of it and cannot merge because the linter has got it wrong

I'm assuming you pointed those out in the review. If you didn't, please do so or I won't know what to fix.

.markdownlint-cli2.mjs
ChloeSartorelli: Done

en/00_Getting_Started/02_Composer.md
ChloeSartorelli: If it's intended to be a link, sure. This isn't intended to be a link - it's just referencing the silverstripe.org website. The website doesn't have a name to call it, or I'd put that instead.

en/00_Getting_Started/02_Composer.md
ChloeSartorelli: Don't see why not - or why it would make a difference? But updated to avoid pingpong.

en/00_Getting_Started/02_Composer.md
ChloeSartorelli: I've removed this FAQ question as it seems like it's from back when people didn't use Composer much anyway.

en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
ChloeSartorelli: They aren't pointless. They're required to pass linting, and they're also reflective of how the `Page` and `PageController` classes are declared in the recipe.

en/02_Developer_Guides/00_Model/02_Relations.md
ChloeSartorelli: Removed it and ran linting, and it passed fine - probably because it's in a PHPDoc comment and not in direct code. In direct code, there's a rule that requires adding a `use` statement.

Not sure why this backslash was added. I'll leave it out.

en/02_Developer_Guides/00_Model/10_Versioning.md
ChloeSartorelli: > I'd much prefer to have the filename above the code fence in bold is best?

That violates [MD036 - emphasis used instead of heading](https://github.com/DavidAnson/markdownlint/blob/main/doc/md036.md), which is why I converted those to be comments inside the codeblock itself.

There wasn't actually a documented standard for those anyway, so we're not violating our documentation standards by changing this - and one of the edits in this PR _does_ explicitly add a new standard to say that these should be added as comments in the code block going forward.

I think having to remove a single line in an example that they copy/paste is not a burden that people will realistically have problems with - and the JSON thing is also fine since these are documentation examples and not real JSON in real project code. People will understand the point, I think.

en/02_Developer_Guides/03_Forms/03_Form_Templates.md
ChloeSartorelli: Updated

en/02_Developer_Guides/03_Forms/Field_types/03_HTMLEditorField.md
ChloeSartorelli: Fixed

en/02_Developer_Guides/04_Configuration/01_SiteConfig.md
ChloeSartorelli: Fixed

en/02_Developer_Guides/05_Extending/01_Extensions.md
ChloeSartorelli: Fixed

en/02_Developer_Guides/06_Testing/01_Functional_Testing.md
ChloeSartorelli: Fixed

en/02_Developer_Guides/06_Testing/01_Functional_Testing.md
ChloeSartorelli: Fixed along with similar headings in this file

en/02_Developer_Guides/09_Security/01_Access_Control.md
ChloeSartorelli: Fixed along with a few other examples I found (assuming you meant the `.PHP`)

en/02_Developer_Guides/11_Integration/00_CSV_Import.md
ChloeSartorelli: Fixed

en/02_Developer_Guides/11_Integration/00_CSV_Import.md
ChloeSartorelli: What do you mean, specifically? What is wrong, and what should it be changed to? And, is the thing you're referring to in scope for linting or was it pre-existing?

en/02_Developer_Guides/11_Integration/02_RSSFeed.md
ChloeSartorelli: Fixed

en/02_Developer_Guides/11_Integration/How_Tos/custom_csvbulkloader.md
ChloeSartorelli: Lowercase "a" is correct, but fixed `CSVBulkLoader`

en/02_Developer_Guides/12_Search/02_FulltextSearch.md
ChloeSartorelli: Fixed

en/02_Developer_Guides/14_Files/05_File_Migration.md
ChloeSartorelli: Fixed by changing to "What if I use"

GuySartorelli:
Squashed all commits that update the documentation, because I accidentally adding linting changes to one of the non-linting doc change commits.

en/02_Developer_Guides/00_Model/10_Versioning.md
emteknetnz: I just don't like that standard

Whenever I do code examples I've always used the bold heading above the code fence. I think it's better for several reasons:
- It leads to better copy paste (no redundant comment above the filename that needs to be deleted, yes it's trivial though not having it there is objectively better)
- We don't have the issue of .json files which don't have native comments
- It's consistent - we use the same style for every file type.  This reduces cognitive friction which is the entire point for standardisation.  Remember people are going to add these manually so it's going to be annoying to be told by the linter that you're formatting is wrong for the silverstripe template code example you're adding - so then you need to lookup how to do a comments in silverstripe template which you've forgotten how. Much easier if it's just "do bold every time"

en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
emteknetnz: Yeah, lots of people in bespoke say they just delete the redundant namespace since it indents everything.  We should update those files in the recipe

> They're required to pass linting

Then seems like we should make an exception in linting rather than enforcing an annoying standard.

emteknetnz:
>> Remove empty namespaces from php files

> Do you mean there are some namespace; somewhere? I can't find any? Please specify.

I mean the instances where it has `namespace {` and then indents everything. It's pointless. The Page and PageController in the recipes have this, though they shouldn't.  People often just delete these because they do nothing.  Probably requires some custom linting rule to require a namespace, unless it's empty/root

>> There's a fair number capitilisation problems, this could get really annoying if failing builds because of it and cannot merge because the linter has got it wrong

> I'm assuming you pointed those out in the review. If you didn't, please do so or I won't know what to fix.

I pointed out several in the peer review, though I've probably missed loads because of the sheer quantity of files changed.  Most of these were 'sentance casing' classnames in headings.  Some others so changing the casing of variables because they match a name in a list e.g. `$url` becoming `$URL`.  My concern here is that the linting is being overly aggressive and getting it wrong.  My concern here is that the linting will fail builds because it has gotten casing wrong. That's going to be very annoying and the negative will exceed the benefit gained from continuously checking the casing of docs.  I'm not sure how to resolve this.  May even need to simply switch this off for CI and just run that particularly bit of linting from a laptop every so often.

en/02_Developer_Guides/11_Integration/00_CSV_Import.md
ChloeSartorelli: @emteknetnz Please answer this so I can fix it if something needs fixing, and ignore it if it's out of scope.

en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
ChloeSartorelli: I don't think it's an annoying standard? It matches what our recipes have, and it is best practice to leave these in. If bespoke choose to remove them, that's up to them. I argue that it isn't redundant, it's explicit - and being explicit is good.

Basically, I think that if we're going to add an exception here, then we should also remove the namespace block from the recipe - and I _don't_ think we should remove the namespace block from the recipe.

en/02_Developer_Guides/00_Model/10_Versioning.md
ChloeSartorelli: The first point is trivial, so probably not worth arguing about.

The second point I've already spoken to and don't think it's a problem.

The third point is actually incorrect - it wasn't being consistently done. But even if it was, changing one consistent thing for another consistent thing invalidates this point. Unless your point regarding consistency is that different language types have different comment styles? But that seems like a moot point since we have comments in the examples anyway, and we're not going to _not_ have comments in them, so both the reader and the writer need to be familiar with the different comment syntax anyway.

Options:
1. Leave it the way it now is. I think this is sensible.
    - It's the same style Symfony uses if anyone cares about that - though that should only weigh in so far as "there is a precedent for this". The fact that the precedent is Symfony specifically shouldn't be used to weigh this decision.
    - it keeps the filename inside the code block which removes any chance someone will add extra content between the filename and the code block
    - it passes linting without having to either globally remove this rule (and therefore potentially reduce the quality of our markdown) or add an exclude comment for every filename.
2. Use emphasis above the code block and add an explicit rule exemption for every filename
    - Likely to be forgotten when initially adding new code examples, and thus adding a little bit of friction when doing so
    - There were several examples in the docs before now where italics instead of bold was used. It's an easy mistake to make and is likely to happen again IMO.
3. Use emphasis above the code block and disable the `MD036` rule globally
    - Results in allowing emphasis when headings should be used, which decreases readability and accessibility.
    - There were several examples in the docs before now where italics instead of bold was used. It's an easy mistake to make and is likely to happen again IMO.
    
@maxime-rainville Do you have any preference here?

GuySartorelli:
> There's a fair number capitilisation problems, this could get really annoying if failing builds because of it and cannot merge because the linter has got it wrong

> I pointed out several in the peer review, though I've probably missed loads because of the sheer quantity of files changed. Most of these were 'sentance casing' classnames in headings. Some others so changing the casing of variables because they match a name in a list e.g. $url becoming $URL. My concern here is that the linting is being overly aggressive and getting it wrong.

I really wish there was a way to reply to that type of comment directly on GitHub lol. Anyway:

I think it's fine for us to have a few cases as a result of this PR where casing isn't 100% correct. We can fix them as we find them, but I don't think there will be many left since you and I have both put in some effort to find them already.

In headings, it will be fairly obvious when looking at the docs that a classname was meant to have been surrounded in backticks, and the actual documentation below the heading will be correct regardless.

In the `$url` vs `$URL` example, it's _very_ clear what was intended, I don't think it will confuse anyone, and again easy to resolve when we see them.

>  My concern here is that the linting will fail builds because it has gotten casing wrong. That's going to be very annoying and the negative will exceed the benefit gained from continuously checking the casing of docs. I'm not sure how to resolve this. May even need to simply switch this off for CI and just run that particularly bit of linting from a laptop every so often.

Existing examples have already been linted and are passing linting now. So it would only fail builds for any _new_ documentation. From the examples you found, and the examples I had found while fixing the issues initially, it's basically only wrong when someone forgets backticks. So for new documentation, if it says "the casing is wrong", it's usually going to be a reminder to us that we should be adding backticks.

I really don't think this is as big a problem as you're thinking.

en/02_Developer_Guides/00_Model/10_Versioning.md
emteknetnz: >  Unless your point regarding consistency is that different language types have different comment styles? But that seems like a moot point since we have comments in the examples anyway

It's not a moot point. If I'm writing a quick PR to add a code example to a changelog then it's simply easier if there's a single way to write filenames not matter what the extensions is.  Means I don't need to have to search to work out how to write it for a template file or for a json file. Not searching is easier than having to search?

emteknetnz:
> So it would only fail builds for any new documentation

Yeah that's my point, for instance I write a heading `## Adding an extension to LinkFieldAdminController` won't it failing linting and tell me it supposed to be `## Adding an extension to linkFieldAdminController`?

en/02_Developer_Guides/11_Integration/00_CSV_Import.md
emteknetnz: It's a non static property on `BulkLoader` which is an ancestor - should be `$this->duplicateChecks = ...` ?

GuySartorelli:
> > So it would only fail builds for any new documentation
> Yeah that's my point, for instance I write a heading `## Adding an extension to LinkFieldAdminController` won't it failing linting and tell me it supposed to be `## Adding an extension to linkFieldAdminController`?

Yes. And as I said:

> So for new documentation, if it says "the casing is wrong", it's usually going to be a reminder to us that we should be adding backticks.

i.e. you would see the linting failure, and update your PR to have ``## Adding an extension to `LinkFieldAdminController` `` instead.

emteknetnz:
OK cool. So what will happen if we put a `$url` variable to an example code block for new documentation?

en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
maxime-rainville: Honestly, I've been ignoring those `namespace` statement for the last 5+ years and never bothered to check what their purpose was.

Are they there to get around the fact PSR-4 would want Page to be in a namespace, but Page needs to be in the root namespace?

en/02_Developer_Guides/00_Model/10_Versioning.md
maxime-rainville: This seems like a very minor point to me. 


I would be incline to pick "Door 1 - Leave it the way it now is" in the name of avoiding linting exemption.

maxime-rainville (commented):
The last two point seem very minor to me.

I generally fall in the "avoid disabling linting rules" camp... especially when those rules are coming from some central source.

If there's still disagreement about this, maybe we have a quick 15 min catch up to go through it.

en/02_Developer_Guides/00_Model/01_Data_Model_and_ORM.md
ChloeSartorelli: I don't know what they're there for in the recipe - but they're there. If we're removing them we should remove them from the recipe. If we're not removing them from the recipe we should keep them in the docs.

Currently they're in the recipe, so it makes sense to put them in the docs in this PR. Removing them should be a separate card, if that's what we decide we want to do.

emteknetnz: I'd say delete from recipe

maxime-rainville: I tried to track down when we added those `namespace` statement. That's what I found https://github.com/silverstripe/recipe-cms/pull/11#issuecomment-378854003

Doesn't look like there was much thought that went into this.

We also have this old RFC which doesn't look like it ever got implemented. https://github.com/silverstripe/silverstripe-framework/issues/5844

In the interest of moving this along, I suggest we leave the `namespace` statement in the doc. For now, if someone cares deeply about removing the `namespace` statement, raise a separate card.

maxime-rainville (commented):
I think I've opined on all the outstanding points that were controversial.

en/02_Developer_Guides/11_Integration/00_CSV_Import.md
ChloeSartorelli: Ahh yup good spot. Fixed.

en/02_Developer_Guides/11_Integration/00_CSV_Import.md
ChloeSartorelli: Fixed

en/02_Developer_Guides/11_Integration/How_Tos/custom_csvbulkloader.md
ChloeSartorelli: Fixed

emteknetnz (approved):

PR #392, opened 2023-11-02
Update a example code comment

n8-dev:
Comment in code example didn't make sense

### DIFF
--- a/en/02_Developer_Guides/08_Performance/01_Caching.md
+++ b/en/02_Developer_Guides/08_Performance/01_Caching.md
@@ -141,7 +141,7 @@ use SilverStripe\Core\Injector\Injector;
 
 $cache = Injector::inst()->get(CacheInterface::class . '.myCache');
 
-// remove the cache item
+// set a cache item with an expiry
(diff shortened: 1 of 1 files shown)


michalkleiner (approved):

michalkleiner:
@GuySartorelli how low a version we should target this at?

GuySartorelli:
IMO it's not suuuper important to target the correct branch for small changes like this - but ideally any change that is a valid change for CMS 4 docs should target `4.13` and be merged up from there.
If this only affect CMS 5 docs then `5.1` is correct.

n8-dev:
Let me know if I can help; or rather how to make this better in the future.

I just clicked the edit on github button whilst reading docs as well.

GuySartorelli (approved):
As mentioned, this is fine where it is. Thanks for the edit.

> Let me know if I can help; or rather how to make this better in the future.

If you want to always target the correct branch, there's some guidance about that in [the contributing docs](https://docs.silverstripe.org/en/5/contributing/documentation/#branches-and-commit-messages) - but again, for small changes like this it doesn't matter too much.

PR #398, opened 2023-11-07
DOC Select all and deselect all files

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-asset-admin/issues/546


en/04_Changelogs/5.2.0.md
```diff
@@ -96,6 +97,10 @@ This is a behavioural change from the change introduced in Silverstripe CMS 5.1
 
 This only applies to users created through the security section of the CMS. Users created programatically with an empty password will not have a random password automatically assigned to them. This is the current behaviour and is often used for non-production purposes such as unit-testing.
 
+### Buttons to select all files and deselect all files{#bulk-action-buttons}
```
ChloeSartorelli:
```suggestion
### Buttons to select all files and deselect all files {#bulk-action-buttons}
```

en/04_Changelogs/5.2.0.md
```diff
@@ -96,6 +97,10 @@ This is a behavioural change from the change introduced in Silverstripe CMS 5.1
...
+
+The files section of the CMS now has buttons to select all files and deselect all files, which is useful when you want to perform an action on all files in the current folder. These buttons show at the bottom of the screen in the bulk actions toolbar. The deselect all files button also shows the number of files currently selected. As these buttons are on the bulk actions toolbar they are only visible if at least one file has been selected.
```
ChloeSartorelli: Need to add a note that it only works on the current page.
Probably worth also adding a screenshot, since it's a bit difficult to imagine what you're describing, but with a screenshot it becomes immediately obvious.

GuySartorelli (changes requested):

en/04_Changelogs/5.2.0.md
emteknetnz: Updated

en/04_Changelogs/5.2.0.md
emteknetnz: Have added screenshot. Have updated text to say that it only works on the current screen, and have updated text to say that it also selects folders

GuySartorelli (approved):
Would have preferred "current page" but it doesn't really matter, I'm okay with this as-is.

PR #402, opened 2023-11-21
DOC LogicException when no react component

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11058

### DIFF
--- a/en/04_Changelogs/5.2.0.md
+++ b/en/04_Changelogs/5.2.0.md
@@ -12,6 +12,7 @@ title: 5.2.0 (unreleased)
   - [Create random passwords for new users](#create-random-passwords-for-new-users)
   - [Buttons to select all files and deselect all files](#bulk-action-buttons)
   - [More nuanced permissions for `/dev/*` routes](#dev-route-permissions)
+  - [New exception in React forms](#react-forms-exception)
   - [Other new features](#other-new-features)
 - [API changes](#api-changes)
 - [Bug fixes](#bug-fixes)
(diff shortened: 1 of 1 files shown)


en/04_Changelogs/5.2.0.md
```diff
@@ -115,6 +115,7 @@ You can also now optionally implement a `canView()` method on your `BuildTask` i
 ### Other new features
 
 - A new [`SiteTree.hide_pagetypes`](api:SilverStripe\CMS\Model\SiteTree->hide_pagetypes) configuration property has been added. Unlike [`SiteTree.hide_ancestor`](api:SilverStripe\CMS\Model\SiteTree->hide_ancestor) (which has [now been deprecated](#api-silverstripe-cms)), this is an array. This allows you to define all page types that should be hidden in a single configuration property in your yaml configuration.
+- A [`LogicException`](https://www.php.net/manual/en/class.logicexception.php) is now thrown by [`FormSchema::getSchema()`](api:SilverStripe\Forms\Schema::getSchema()) if a react component was not found for a field type.
```
ChloeSartorelli: This really should have its own section, and probably even have a warning block since this _does_ have the chance to break some peoples' forms which are currently failing silently. Something like:
```text
[warning]
If your project or a module you're using is currently trying to include a field which doesn't have a react component into a react-rendered form, it will have been silently failing. The form will have been rendering everything except for the field(s) which have no react component.

This will now fail by throwing an exception, which means your form won't render at all until you remove or replace the field(s).
[/warning]
```

GuySartorelli (changes requested):

en/04_Changelogs/5.2.0.md
emteknetnz: Updated

en/04_Changelogs/5.2.0.md
```diff
@@ -12,6 +12,7 @@ title: 5.2.0 (unreleased)
   - [Create random passwords for new users](#create-random-passwords-for-new-users)
   - [Buttons to select all files and deselect all files](#bulk-action-buttons)
   - [More nuanced permissions for `/dev/*` routes](#dev-route-permissions)
+  - [React forms now throw an exception if a field type doesn't have a react component](#react-forms-exception)
```
ChloeSartorelli:
```suggestion
  - [New exception in React forms](#react-forms-exception)
```
Update the heading below as well. You can change the wording but it should be short. The heading doesn't need to convey all the information :p

GuySartorelli (changes requested):

en/04_Changelogs/5.2.0.md
emteknetnz: Updated

GuySartorelli (approved):
LGTM

Issue #405, opened 2023-11-23
Add method signatures to definition of public API


GuySartorelli:
Method signatures, including parameter names, are part of our definition of public API - but they're missing from the list in https://docs.silverstripe.org/en/project_governance/public_api/

## PRs
- https://github.com/silverstripe/developer-docs/pull/406

sabina-talipova:
PR is merged. Issue was solved.

PR #406, opened 2023-11-23
DOC Include method signatures in public API definition

GuySartorelli:
## Issue
- https://github.com/silverstripe/developer-docs/issues/405

### DIFF
--- a/en/06_Project_Governance/08_Public_API.md
+++ b/en/06_Project_Governance/08_Public_API.md
@@ -19,6 +19,8 @@ These are explicitly included in our definition of public API (unless excluded b
 * **global** functions, constants, and variables
 * namespaces, classes, interfaces, enums, and traits
 * public and protected scope (including methods, properties and constants)
+* signatures of functions/methods that are considered public API
+  * including typehints, parameters, and parameter names
 * private static class property declarations (considered to be configuration variables)
 * configuration default values (in yml files and in private statics)
 * yml configuration file and fragment names (see [yml configuration syntax](/developer_guides/configuration/configuration/#syntax))


en/06_Project_Governance/08_Public_API.md
```diff
@@ -19,6 +19,8 @@ These are explicitly included in our definition of public API (unless excluded b
 * **global** functions, constants, and variables
 * namespaces, classes, interfaces, enums, and traits
 * public and protected scope (including methods, properties and constants)
+* signatures of functions/methods that are considered public API
+  * including typehints, parameters, and parameter names
```
sabina-talipova, line 23: Looks good, but I have small question.
I'm not sure, but I'll ask. Should we include the return type and possible exceptions thrown in this list? Not as part of the method signature, but as part of the method contract.

ChloeSartorelli: The return type _is_ part of the method signature.

The only reason I explicitly mention parameters is because changing parameter _names_ isn't widely known as a breaking change (it only became so with PHP 8.0 introducing named arguments) and because people routinely try to add optional parameters to pre-existing methods. The idea is to stop people doing that - and that problem doesn't exist with return types so I don't think we need to explicitly mention them here.

Exceptions is a bit of a weird one, because _most_ of the time changing exceptions would be a breaking change, but there might be cases where something currently throws an exception and we want to _stop_ throwing that exception, e.g. if the exception implies a feature hasn't been implemented, but we decide to implement it. We'd obviously need to remove that exception.

sabina-talipova (approved):
Looks good to me.

PR #423, opened 2023-12-14
Typo error fix simple contact form

erangaddd:
Fixed typo in line 32 and added use SilverStripe\Forms\TextareaField;


michalkleiner:
@erangaddd from which branch did you create the patch, please? Can you redo it from `5.1` or rebase onto `5.1`? Even after changing the target branch from 5.0 to 5.1 the PR still carries a lot of unrelated commits so I don't even see what the change was.

PR #458, opened 2024-02-19
DOC Update percentages for eagerloading benchmark

GuySartorelli:
## Description

While working on https://github.com/silverstripe/silverstripe-framework/issues/10929 I had to do some benchmarking, and I wanted to validate my findings against the ones that were done for 5.1.

The percentages don't seem to be correct - this PR updates them based on https://www.calculatorsoup.com/calculators/algebra/percent-change-calculator.php

## Issues

- https://github.com/silverstripe/silverstripe-framework/issues/10929

### DIFF
--- a/en/04_Changelogs/5.1.0.md
+++ b/en/04_Changelogs/5.1.0.md
@@ -146,10 +146,10 @@ foreach ($teams as $team) {
 
 In a test setup with looping through 100 DataObjects each with 100 related DataObjects for a total of 10,000 records per test run, the following performance improvements were observed for different types of relations (eager-loading vs not eager-loading):
 
-- HasOne - 3227% faster (0.0078s vs 0.2595s)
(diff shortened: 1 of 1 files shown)


emteknetnz (changes requested):
I'm pretty sure this is incorrect

Putting in 0.1819 and 0.1453 into the linked calculator shows a 20.1209% decrease, so I see where you got the 20% from. 

However the changelog says that it's performs 25% FASTER, not that it's resulted in a 20% decrease in time

e.g. You make a 500 km trip at 100kmh in 5 hours.  If you make the same trip in 4 hours that's a 20% decrease in time.  However you'd be going at 125kmh which is a 25% increase in speed, e.g. your vehicle is now 25% faster

We had some argument when the work was orginally done about which metric to use i.e. "time taken" or "faster". We ended up going with faster

I'll close the PR

PR #467, opened 2024-03-04
DOC Update model validation docs

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/329

- swapped the order of "Model Validation" and "Contraints" to match the page title, and also because it's a more logical order IMO
- Added an addFieldError() example which I think is the most common use case for using model validation
- Updated h1 to match page title


GuySartorelli:
linting failure

emteknetnz:
Updated

maxime-rainville (approved):
LGTM

PR #523, opened 2024-05-29
DOC Update `cascade_deletes` information for `has_many` relationships

edwilde:
## Description

Given that [this issue](https://github.com/silverstripe/silverstripe-framework/issues/9612) has been open for some time, we should update the documentation which is giving incorrect information to developers.

I have added an example of how best to manage `cascade_deletes` between `many_many` related objects.

To avoid adding confusion, I decided against adding the alternative solution using `onBeforeDelete()` and `removeAll()`. However this can also be added if you think it is needed, or better?

## Issues
- https://github.com/silverstripe/silverstripe-framework/issues/9612

## Pull request checklist

- [x] The target branch is correct
    - See [branches and commit messages](https://docs.silverstripe.org/en/contributing/documentation#branches-and-commit-messages)
- [x] All commits are relevant to the purpose of the PR (e.g. no TODO comments, unrelated rewording/restructuring, or arbitrary changes)
    - Small amounts of additional changes are usually okay, but if it makes it hard to concentrate on the relevant changes, ask for the unrelated changes to be reverted, and submitted as a separate PR.
- [x] The commit messages follow our [commit message guidelines](https://docs.silverstripe.org/en/contributing/code/#commit-messages)
- [x] The PR follows our [contribution guidelines](https://docs.silverstripe.org/en/contributing/documentation/)
- [x] The changes follow our [writing style guide](https://docs.silverstripe.org/en/contributing/documentation/#writing-style)
- [x] Code examples follow our [coding conventions](https://docs.silverstripe.org/en/contributing/coding_conventions/)
- [x] CI is green

### DIFF
--- a/en/02_Developer_Guides/00_Model/02_Relations.md
+++ b/en/02_Developer_Guides/00_Model/02_Relations.md
@@ -991,7 +991,7 @@ class ParentObject extends DataObject
 
 In this example, when the parent object is deleted, the child specified by the `has_one` relation will also be deleted. The relation types `has_many`, `belongs_to`, and `has_one` are supported, as are methods that return lists of objects but do not correspond to a physical database relation.
 
-### Cascading deletions with `has_many`
+### Cascading deletions for `many_many` relations
 
 When the parent object is deleted, the end child specified by the `cascade_deletes` will also be removed, not unlinked. To remove the link between the two objects, use `has_many` on the relation object directly to create an alternate view of the relationship which can be unlinked.
 
--- a/en/02_Developer_Guides/00_Model/02_Relations.md
+++ b/en/02_Developer_Guides/00_Model/02_Relations.md
@@ -989,11 +989,13 @@ class ParentObject extends DataObject
 }
 ```
 
-In this example, when the parent object is deleted, the child specified by the `has_one` relation will also be deleted. The relation types `has_many`, `belongs_to`, and `has_one` are supported, as are methods that return lists of objects but do not correspond to a physical database relation.
+In this example, when the parent object is deleted, the child specified by the `has_one` relation will also be deleted.
+
+All relation types (`has_many`, `many_many`, `belongs_many_many`, `belongs_to`, and `has_one`) are supported, as are methods that return lists of objects but do not correspond to a physical database relation. In all cases the child defined in `cascade_deletes` will be deleted, never unlinked.
 
 ### Cascading deletions for `many_many` relations
 
-When the parent object is deleted, the end child specified by the `cascade_deletes` will also be removed, not unlinked. To remove the link between the two objects, use `has_many` on the relation object directly to create an alternate view of the relationship which can be unlinked.
+When the parent object is deleted, the end child specified by the `cascade_deletes` will also be removed, not unlinked. To remove just the link between the two objects, use `has_many` on the relation object directly to create an alternate view of the relationship which can be unlinked.
 
 For example:
 
@@ -1026,6 +1028,8 @@ class Team extends DataObject
     // ...
 }
 ```
+
+This will avoid the situation where related children used by many data objects are removed from all parents because a single parent was deleted.
 
 ### Cascading deletions with versions
 
--- a/en/02_Developer_Guides/00_Model/02_Relations.md
+++ b/en/02_Developer_Guides/00_Model/02_Relations.md
@@ -1019,7 +1019,7 @@ class Team extends DataObject
     ];
 
     // Unlink the relationship when the parent object is deleted
-    private static array $cascade_duplicates = [
+    private static array $cascade_deletes = [
         'SupportersRelation',
     ];
 
--- a/en/02_Developer_Guides/00_Model/02_Relations.md
+++ b/en/02_Developer_Guides/00_Model/02_Relations.md
@@ -997,6 +997,8 @@ All relation types (`has_many`, `many_many`, `belongs_many_many`, `belongs_to`, 
 
 When the parent object is deleted, the end child specified by the `cascade_deletes` will also be removed, not unlinked. To remove just the link between the two objects, use `has_many` on the relation object directly to create an alternate view of the relationship which can be unlinked.
 
+Note that this requires using the [`many_many` through](#many-many-through) style relation. A regular `many_many` relation cannot be used this way.
+
 For example:
 
 ```php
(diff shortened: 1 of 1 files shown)


en/02_Developer_Guides/00_Model/02_Relations.md
```diff
@@ -989,9 +989,45 @@ class ParentObject extends DataObject
...
+    // Unlink the relationship when the parent object is deleted
+    private static array $cascade_duplicates = [
+        'SupportersRelation',
+    ];
+
+    // ...
+}
+```
+
+### Cascading deletions with versions
```
mfendeksilverstripe, line 1036: It may be also worth noting that if you have a a setup when parent model is versioned and the related model is not, the un-publish action will delete the related model because un-publish is a shorthand for delete from live stage. Since the related model is not versioned the deletion occurs and the related model is gone completely.

en/02_Developer_Guides/00_Model/02_Relations.md
```diff
@@ -989,9 +989,45 @@ class ParentObject extends DataObject
 }
 ```
 
-In this example, when the parent object is deleted, the child specified by the `has_one` relation will also
-be deleted. Note that all relation types (`has_many`, `many_many`, `belongs_many_many`, `belongs_to`, and `has_one`)
-are supported, as are methods that return lists of objects but do not correspond to a physical database relation.
+In this example, when the parent object is deleted, the child specified by the `has_one` relation will also be deleted. The relation types `has_many`, `belongs_to`, and `has_one` are supported, as are methods that return lists of objects but do not correspond to a physical database relation.
```
ChloeSartorelli: This implies you cannot use `cascade_deletes` with a `many_many` relationship, which is not the case. The behaviour is different, but it is still supported.

Please update these docs so that the difference in behaviour is highlighted (perhaps with a `[!WARNING]` callout block) rather than implying that it's not supported.

en/02_Developer_Guides/00_Model/02_Relations.md
```diff
@@ -989,9 +989,45 @@ class ParentObject extends DataObject
...
+
+### Cascading deletions with `has_many`
```
ChloeSartorelli: This says `has_many` but the example shows a `many_many`... that's a bit confusing. This should probably be
```suggestion
### Cascading deletions for `many_many` relations
```
or perhaps
```suggestion
### Cascading deletions for `many_many` through relations
```

And then explain what using `cascade_deletes` directly on the `many_many` relation does, and provide your example as an alternative and describe what the behaviour is for that and how it differs.

en/02_Developer_Guides/00_Model/02_Relations.md
```diff
@@ -989,9 +989,45 @@ class ParentObject extends DataObject
...
+        ],
+    ];
+
+    // Add a different perspective of the relationship data
+    private static array $has_many = [
+        'SupportersRelation' => TeamSupporter::class . '.Team',
+    ];
+
+    // Unlink the relationship when the parent object is deleted
+    private static array $cascade_duplicates = [
```
ChloeSartorelli:
```suggestion
    private static array $cascade_deletes = [
```

GuySartorelli (changes requested):

en/02_Developer_Guides/00_Model/02_Relations.md
edwilde: Gah, sorry...copy/paste without reading 🙃

en/02_Developer_Guides/00_Model/02_Relations.md
edwilde: There are already a bunch of callouts in this area, so adding another will make it tough to read.

I have updated the text to hopefully make the distinction between deleting and, as an alternative, unlinking, clearer.

en/02_Developer_Guides/00_Model/02_Relations.md
edwilde: I'd say that is probably out-of-scope for this PR, but there is now a whole sub-section for these 😄

en/02_Developer_Guides/00_Model/02_Relations.md
edwilde: 🤦

en/02_Developer_Guides/00_Model/02_Relations.md
```diff
@@ -989,9 +989,49 @@ class ParentObject extends DataObject
...
-In this example, when the parent object is deleted, the child specified by the `has_one` relation will also
-be deleted. Note that all relation types (`has_many`, `many_many`, `belongs_many_many`, `belongs_to`, and `has_one`)
-are supported, as are methods that return lists of objects but do not correspond to a physical database relation.
+In this example, when the parent object is deleted, the child specified by the `has_one` relation will also be deleted.
+
+All relation types (`has_many`, `many_many`, `belongs_many_many`, `belongs_to`, and `has_one`) are supported, as are methods that return lists of objects but do not correspond to a physical database relation. In all cases the child defined in `cascade_deletes` will be deleted, never unlinked.
+
+### Cascading deletions for `many_many` relations
+
+When the parent object is deleted, the end child specified by the `cascade_deletes` will also be removed, not unlinked. To remove just the link between the two objects, use `has_many` on the relation object directly to create an alternate view of the relationship which can be unlinked.
```
ChloeSartorelli, line 998:
```suggestion
When the parent object is deleted, the end child specified by the `cascade_deletes` will also be removed, not unlinked. To remove just the link between the two objects, use `has_many` on the relation object directly to create an alternate view of the relationship which can be unlinked.

Note that this requires using the [`many_many` through](#many-many-through) style relation. A regular `many_many` relation cannot be used this way.
```

GuySartorelli (changes requested):
Thanks for making those changes - just one small addition and then I'll be happy.

en/02_Developer_Guides/00_Model/02_Relations.md
edwilde: Good to know 👍

GuySartorelli (approved):
Thanks for doing this, it's a great clarification to have.

I'll merge once CI goes green (all our runners are being used right now so it might be a bit)

edwilde:
Thanks for the speedy turnaround @GuySartorelli 🚀

Issue #486, opened 2024-03-26
Update PHP coding standards to avoid using `self`


GuySartorelli:
We had a bug which it turned out was caused by an unexpected interaction between `self` and late static binding (see [this PR](https://github.com/silverstripe/silverstripe-linkfield/pull/262)).
This was [raised in php-src](https://github.com/php/doc-en/issues/3293) where the response was that it was working as intended.

Going forward we should avoid calling static methods from `self` to avoid this class of bug.

## Notes
- In refinement we decided it's easiest to just avoid using `self` altogether, rather than only avoiding it for calling methods

## Acceptance Criteria
- [x] The [PHP coding standards](https://docs.silverstripe.org/en/5/contributing/coding_conventions/php_coding_conventions/) are updated to explicitly disallow using `self` (e.g. `self::get()`)
- [x] The docs mention the rationale, i.e. we're avoiding a specific class of bug that is introduced when using late static binding downstream from `self`
- [x] `static` is fine, because it explicitly uses late static binding anyway
- [x] References to `self` in our codebase are updated to use the actual class that is intended to received the call.
- [x] Update code-writing script and re-run to use `self::class::myMethod()` in traits

## Code writing script
- https://github.com/emteknetnz/self-remover

## Kitchen sink CI
- https://github.com/emteknetnz/recipe-kitchen-sink/actions/runs/9514611296

## PRs
- https://github.com/silverstripe/developer-docs/pull/528
- https://github.com/bringyourownideas/silverstripe-composer-update-checker/pull/90
- https://github.com/silverstripe/cwp-agencyextensions/pull/123
- https://github.com/silverstripe/cwp/pull/337
- https://github.com/silverstripe/cwp-core/pull/131
- https://github.com/silverstripe/silverstripe-elemental/pull/1201
- https://github.com/silverstripe/silverstripe-admin/pull/1773
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1467
- https://github.com/silverstripe/silverstripe-assets/pull/614
- https://github.com/silverstripe/silverstripe-behat-extension/pull/274
- https://github.com/silverstripe/silverstripe-blog/pull/761
- https://github.com/silverstripe/silverstripe-cms/pull/2955
- https://github.com/silverstripe/silverstripe-config/pull/115
- https://github.com/silverstripe/silverstripe-contentreview/pull/237
- https://github.com/silverstripe/silverstripe-crontask/pull/89
[...]

PR #528, opened 2024-06-05
DOC Do not use self

emteknetnz:
Issue https://github.com/silverstripe/developer-docs/issues/486

### DIFF
--- a/en/10_Contributing/05_Coding_Conventions/01_PHP_Coding_Conventions.md
+++ b/en/10_Contributing/05_Coding_Conventions/01_PHP_Coding_Conventions.md
@@ -172,3 +172,6 @@ PHPDocs are not only useful when looking at the source code, but are also used i
 - Prefer the identical `===` operator over the equality `==` operator for comparisons.
 - If you directly reference a third-party dependency in code, then ensure the dependency is required in the module's `composer.json` file.
 - Avoid hardcoding values when there is a method available to dynamically get a value, for example use [`DataObjectSchema::tableName()`](api:SilverStripe\ORM\DataObjectSchema::tableName()) to get the table name for a `DataObject` model rather than hard coding it.
+- Do not use the `self` keyword, e.g. `self::myMethod()` instead use the short-name of the current class e.g. `MyClass::myMethod()` which is functionally equivalent. This avoids a specific class of bug that is introduced when using late static binding downstream from `self`.
+  - The one exception to this rule is traits as there is no way to know the class the trait is applied to. In that case, it's fine to use for getting the class name (e.g. `self::class`) but should only be used for calling static methods (i.e. `self::class::myMethod()`) if you can't refactor the code to avoid needing to use `self`.
+  - The use of `static` keyword is perfectly OK and is often appropriate.


en/10_Contributing/05_Coding_Conventions/01_PHP_Coding_Conventions.md
```diff
@@ -172,3 +172,7 @@ PHPDocs are not only useful when looking at the source code, but are also used i
 - Prefer the identical `===` operator over the equality `==` operator for comparisons.
 - If you directly reference a third-party dependency in code, then ensure the dependency is required in the module's `composer.json` file.
 - Avoid hardcoding values when there is a method available to dynamically get a value, for example use [`DataObjectSchema::tableName()`](api:SilverStripe\ORM\DataObjectSchema::tableName()) to get the table name for a `DataObject` model rather than hard coding it.
+- Do not use the `self` keyword, e.g. `self::myMethod()` instead use the short-name of the current class e.g. `MyClass::myMethod()` which is functionally equivalent. This avoids a specific class of bug that is introduced when using late static binding downstream from `self`.
+  - The one exception to this rule is traits as there is no way to know the class the trait is applied to.
+  - The use of `static` keyword is perfectly OK and is often appropriate.
+  - You should probably not use `MyClass::config()` and instead use `static::config()` as that will read any overriden config from subclasses. `self::config()` somewhat unintuitively behaves like `static::config()` and should not be used.
```
emteknetnz: I've validated that `self::config()` does behave like `static::config()` and not like `MyClass::config()` as would be intuitive.

**MyDataObject.php**
```php
<?php

use SilverStripe\Core\Config\Configurable;
use SilverStripe\ORM\DataObject;

class MyDataObject extends DataObject
{
    use Configurable;

    private static $table_name = 'MyDataObject';

    private static $something = 'one';

    private static function go()
    {
        var_dump([
            self::config()->get('something'),
            static::config()->get('something'),
            self::config()->something,
            static::config()->something,
            MyDataObject::config()->something,
        ]);
    }

    public function test()
    {
        $this->go();
    }

    public static function self_test()
    {
        self::go();
    }

    public static function static_test()
    {
        static::go();
    }
}
```

**MyDataObjectChild.php**
```php
<?php

use SilverStripe\Core\Config\Configurable;

class MyDataObjectChild extends MyDataObject
{
    private static $table_name = 'MyDataObjectChild';

    private static $something = 'two';
}
```

```php
MyDataObjectChild::create()->test();
/*
  0 => string 'two'
  1 => string 'two'
  2 => string 'two'
  3 => string 'two'
  4 => string 'one'
*/
MyDataObjectChild::self_test();
/*
  0 => string 'two'
  1 => string 'two'
  2 => string 'two'
  3 => string 'two'
  4 => string 'one'
*/
MyDataObjectChild::static_test();
/*
  0 => string 'two'
  1 => string 'two'
  2 => string 'two'
  3 => string 'two'
  4 => string 'one'
*/
```

ChloeSartorelli:
```suggestion
```
We don't need to treat `config()` in a special way. This is behaving the same way as `get()` which is what triggered this whole process in the first place.

For any static method, if you want to call the method on the class the code is defined in, use the class name - and if you want to call the method on the class the object is a direct instance of, use `static`. `config()` is no different here than any other static method.

en/10_Contributing/05_Coding_Conventions/01_PHP_Coding_Conventions.md
```diff
@@ -172,3 +172,7 @@ PHPDocs are not only useful when looking at the source code, but are also used i
 - Prefer the identical `===` operator over the equality `==` operator for comparisons.
 - If you directly reference a third-party dependency in code, then ensure the dependency is required in the module's `composer.json` file.
 - Avoid hardcoding values when there is a method available to dynamically get a value, for example use [`DataObjectSchema::tableName()`](api:SilverStripe\ORM\DataObjectSchema::tableName()) to get the table name for a `DataObject` model rather than hard coding it.
+- Do not use the `self` keyword, e.g. `self::myMethod()` instead use the short-name of the current class e.g. `MyClass::myMethod()` which is functionally equivalent. This avoids a specific class of bug that is introduced when using late static binding downstream from `self`.
+  - The one exception to this rule is traits as there is no way to know the class the trait is applied to.
```
ChloeSartorelli:
```suggestion
  - The one exception to this rule is traits as there is no way to know the class the trait is applied to. In that case, it's fine to use for getting the class name (e.g. `self::class`) but should only be used for calling static methods if you can't refactor the code to avoid needing to use `self`.
```
In other words, even with traits make an effort to avoid using `self` so we can avoid the late-static-binding bug.

GuySartorelli (changes requested):

en/10_Contributing/05_Coding_Conventions/01_PHP_Coding_Conventions.md
ChloeSartorelli: I've just checked, and though it looks a little strange I've found that `self::class::myMethod()` works the way you'd expect it to - i.e. it calls the method on the class the code is defined in (in this case the class that directly uses the trait), not the class the object is an instance of.

Probably this should be the recommended way to call static methods and get static properties using `self` in traits (if its use can't be avoided altogether), since this eliminates the bug.

If you agree with that, it'll probably be worth updating the codewriter to respect this and rerun it against traits.

emteknetnz: Have updated and added mention of self::class::myMethod().  I'll update code-writer and run again post merging this other PRs - I've added an AC for this

GuySartorelli (approved):
LGTM

PR #538, opened 2024-07-01
DOC Modules no longer need to have a _config.php file or _config dire…

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11254


GuySartorelli (approved):
LGTM

PR #547, opened 2024-07-10
DOC Document upgrading intervention/image

GuySartorelli:
Needs https://github.com/silverstripe/documentation-lint/pull/6 for linting to pass

## Issue
- https://github.com/silverstripe/silverstripe-assets/issues/619

### DIFF
--- a/en/08_Changelogs/6.0.0.md
+++ b/en/08_Changelogs/6.0.0.md
@@ -10,6 +10,8 @@ title: 6.0.0 (unreleased)
   - [Run `CanonicalURLMiddleware` in all environments by default](#url-middleware)
   - [Changes to scaffolded form fields](#scaffolded-fields)
   - [Other new features](#other-new-features)
+- [Dependency changes](#dependency-changes)
+  - [`intervention/image` has been upgraded from v2 to v3](#intervention-image)
 - [Bug fixes](#bug-fixes)
 - [API changes](#api-changes)
   - [Most extension hook methods are now protected](#hooks-protected)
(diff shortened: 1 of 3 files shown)


en/08_Changelogs/6.0.0.md
```diff
@@ -58,6 +60,71 @@ All other models used `SearchableDropdownField` for `has_one` relations and `Gri
...
+
+We've upgraded from `intervention/image` v2 to v3. One of the main improvements included in this upgrade is full support for animated GIFs.
+
+If you are directly interacting with APIs from `intervention/image` in your project or module you should check out [their upgrade guide](https://image.intervention.io/v3/introduction/upgrade).
+
+#### Animated vs still images {#intervention-image-animations}
+
+Manipulating animated images takes longer, and results in a larger filesize.
+
+Because of this, the [`ThumbnailGenerator`](api:SilverStripe\AssetAdmin\Model\ThumbnailGenerator) will provide still images as thumbnails for animated gifs by default. You can change that for a given instance of `ThumbnailGenerator` by passing `true` to the [`setAllowsAnimation()`](api:SilverStripe\AssetAdmin\Model\ThumbnailGenerator::setAllowsAnimation()) method. For example, to allow animated thumbnails for `UploadField`:
```
emteknetnz, line 75: This also applies to all other PRs to changelogs - we're essentially duplicating docs on doc.silverstripe.org.  We should just link to those docs instead of duplicating them in the changelog.

emteknetnz (changes requested):

en/08_Changelogs/6.0.0.md
ChloeSartorelli: It's not at all uncommon for us to do this - we often give the most pertinent information directly in the changelog, even if it's also documented elsewhere.

> This also applies to all other PRs to changelogs

I assume you want some amount of this removed, can you please indicate how much so I know how much I want to push back on it? :p

emteknetnz: > It's not at all uncommon for us to do this - we often give the most pertinent information directly in the changelog, even if it's also documented elsewhere.

Yeah we do ... though just looking at this PR I actually have no idea why we've been doing this. We're quick to call out duplicate documentation in all other scenarios and say "You should just link to the original source" ... I'm not sure why the changelog should be treated any differently.

> I assume you want some amount of this removed, can you please indicate how much so I know how much I want to push back on it? :p

Probably everything not in the first section `intervention/image has been upgraded from v2 to v3 {#intervention-image}` - I'd keep that and say that imagick is now the default if it's available and link to the docs

Most of the rest is duplication of "how to configure it", which is basically straight duplication

I'd also remove the "New API" bit, it just seems unnecessary.  Very few people are going to be interested in calling the PHP methods directly, the configuration is much more relevant.  If they really are interested in calling PHP methods directly then there's typehinting available anyway.

emteknetnz (changes requested):

GuySartorelli:
>  I'm not sure why the changelog should be treated any differently.

The changelog is the one thing developers will look at when evaluating what changes they need to make, and it serves double duty as an upgrade guide. Giving all of the most important information in the changelog does two things:

1. Reduces the number of extra pages developers need to look at to get the most relevant information
2. Increases the chance that developers will actually see it - because lets face it a lot of devs will just not click on the links, but are fairly likely to copy/paste the code examples when we put them right in front of them.

I like the precedent that the important information is included in the changelog even if that same information is provided in other sections of the docs. I see the changelog as "here's all the most important information, including the most relevant info you need for upgrading", whereas the rest of the docs is for future reference or searching for specific information, outside of the context of releases.

GuySartorelli:
I'm not keen to change that precedent now, especially when we don't have a team lead.

GuySartorelli:
We've also got a while 'til this gets released, and if it's anything like CMS 5 we'll spend a fair amount of time going back over the changelog to decide what stays in there, so we can revisit this precedent then, if we haven't already revisited it for a minor release in the meantime (again, when we have a team lead preferably)

emteknetnz (approved):

PR #563, opened 2024-08-20
DOC Deprecate DataExtension and others

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11050

PR #566, opened 2024-08-20
DOC Replace Extension subclasses

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11050

Note that re the changelog - this was also mentioned in the https://github.com/silverstripe/developer-docs/pull/563, and we don't normally repeat that in the 6.0.0 changelog, however Deprecation::withNoReplacement() was used in that CMS 5 PRs which will prevent the deprecation notices from showing in CMS 5, so I think we should at least mention this in the CMS 6 changelog

### DIFF
--- a/en/02_Developer_Guides/00_Model/10_Versioning.md
+++ b/en/02_Developer_Guides/00_Model/10_Versioning.md
@@ -8,7 +8,7 @@ summary: Add versioning to your database content through the Versioned extension
 Database content in Silverstripe CMS can be "staged" before its publication, as well as track all changes through the
 lifetime of a database record.
 
-Versioning in Silverstripe CMS is handled through the [`Versioned`](api:SilverStripe\Versioned\Versioned) extension class. As an [`DataExtension`](api:SilverStripe/ORM/DataExtension) it is possible to be applied to any [`DataObject`](api:SilverStripe\ORM\DataObject) subclass. The extension class will automatically update read and write operations performed via the ORM because it implements the [`augmentSQL()`](api:SilverStripe/ORM/DataExtension::augmentSql()) extension hook method.
+Versioning in Silverstripe CMS is handled through the [`Versioned`](api:SilverStripe\Versioned\Versioned) extension class. As an [`Extension`](api:SilverStripe\Core\Extension) it is possible to be applied to any [`DataObject`](api:SilverStripe\ORM\DataObject) subclass. The extension class will automatically update read and write operations performed via the ORM because it implements the `augmentSQL` extension hooks defined in [`DataQuery::getFinalisedQuery()`](api:SilverStripe\ORM\DataQuery::getFinalisedQuery()).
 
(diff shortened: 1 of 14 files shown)


en/02_Developer_Guides/00_Model/10_Versioning.md
```diff
@@ -8,7 +8,7 @@ summary: Add versioning to your database content through the Versioned extension
 Database content in Silverstripe CMS can be "staged" before its publication, as well as track all changes through the
 lifetime of a database record.
 
-Versioning in Silverstripe CMS is handled through the [`Versioned`](api:SilverStripe\Versioned\Versioned) extension class. As an [`DataExtension`](api:SilverStripe/ORM/DataExtension) it is possible to be applied to any [`DataObject`](api:SilverStripe\ORM\DataObject) subclass. The extension class will automatically update read and write operations performed via the ORM because it implements the [`augmentSQL()`](api:SilverStripe/ORM/DataExtension::augmentSql()) extension hook method.
```
ChloeSartorelli, line 11: Why are you removing context about `augmentSql()`? Does the `Versioned` extension not actually do that?

en/08_Changelogs/6.0.0.md
```diff
@@ -338,6 +338,7 @@ Injector::inst()->load([
...
 - [`FieldList`](api:SilverStripe\Forms\FieldList) is now strongly typed. Methods that previously allowed any iterable variables to be passed, namely [`FieldList::addFieldsToTab()`](api:SilverStripe\Forms\FieldList::addFieldsToTab()) and [`FieldList::removeFieldsFromTab()`](api:SilverStripe\Forms\FieldList::removeFieldsFromTab()), now require an array to be passed instead.
 - [`BaseElement::getDescription()`](api:DNADesign\Elemental\Models\BaseElement::getDescription()) has been removed. If you had implemented this method in your custom elemental blocks, either set the [`description`](api:DNADesign\Elemental\Models\BaseElement->description) configuration property or override the [`getTypeNice()`](api:DNADesign\Elemental\Models\BaseElement::getTypeNice()) method.
+- [`DataExtension`](api:SilverStripe\ORM\DataExtension), [`SiteTreeExtension`](api:SilverStripe\CMS\Model\SiteTreeExtension), and [`LeftAndMainExtension`](api:SilverStripe\Admin\LeftAndMainExtension) have been removed. If you subclass any of these classes, you must now subclass [`Extension`](api:SilverStripe\Core\Extension) instead.
```
ChloeSartorelli, line 341: I would almost suggest this gets its own heading, since these are exceptionally high traffic classes. But it's probably fine here like this.

ChloeSartorelli: Also, I just realised that the API links won't work for these - don't bother fixing that right now, there's a bunch of similar links that will stop working in this changelog. We can resolve that when we do the first alpha by giving it a once-over.

en/02_Developer_Guides/00_Model/10_Versioning.md
emteknetnz: It's technically wrong, updated it to say it implements the augmentSql hook that's defined in DataQuery

GuySartorelli (approved):
LGTM

PR #574, opened 2024-09-03
DOC Updated symfony dependencies

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/296

### DIFF
--- a/en/08_Changelogs/6.0.0.md
+++ b/en/08_Changelogs/6.0.0.md
@@ -15,6 +15,7 @@ title: 6.0.0 (unreleased)
   - [Other new features](#other-new-features)
 - [Dependency changes](#dependency-changes)
   - [`intervention/image` has been upgraded from v2 to v3](#intervention-image)
+  - [symfony dependencies have been upgraded from v6 to v7](#symfony)
 - [Bug fixes](#bug-fixes)
 - [API changes](#api-changes)
   - [Most extension hook methods are now protected](#hooks-protected)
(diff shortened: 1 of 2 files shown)


en/08_Changelogs/6.0.0.md
```diff
@@ -15,6 +15,7 @@ title: 6.0.0 (unreleased)
   - [Other new features](#other-new-features)
 - [Dependency changes](#dependency-changes)
   - [`intervention/image` has been upgraded from v2 to v3](#intervention-image)
+  - [`symfony dependencies have been upgraded from v6 to v7](#symfony)
```
ChloeSartorelli:
```suggestion
  - [symfony dependencies have been upgraded from v6 to v7](#symfony)
```

en/08_Changelogs/6.0.0.md
```diff
@@ -215,6 +216,10 @@ We've upgraded from `intervention/image` v2 to v3. One of the main improvements
 
 If you are directly interacting with APIs from `intervention/image` in your project or module you should check out [their upgrade guide](https://image.intervention.io/v3/introduction/upgrade).
 
+### Symfony dependencies have been upgraded from v6 to v7 {#symfony}
+
+We've upgraded the various `symfony` dependencies from v6 to v7. If you are directly interacting with `symfony` components in your project or module you should check out the [symfony upgrade guide](https://symfony.com/doc/current/setup/upgrade_major.html).
```
ChloeSartorelli: `symfony` is a bit weird... probably should either be "Symfony" or `symfony/*`

ChloeSartorelli: We also shouldn't call out that upgrade guide, as it's geared towards people who are using symfony framework for everything in their project (the profiler for example probably won't work in a silverstripe app). Or if we do call it out, we should point out that some things in that guide won't work for a silverstripe app.

We didn't call out an upgrade guide for 5.0, probably don't need to here either.

GuySartorelli (changes requested):

en/08_Changelogs/6.0.0.md
emteknetnz: Updated

en/08_Changelogs/6.0.0.md
emteknetnz: Updated

GuySartorelli (approved):
LGTM

PR #577, opened 2024-09-04
DOC Changing ClassName from Enum to Varchar

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11358

### DIFF
--- a/en/02_Developer_Guides/08_Performance/06_ORM.md
+++ b/en/02_Developer_Guides/08_Performance/06_ORM.md
@@ -49,3 +49,26 @@ SilverStripe\ORM\Connect\DBSchemaManager:
 ```
 
 You can always manually trigger a check and repair (e.g. in a [`BuildTask`](api:SilverStripe/Dev/BuildTask)) by calling [`DB::check_and_repair_table()`](api:SilverStripe\ORM\DB::check_and_repair_table()). This ignores the above configuration.
+
+## Changing `ClassName` column from enum to varchar {#classname-varchar}
+
+On websites with very large database tables it can take a long time to run `dev/build`, which can be a problem when deploying changes to production. This is because the `ClassName` column is an `enum` type which requires an a `ALTER TABLE` query to be run affecting every row whenever there is a new valid value for the column.
+
+For a very rough benchmark, running an `ALTER TABLE` query on a database table of 10 million records took 28.52 seconds on a mid-range 2023 laptop, though this time will vary depending on the database and hardware being used.
+
+You may wish to change the `ClassName` column to a `varchar` type which remove the need to run `ALTER TABLE` whenever there is a new valid value. Enabling this will result in a trade-off where the size of the database will increase by approximately 7 MB per 100,000 rows.
+
+> [!WARNING]
+> There will also be a very slow initial `dev/build` as all of the `ClassName` columns are switched to `varchar`.
+
+To enable this, add the following configuration:
+
+```yml
+SilverStripe\ORM\DataObject:
+  fixed_fields:
+    ClassName: DBClassNameVarchar
+
+SilverStripe\ORM\FieldType\DBPolymorphicForeignKey:
+  composite_db:
+    Class: "DBClassNameVarchar('SilverStripe\\ORM\\DataObject', ['index' => false])"
+```
--- a/en/08_Changelogs/5.4.0.md
+++ b/en/08_Changelogs/5.4.0.md
@@ -7,11 +7,30 @@ title: 5.4.0 (unreleased)
 ## Overview
 
 - [Features and enhancements](#features-and-enhancements)
+  - [Option to change `ClassName` column from enum to varchar](#classname-varchar)
   - [Other new features](#other-new-features)
 - [API changes](#api-changes)
 - [Bug fixes](#bug-fixes)
 
 ## Features and enhancements
+
+### Option to change `ClassName` column from enum to varchar {#classname-varchar}
+
+On websites with very large database tables it can take a long time to run `dev/build`, which can be a problem when deploying changes to production. This is because the `ClassName` column is an `enum` type which requires an a `ALTER TABLE` query to be run affecting every row whenever there is a new valid value for the column. For a very rough benchmark, running an `ALTER TABLE` query on a database table of 10 million records took 28.52 seconds on a mid-range 2023 laptop, though this time will vary depending on the database and hardware being used.
+
+This release introduces a new configuration option to change the `ClassName` column to a `varchar` type which removes the need to run `ALTER TABLE` whenever there is a new valid value.
+
+Enabling this will result in a trade-off where the size of the database will increase by approximately 7 MB per 100,000 rows. There will also be a very slow initial `dev/build` as all of the `ClassName` columns are switched to `varchar`. To enable this, add the following configuration:
+
+```yml
+SilverStripe\ORM\DataObject:
+  fixed_fields:
+    ClassName: DBClassNameVarchar
+
+SilverStripe\ORM\FieldType\DBPolymorphicForeignKey:
+  composite_db:
+    Class: "DBClassNameVarchar('SilverStripe\\ORM\\DataObject', ['index' => false])"
+```
 
 ### Other new features
 


en/08_Changelogs/5.4.0.md
```diff
@@ -0,0 +1,43 @@
...
+  - [Option to change `ClassName` column from enum to varchar](#classname-varchar)
+  - [Other new features](#other-new-features)
+- [API changes](#api-changes)
+- [Bug fixes](#bug-fixes)
+
+## Features and enhancements
+
+### Option to change `ClassName` column from enum to varchar {#classname-varchar}
+
+On websites with very large database tables it can take a long time to run `dev/build`, which can be a problem when deploying changes to production. This is because the `ClassName` column is an `enum` type which requires an a `ALTER TABLE` query to be run affecting every row whenever there is a new valid value for the column. This release introduces a new configuration option to change the `ClassName` column to a `varchar` type which remove the need to run `ALTER TABLE` whenever there is a new valid value. Enabling this will result in a trade-off where the size of the database will increase by approximately 6mb per 100,000 rows. There will also be a very slow initial `dev/build` as all of the `ClassName` columns are switched to `varchar`. To enable this, add the following configuration:
```
ChloeSartorelli:
```suggestion
On websites with very large database tables it can take a long time to run `dev/build`, which can be a problem when deploying changes to production. This is because the `ClassName` column is an `enum` type which requires an a `ALTER TABLE` query to be run affecting every row whenever there is a new valid value for the column. This release introduces a new configuration option to change the `ClassName` column to a `varchar` type which removes the need to run `ALTER TABLE` whenever there is a new valid value.

Enabling this will result in a trade-off where the size of the database will increase by approximately 6mb per 100,000 rows. There will also be a very slow initial `dev/build` as all of the `ClassName` columns are switched to `varchar`. To enable this, add the following configuration:
```

en/08_Changelogs/5.4.0.md
ChloeSartorelli, line 1: You've given an approximation of the DB size increase, but not of how much the performance will improve. I think we need to give at least a rough estimate for each side of the trade-off.

en/02_Developer_Guides/08_Performance/06_ORM.md
```diff
@@ -49,3 +49,15 @@ SilverStripe\ORM\Connect\DBSchemaManager:
 ```
 
 You can always manually trigger a check and repair (e.g. in a [`BuildTask`](api:SilverStripe/Dev/BuildTask)) by calling [`DB::check_and_repair_table()`](api:SilverStripe\ORM\DB::check_and_repair_table()). This ignores the above configuration.
+
+## Changing `ClassName` column from enum to varchar {#classname-varchar}
+
+On websites with very large database tables it can take a long time to run `dev/build`, which can be a problem when deploying changes to production. This is because the `ClassName` column is an `Enum` type which requires an a `ALTER TABLE` query to be run affecting every row whenever there is a new valid value for the column.
```
ChloeSartorelli: You used `enum` lowercase in the changelog but `Enum` here. Please pick one. Same with `varchar` vs `Varchar` in the next paragraph.

en/02_Developer_Guides/08_Performance/06_ORM.md
```diff
@@ -49,3 +49,15 @@ SilverStripe\ORM\Connect\DBSchemaManager:
...
+
+You may wish to change the `ClassName` column to a `Varchar` type which remove the need to run `ALTER TABLE` whenever there is a new valid value. Enabling this will result in a trade-off where the size of the database will increase by approximately 6mb per 100,000 rows. There will also be a very slow initial `dev/build` as all of the `ClassName` columns are switched to `Varchar`. To enable this, add the following configuration:
```
ChloeSartorelli: Just like with the changelog - please add a rough estimate of the expected performance increase so people can make an informed decision about this.

GuySartorelli (changes requested):

en/02_Developer_Guides/08_Performance/06_ORM.md
emteknetnz: Done, have gone with lowercase

en/02_Developer_Guides/08_Performance/06_ORM.md
emteknetnz: I don't think I can provide any meaningful measurements here.

The performance difference between enum vs varchar is technically unlimited if you look at that one column in isolation, because with varchar you simply don't run ALTER TABLE queries.

In terms of the % of dev/build time spent doing ALTER TABLE queries, is be totally different of different websites.  This means I there's no way I can say "dev/build time is x% faster using varchars".

I created a [module](https://github.com/emteknetnz/silverstripe-dev-build-benchmark) that shows the % of dev/build query time spent on ALTER TABLE queries, and on a couple of different sites (a barebones local, and a deployed test setup with some data) I got completely different results. Real world sites will have a completely different % again.

en/08_Changelogs/5.4.0.md
emteknetnz: Left a comment about this [here](https://github.com/silverstripe/developer-docs/pull/577#discussion_r1746232216)

en/08_Changelogs/5.4.0.md
emteknetnz: Done

en/02_Developer_Guides/08_Performance/06_ORM.md
ChloeSartorelli: This doesn't give me much confidence that it does improve performance, if you're saying it's not reliably measurable.

emteknetnz: Well it's removing queries during dev/build, however the % that those queries take up has massive variance between websites.  On some websites it will be negliable, in others it could well be 10s of minutes.

I'm not sure how I'm supposed to provide a measurement here

This solution is very much only for a niche of websites with huge database tables that struggle with deployments. As I've mentioned above I've provided a module that can help qualify if this solution may be useful as it will show roughly the potential time saving from removing the ALTER TABLE queries.

ChloeSartorelli: You were able to provide an estimate of the amount of extra db space required, which will also vary from project to project. You did this by equating it to an amount of space per x records.

Surely a similar thing can be done with performance? For each X records, roughly Y seconds will be saved? It doesn't have to be exact, just a ballpark based on observations in your local.

emteknetnz: Extra db space was very easy measure as what I provided is simply the extra bytes for an averaged sized ClassName which I defined myself. There is no variance there.

However I cannot provide a estimate for the total bytes saved for a project, because there will be massive variance between projects.

emteknetnz: Will say an ALTER TABLE query with 10 million records took 28.52 seconds on my local

GuySartorelli:
LGTM  but needs to be rebased

en/02_Developer_Guides/08_Performance/06_ORM.md
ChloeSartorelli, line 1: We should probably add a warning that the first dev/build after they make this change may take a while, because it's updating the structure of _every_ table.

GuySartorelli (changes requested):

en/02_Developer_Guides/08_Performance/06_ORM.md
emteknetnz: Added a warning

en/02_Developer_Guides/08_Performance/06_ORM.md
```diff
@@ -49,3 +49,26 @@ SilverStripe\ORM\Connect\DBSchemaManager:
...
+To enable this, add the following configuration:
+
+```yml
+SilverStripe\ORM\DataObject:
+  fixed_fields:
+    ClassName: DBClassNameVarchar
+
+SilverStripe\ORM\FieldType\DBPolymorphicForeignKey:
+  composite_db:
+    Class: DBClassNameVarchar('SilverStripe\ORM\DataObject', ['index' => false])
```
ChloeSartorelli:
```suggestion
    Class: "DBClassNameVarchar('SilverStripe\ORM\DataObject', ['index' => false])"
```
Looks like something special is happening if we don't wrap it as an explicit string

en/08_Changelogs/5.4.0.md
```diff
@@ -7,12 +7,31 @@ title: 5.4.0 (unreleased)
...
+Enabling this will result in a trade-off where the size of the database will increase by approximately 7 MB per 100,000 rows. There will also be a very slow initial `dev/build` as all of the `ClassName` columns are switched to `varchar`. To enable this, add the following configuration:
+
+```yml
+SilverStripe\ORM\DataObject:
+  fixed_fields:
+    ClassName: DBClassNameVarchar
+
+SilverStripe\ORM\FieldType\DBPolymorphicForeignKey:
+  composite_db:
+    Class: DBClassNameVarchar('SilverStripe\ORM\DataObject', ['index' => false])
```
ChloeSartorelli:
```suggestion
    Class: "DBClassNameVarchar('SilverStripe\ORM\DataObject', ['index' => false])"
```

GuySartorelli (changes requested):

en/02_Developer_Guides/08_Performance/06_ORM.md
emteknetnz: Done. Needs double slashes when quoted.

en/08_Changelogs/5.4.0.md
emteknetnz: Done

GuySartorelli (approved):
LGTM

PR #583, opened 2024-09-11
DOC Update PHPUnit code sample

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/305

### DIFF
--- a/en/02_Developer_Guides/04_Configuration/00_Configuration.md
+++ b/en/02_Developer_Guides/04_Configuration/00_Configuration.md
@@ -400,6 +400,7 @@ Note that the configuration change is active only within the callback function.
 namespace App\Test\Service;
 
 use App\Service\MyService;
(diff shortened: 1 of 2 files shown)


GuySartorelli (changes requested):
Please add an example of how to use `enableErrorHandler()` to the [unit and integration testing](https://docs.silverstripe.org/en/5/developer_guides/testing/unit_testing/) page (or elsewhere if you find somewhere more appropriate).

emteknetnz:
Done - during this I also noticed that I didn't have any unit tests for SapphireTest::enableErrorHandler, so I've added those

Also I made some changes to the code as well to simply fallback to the standard php error handler if the errno wasn't one of the 6 that we're catching. This means that deprecation notices will still display properly instead of being thrown as Exceptions.

GuySartorelli (approved):
LGTM will merge when all other PRs are approved

PR #596, opened 2024-10-11
DOC Validate DBFields

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11403

### DIFF
--- a/en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
+++ b/en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
@@ -31,33 +31,41 @@ class Player extends DataObject
 }
 ```
 
+Most `DBField` subclasses will be validated using a [`FieldValidator`](api:SilverStripe\Core\Validation\FieldValidation\FieldValidator) subclass which is call as part of the `DataObject::validate()` method. This means that when a value is set on a `DBField` subclass, it will be validated against the constraints of that field. If the value is invalid then a [`ValidationException`](api:SilverStripe\Core\Validation\ValidationException) will be thrown.
+
 ## Available types
 
-- `'BigInt'`: An 8-byte signed integer field (see: [DBBigInt](api:SilverStripe\ORM\FieldType\DBBigInt)).
-- `'Boolean'`: A boolean field (see: [DBBoolean](api:SilverStripe\ORM\FieldType\DBBoolean)).
-- `'Currency'`: A number with 2 decimal points of precision, designed to store currency values. Only supports single currencies (see: [DBCurrency](api:SilverStripe\ORM\FieldType\DBCurrency)).
-- `'Date'`: A date field (see: [DBDate](api:SilverStripe\ORM\FieldType\DBDate)).
-- `'Datetime'`: A date/time field (see: [DBDatetime](api:SilverStripe\ORM\FieldType\DBDatetime)).
-- `'DBClassName'`: A special enumeration for storing class names (see: [DBClassName](api:SilverStripe\ORM\FieldType\DBClassName)).
-- `'Decimal'`: A decimal number (see: [DBDecimal](api:SilverStripe\ORM\FieldType\DBDecimal)).
-- `'Double'`: A floating point number with double precision (see: [DBDouble](api:SilverStripe\ORM\FieldType\DBDouble)).
-- `'Enum'`: An enumeration of a set of strings that can store a single value (see: [DBEnum](api:SilverStripe\ORM\FieldType\DBEnum)).
-- `'Float'`: A floating point number (see: [DBFloat](api:SilverStripe\ORM\FieldType\DBFloat)).
-- `'Foreignkey'`: A special `Int` field used for foreign keys in `has_one` relationships (see: [DBForeignKey](api:SilverStripe\ORM\FieldType\DBForeignKey)).
-- `'HTMLFragment'`: A variable-length string of up to 2MB, designed to store HTML. Doesn't process [shortcodes](/developer_guides/extending/shortcodes/). (see: [DBHTMLText](api:SilverStripe\ORM\FieldType\DBHTMLText)).
-- `'HTMLText'`: A variable-length string of up to 2MB, designed to store HTML. Processes [shortcodes](/developer_guides/extending/shortcodes/). (see: [DBHTMLText](api:SilverStripe\ORM\FieldType\DBHTMLText)).
-- `'HTMLVarchar'`: A variable-length string of up to 255 characters, designed to store HTML. Can process [shortcodes](/developer_guides/extending/shortcodes/) with additional configuration. (see: [DBHTMLVarchar](api:SilverStripe\ORM\FieldType\DBHTMLVarchar)).
-- `'Int'`: A 32-bit signed integer field (see: [DBInt](api:SilverStripe\ORM\FieldType\DBInt)).
-- `'Locale'`: A field for storing locales (see: [DBLocale](api:SilverStripe\ORM\FieldType\DBLocale)).
-- `'Money'`: Similar to Currency, but with localisation support (see: [DBMoney](api:SilverStripe\ORM\FieldType\DBMoney)).
-- `'MultiEnum'`: An enumeration set of strings that can store multiple values (see: [DBMultiEnum](api:SilverStripe\ORM\FieldType\DBMultiEnum)).
-- `'Percentage'`: A decimal number between 0 and 1 that represents a percentage (see: [DBPercentage](api:SilverStripe\ORM\FieldType\DBPercentage)).
-- `'PolymorphicForeignKey'`: A special ForeignKey class that handles relations with arbitrary class types (see: [DBPolymorphicForeignKey](api:SilverStripe\ORM\FieldType\DBPolymorphicForeignKey)).
-- `'PrimaryKey'`: A special type Int field used for primary keys. (see: [DBPrimaryKey](api:SilverStripe\ORM\FieldType\DBPrimaryKey)).
-- `'Text'`: A variable-length string of up to 2MB, designed to store raw text (see: [DBText](api:SilverStripe\ORM\FieldType\DBText)).
-- `'Time'`: A time field (see: [DBTime](api:SilverStripe\ORM\FieldType\DBTime)).
-- `'Varchar'`: A variable-length string of up to 255 characters, designed to store raw text (see: [DBVarchar](api:SilverStripe\ORM\FieldType\DBVarchar)).
-- `'Year'`: Represents a single year field (see: [DBYear](api:SilverStripe\ORM\FieldType\DBYear)).
+| Field | API link | Description | Validation |
+| --- | --- | --- | --- |
+| `BigInt` | [`DBBigInt`](api:SilverStripe\ORM\FieldType\DBBigInt) | An 8-byte signed integer field | Must be an int between -9223372036854775808 and 9223372036854775807 |
+| `Boolean` | [`DBBoolean`](api:SilverStripe\ORM\FieldType\DBBoolean) | A boolean field stored as a tinyint | Must be a boolean |
+| `Currency` | [`DBCurrency`](api:SilverStripe\ORM\FieldType\DBCurrency) | A number with 2 decimal points of precision, designed to store currency values | Must be a decimal |
+| `Date` | [`DBDate`](api:SilverStripe\ORM\FieldType\DBDate) | A date field | Must be a valid date in `Y-m-d` format |
+| `Datetime` | [`DBDatetime`](api:SilverStripe\ORM\FieldType\DBDatetime) | A date/time field | Must be a valid datetime in `Y-m-d H:i:s` format |
+| `ClassName` | [`DBClassName`](api:SilverStripe\ORM\FieldType\DBClassName) | A special enumeration for storing class names | Must be a valid FQCN of a `DataObject` subclass |
+| `ClassNameVarchar` | [`DBClassNameVarchar`](api:SilverStripe\ORM\FieldType\DBClassNameVarchar) | A special enumeration for storing class names in a `Varchar` field | Must be a valid FQCN of a `DataObject` subclass |
+| `Decimal` | [`DBDecimal`](api:SilverStripe\ORM\FieldType\DBDecimal) | A decimal number | Must be a decimal. |
+| `Double` | [`DBDouble`](api:SilverStripe\ORM\FieldType\DBDouble) | A floating point number with double precision | Must be numeric |
+| `Email` | [`DBEmail`](api:SilverStripe\ORM\FieldType\DBEmail) | An email field | Must be a valid email address |
+| `Enum` | [`DBEnum`](api:SilverStripe\ORM\FieldType\DBEnum) | An enumeration of a set of strings that can store a single value | Must be one of the defined values |
+| `Float` | [`DBFloat`](api:SilverStripe\ORM\FieldType\DBFloat) | A floating point number | Must be numeric |
+| `ForeignKey` | [`DBForeignKey`](api:SilverStripe\ORM\FieldType\DBForeignKey) | A special `Int` field used for foreign keys in `has_one` relationships | Must be an int |
+| `HTMLFragment` | [`DBHTMLText`](api:SilverStripe\ORM\FieldType\DBHTMLText) | A variable-length string of up to 2MB, designed to store HTML. Doesn't process [shortcodes](/developer_guides/extending/shortcodes/) | Must be a string |
+| `HTMLText` | [`DBHTMLText`](api:SilverStripe\ORM\FieldType\DBHTMLText) | A variable-length string of up to 2MB, designed to store HTML. Processes [shortcodes](/developer_guides/extending/shortcodes/) | Must be a string |
+| `HTMLVarchar` | [`DBHTMLVarchar`](api:SilverStripe\ORM\FieldType\DBHTMLVarchar) | A variable-length string of up to 255 characters, designed to store HTML. Can process [shortcodes](/developer_guides/extending/shortcodes/) with additional configuration | String must not be longer than specified length |
+| `Int` | [`DBInt`](api:SilverStripe\ORM\FieldType\DBInt) | A 32-bit signed integer field | Must be an int between -2147483648 and 2147483647 |
+| `IP` | [`DBIp`](api:SilverStripe\ORM\FieldType\DBIp) | An IP field | Must be a valid IP address, either IPv4 or IPv6 |
+| `Locale` | [`DBLocale`](api:SilverStripe\ORM\FieldType\DBLocale) | A field for storing locales | Must be a valid locale |
+| `Money` | [`DBMoney`](api:SilverStripe\ORM\FieldType\DBMoney) | A localised money field with a 3 character currency component, and an amount component. | Currency string must not be greater than 3 characters, amount must be a decimal |
+| `MutliEnum` | [`DBMultiEnum`](api:SilverStripe\ORM\FieldType\DBMultiEnum) | An enumeration set of strings that can store multiple values | Must be one of the allowable values |
+| `Percentage` | [`DBPercentage`](api:SilverStripe\ORM\FieldType\DBPercentage) | A decimal number between 0 and 1 that represents a percentage | Must be a decimal between 0 and 1 |
+| `PolymorphicForeignKey` | [`DBPolymorphicForeignKey`](api:SilverStripe\ORM\FieldType\DBPolymorphicForeignKey) | A special ForeignKey class that handles relations with arbitrary class types | Must be an int |
+| `PrimaryKey` | [`DBPrimaryKey`](api:SilverStripe\ORM\FieldType\DBPrimaryKey) | A special type Int field used for primary keys | Must be an int |
+| `Text` | [`DBText`](api:SilverStripe\ORM\FieldType\DBText) | A variable-length string of up to 2MB, designed to store raw text | Must be a string |
+| `Time` | [`DBTime`](api:SilverStripe\ORM\FieldType\DBTime) | A time field | Must be a valid time in `H:i:s` format |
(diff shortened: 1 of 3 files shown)


en/08_Changelogs/rc/5.3.0-rc1.md
```diff
@@ -101,7 +101,7 @@ A full list of module versions included in CMS Recipe 5.3.0-rc1 is provided belo
 
 </details>
 
-## Release Candidate
+## Release candidate
```
emteknetnz: A few linting fixes in this file

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
```diff
@@ -31,33 +31,41 @@ class Player extends DataObject
...
-- `'MultiEnum'`: An enumeration set of strings that can store multiple values (see: [DBMultiEnum](api:SilverStripe\ORM\FieldType\DBMultiEnum)).
-- `'Percentage'`: A decimal number between 0 and 1 that represents a percentage (see: [DBPercentage](api:SilverStripe\ORM\FieldType\DBPercentage)).
-- `'PolymorphicForeignKey'`: A special ForeignKey class that handles relations with arbitrary class types (see: [DBPolymorphicForeignKey](api:SilverStripe\ORM\FieldType\DBPolymorphicForeignKey)).
-- `'PrimaryKey'`: A special type Int field used for primary keys. (see: [DBPrimaryKey](api:SilverStripe\ORM\FieldType\DBPrimaryKey)).
-- `'Text'`: A variable-length string of up to 2MB, designed to store raw text (see: [DBText](api:SilverStripe\ORM\FieldType\DBText)).
-- `'Time'`: A time field (see: [DBTime](api:SilverStripe\ORM\FieldType\DBTime)).
-- `'Varchar'`: A variable-length string of up to 255 characters, designed to store raw text (see: [DBVarchar](api:SilverStripe\ORM\FieldType\DBVarchar)).
-- `'Year'`: Represents a single year field (see: [DBYear](api:SilverStripe\ORM\FieldType\DBYear)).
+| Field | Description | Validation | API Reference |
```
emteknetnz: Note for the sake of brevity, I have chosen not to define any conversions that take place in `setValue()`, for instance `'t'` would be converted to `true` for `DBBoolean`

en/08_Changelogs/rc/5.3.0-rc1.md
ChloeSartorelli: If you rebase this should be done already - I won't hold up merging for that though.

emteknetnz: Have rebased

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
```diff
@@ -31,33 +31,41 @@ class Player extends DataObject
...
+| --- | --- | --- | --- |
+| `BigInt` | An 8-byte signed integer field | Must be an int between -9223372036854775808 and 9223372036854775807 | [`DBBigInt`](api:SilverStripe\ORM\FieldType\DBBigInt) |
```
ChloeSartorelli: Either move the "API Reference" to be directly right of the "Field" column, or else just have the "Field" column references be links to the API

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
```diff
@@ -31,33 +31,41 @@ class Player extends DataObject
...
+| `Boolean` | A boolean field stored as a tinyint | Must be a boolean | [`DBBoolean`](api:SilverStripe\ORM\FieldType\DBBoolean) |
+| `Currency` | A number with 2 decimal points of precision, designed to store currency values | Must be a decimal | [`DBCurrency`](api:SilverStripe\ORM\FieldType\DBCurrency) |
+| `Date` | A date field | Must be a valid date in `Y-m-d` format | [`DBDate`](api:SilverStripe\ORM\FieldType\DBDate) |
+| `Datetime` | A date/time field | Must be a valid datetime in `Y-m-d H:i:s` format | [`DBDatetime`](api:SilverStripe\ORM\FieldType\DBDatetime) |
+| `DBClassName` | A special enumeration for storing class names | Must be a valid subclass name | [`DBClassName`](api:SilverStripe\ORM\FieldType\DBClassName) |
```
ChloeSartorelli:
```suggestion
| `DBClassName` | A special enumeration for storing class names | Must be a valid FQCN of a `DataObject` subclass | [`DBClassName`](api:SilverStripe\ORM\FieldType\DBClassName) |
```
Same with the varchar variant below

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
```diff
@@ -31,33 +31,41 @@ class Player extends DataObject
...
+| `DBClassNameVarchar` | A special enumeration for storing class names in a `Varchar` field | Must be a valid subclass name | [`DBClassNameVarchar`](api:SilverStripe\ORM\FieldType\DBClassNameVarchar) |
+| `Decimal` | A decimal number | Must be a decimal. | [`DBDecimal`](api:SilverStripe\ORM\FieldType\DBDecimal) |
+| `Double` | A floating point number with double precision | Not validated | [`DBDouble`](api:SilverStripe\ORM\FieldType\DBDouble) |
```
ChloeSartorelli: We don't validate it at all? Please use a relevant validator. Even just straight numeric validator would be better than nothing.

ChloeSartorelli: Looks like there are some others below that also lack validation. Please give them appropriate validation instead.

en/08_Changelogs/6.0.0.md
ChloeSartorelli, line 1: There's a non-zero chance someone has invalid values in the database right now, and when a content author tries to save their record, it'll give them validation errors that are:
a) not related to what they're actually trying to edit, and
b) not even presented in the form as values they can change

We should give some advice about how to handle this scenario.

GuySartorelli (changes requested):

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
emteknetnz: Updated

en/08_Changelogs/6.0.0.md
emteknetnz: Seems like if that happens then it would be a bug at our end?

I'm not sure how you'd get into that sort of state? The validation that we're adding *should* match up to what the underlying database column is specified to e.g. a Varchar(32) validator will have a max length of 32, and the underlying database column will have a max size of 32. Previously saving 48 characters into it would truncate the last 16 characters.

If someone did something like convert an existing column with data in it from Varchar to Email, then yeah that could cause issues, though that's not something we have any control over and I'd classify that as a developer implementation mistake

ChloeSartorelli: One example is if data is migrated from another source using low-level SQL queries - which is a fairly common practice.

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
emteknetnz: Have added NumericFieldValidator to DBFloat + DBDouble, will redo kitchen sink run shortly

DBText, DBHTMLText and DBHTMLFragment actually already has StringFieldValidator on them, have updated docs

en/08_Changelogs/6.0.0.md
emteknetnz: I'm still not sure why using low-level SQL queries would cause an issue, that's no different from using the ORM to insert values into the database in CMS 5 where there's no DBField validation

I don't know what you want me to write here, since I don't really know what the problem is, or how you're supposed to solve it.  Could you please suggest some text to write here, or we just leave as is?

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
ChloeSartorelli: Still hasn't been done or responded to

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
```diff
@@ -31,33 +31,41 @@ class Player extends DataObject
...
+| `HTMLText` | A variable-length string of up to 2MB, designed to store HTML. Processes [shortcodes](/developer_guides/extending/shortcodes/) | Must be a string | [`DBHTMLText`](api:SilverStripe\ORM\FieldType\DBHTMLText) |
+| `HTMLVarchar` | A variable-length string of up to 255 characters, designed to store HTML. Can process [shortcodes](/developer_guides/extending/shortcodes/) with additional configuration | String must not be longer than specified length | [`DBHTMLVarchar`](api:SilverStripe\ORM\FieldType\DBHTMLVarchar) |
+| `Int` | A 32-bit signed integer field | Must be an int between -2147483648 and 2147483647 | [`DBInt`](api:SilverStripe\ORM\FieldType\DBInt) |
+| `IP` | An IP field | Must be a valid IP address, either IPv4 or IPv6 | [`DBIp`](api:SilverStripe\ORM\FieldType\DBIp) |
+| `Locale` | A field for storing locales | Must be a valid locale | [`DBLocale`](api:SilverStripe\ORM\FieldType\DBLocale) |
+| `Money` | Similar to Currency, but with localisation support | Currency string must not be greater than 4 characters, amount must be a decimal | [`DBMoney`](api:SilverStripe\ORM\FieldType\DBMoney) |
```
ChloeSartorelli: Isn't it 4 digits including the numbers after the decimal place? e.g. `12.34` is 5 characters. Looking at the framework PR, there's nothing specific for `DBMoney` so why does it say here it has specific validation? Probably just update to use the same wording used for its parent class.

Sorry I didn't notice this before.

GuySartorelli (changes requested):

en/08_Changelogs/6.0.0.md
ChloeSartorelli: > I'm still not sure why using low-level SQL queries would cause an issue

If I migrate data from a varchar named `URL` from an old model into the column of a `DBUrl`, then if the value doesn't have a protocol for example that will be invalid.

This is just an example I can think of off the top of my head. There are probably plenty of possible scenarios where someone either gets into a situation like this or _is already in a situation like this_ where they have a value in their database which won't pass this new validation logic.

The validation wasn't there before. It will be there now. People might run into problems with data which is _already in their database_ that now prevents them from saving the dataobject. We need to give them _some_ advice about this.

Possible options we could given them:
1. We could implement a way to disable DBField validation globally (not a great idea but it's an option)
1. We could suggest writing a new `BuildTask` that updates invalid values if they encounter problems when they upgrade
1. I really don't know what else we could suggest but we should suggest something or at least point out that it's a possible scenario they could encounter

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
emteknetnz: Funny, I had it wrong it's actually 3 characters

```php
class DBMoney extends DBComposite
{
    private static array $composite_db = [
        'Currency' => 'Varchar(3)',
        'Amount' => 'Decimal(19,4)'
    ];
```
Currency refers to https://en.wikipedia.org/wiki/ISO_4217, though isn't validated that it that so it's just validated as 3 characters  - will update docs

GuySartorelli:
Please address all of the outstanding comments before reassigning this to me again.

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
emteknetnz: Updated

en/08_Changelogs/6.0.0.md
emteknetnz: Have added some additional text to say that when changing existing data to one of the new fields with validation you may wish to create a BuildTask to run validate on the data

ChloeSartorelli: That's good, but that's about making changes to data _after_ the migration.
My main concern is if data which is in their database in CMS 5 doesn't pass validation after upgrading to CMS 6.

emteknetnz: I think it's correct. Major upgrades / migrations are projects in themselves, so the build task would be run as part of the migration project, not after the migration

Example workflow for developer, assuming there's an existing Varchar MyEmail column with on a MyDataObject a coupe of invalid emails - website is probably not live at this point, is either on local, or in test environment
- Changes MyEmail from Varchar to Email in $db
- Run dev/build flush, actual db column remains a Varchar, DBField changes from DBVarchar to DBEmail
- Runs ValidateMyEmailTask - foreach MyDataObject::get() as $obj ... if !$obj->validate()->isValid() ... echo $obj['ID'] . ' ' . $obj['MyEmail'];
- Then does something about the invalid email address e.g. delete record, delete email address, etc

Ultimately it's the developers responsibility to ensure existing data is valid after adding an additional layer of validation on it

ChloeSartorelli: Ignore the new `DBField` classes for now.
In CMS 5 there is no validation for any DBField.
In CMS 6 there is now validation for most DBField classes.

There could be data in some column from CMS 5 that fails validation in CMS 6.

ChloeSartorelli: > Ultimately it's the developers responsibility to ensure existing data is valid after adding an additional layer of validation on it

We're the ones adding the validation, so it's our responsibility to consider the ramifications of adding that validation, and at the very _very_ least warning that there might be a scenario where old data doesn't meet the new validation criteria, and ideally suggesting a way to resolve that.

emteknetnz: > There could be data in some column from CMS 5 that fails validation in CMS 6.

I honestly don't know what scenario this happens, outside the a developer changing the column type as part of migration e.g. Varchar to Email

I honestly don't know what you want written. Please provide the exact text you want in the changelog and I'll put it in.

ChloeSartorelli: I think it would be better if we just discussed so that we're on the same page, but so be it.

> While we have tried to match the validation rules up to what would already have been stored in the database, there is a chance you'll find yourself with pre-existing data which doesn't meet the validation rules, and which therefore causes validation exceptions to be thrown the next time you try to save those records.
>
> If that happens, you may be able to resolve it with one of the following solutions:
> 1. If there is a form field for that database column, update the value in the form to a valid value before saving the record.
> 1. Write a `BuildTask` that updates any invalid values in your database.
> 1. <do we have a mechanism to turn that validation off?>

emteknetnz: Done

ChloeSartorelli: Please add back the bit about changing db fields. There's two scenarios there. Existing data that's bad, and changing DB fields to one of the new fields which results in invalid data.

```text
If you have an existing project that uses Varchar fields for email addresses, IP addresses, or URLs, and you switch to using one of the new DBField types, be aware that some of the values in the database may fail validation the next time they are saved which could cause issues for CMS editors. You may wish to create a [`BuildTask`](api:SilverStripe\Dev\BuildTask) that calls [`DataObject::validate()`](api:SilverStripe\ORM\DataObject::validate()) on all affected records.
```

Maybe even put this in a `> ![WARNING]` block.

emteknetnz: Done

GuySartorelli (approved):
LGTM

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
```diff
@@ -31,33 +31,41 @@ class Player extends DataObject
...
-- `'Percentage'`: A decimal number between 0 and 1 that represents a percentage (see: [DBPercentage](api:SilverStripe\ORM\FieldType\DBPercentage)).
-- `'PolymorphicForeignKey'`: A special ForeignKey class that handles relations with arbitrary class types (see: [DBPolymorphicForeignKey](api:SilverStripe\ORM\FieldType\DBPolymorphicForeignKey)).
-- `'PrimaryKey'`: A special type Int field used for primary keys. (see: [DBPrimaryKey](api:SilverStripe\ORM\FieldType\DBPrimaryKey)).
-- `'Text'`: A variable-length string of up to 2MB, designed to store raw text (see: [DBText](api:SilverStripe\ORM\FieldType\DBText)).
-- `'Time'`: A time field (see: [DBTime](api:SilverStripe\ORM\FieldType\DBTime)).
-- `'Varchar'`: A variable-length string of up to 255 characters, designed to store raw text (see: [DBVarchar](api:SilverStripe\ORM\FieldType\DBVarchar)).
-- `'Year'`: Represents a single year field (see: [DBYear](api:SilverStripe\ORM\FieldType\DBYear)).
+| Field | Description | Validation | API Reference |
+| --- | --- | --- | --- |
+| [`DBBigInt`](api:SilverStripe\ORM\FieldType\DBBigInt) | An 8-byte signed integer field | Must be an int between -9223372036854775808 and 9223372036854775807 |
```
ChloeSartorelli: Please keep the links, but remove the `DB` part from the text, or add back to old column (but keep it right next to this one)

GuySartorelli (changes requested):
Sorry, didn't realise you'd changed the config names to the class names in that table.

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
emteknetnz: Have added new column to the left for what's defined in the $db schema

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
```diff
@@ -31,33 +31,41 @@ class Player extends DataObject
...
-- `'MultiEnum'`: An enumeration set of strings that can store multiple values (see: [DBMultiEnum](api:SilverStripe\ORM\FieldType\DBMultiEnum)).
-- `'Percentage'`: A decimal number between 0 and 1 that represents a percentage (see: [DBPercentage](api:SilverStripe\ORM\FieldType\DBPercentage)).
-- `'PolymorphicForeignKey'`: A special ForeignKey class that handles relations with arbitrary class types (see: [DBPolymorphicForeignKey](api:SilverStripe\ORM\FieldType\DBPolymorphicForeignKey)).
-- `'PrimaryKey'`: A special type Int field used for primary keys. (see: [DBPrimaryKey](api:SilverStripe\ORM\FieldType\DBPrimaryKey)).
-- `'Text'`: A variable-length string of up to 2MB, designed to store raw text (see: [DBText](api:SilverStripe\ORM\FieldType\DBText)).
-- `'Time'`: A time field (see: [DBTime](api:SilverStripe\ORM\FieldType\DBTime)).
-- `'Varchar'`: A variable-length string of up to 255 characters, designed to store raw text (see: [DBVarchar](api:SilverStripe\ORM\FieldType\DBVarchar)).
-- `'Year'`: Represents a single year field (see: [DBYear](api:SilverStripe\ORM\FieldType\DBYear)).
+| Field | API link | Description | Validation | API Reference |
```
ChloeSartorelli: Table doesn't render correctly, I think there's more header than there are columns

GuySartorelli (changes requested):

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
```diff
@@ -31,33 +31,41 @@ class Player extends DataObject
...
+| `IP` | [`DBIp`](api:SilverStripe\ORM\FieldType\DBIp) | An IP field | Must be a valid IP address, either IPv4 or IPv6 |
+| `Locale` | [`DBLocale`](api:SilverStripe\ORM\FieldType\DBLocale) | A field for storing locales | Must be a valid locale |
+| `Money` | [`DBMoney`](api:SilverStripe\ORM\FieldType\DBMoney) | A localised money field with a 3 character currency component, and an amount component. | Currency string must not be greater than 3 characters, amount must be a decimal |
+| `MutliEnum` | [`DBMultiEnum`](api:SilverStripe\ORM\FieldType\DBMultiEnum) | An enumeration set of strings that can store multiple values | Must be one of the allowable values |
+| `Percentage` | [`DBPercentage`](api:SilverStripe\ORM\FieldType\DBPercentage) | A decimal number between 0 and 1 that represents a percentage | Must be a decimal between 0 and 1 |
+| `DBPolymorphicForeignKey` | [`DBPolymorphicForeignKey`](api:SilverStripe\ORM\FieldType\DBPolymorphicForeignKey) | A special ForeignKey class that handles relations with arbitrary class types | Must be an int |
```
ChloeSartorelli:
```suggestion
| `PolymorphicForeignKey` | [`DBPolymorphicForeignKey`](api:SilverStripe\ORM\FieldType\DBPolymorphicForeignKey) | A special ForeignKey class that handles relations with arbitrary class types | Must be an int |
```
I'm guessing - didn't double check the yml.

GuySartorelli (changes requested):

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
ChloeSartorelli: "API Reference" is a duplicate of "API link" probably

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
ChloeSartorelli: Have checked now since this wasn't fixed last round, and I'm correct, it shouldn't have the `DB`

emteknetnz: Opps, fixed

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
emteknetnz: Has been removed

GuySartorelli (approved):
LGTM

PR #636, opened 2024-11-23
Fix: Fixing documentation link

3Dgoo:


GuySartorelli (approved):
Thanks.
For future reference the correct commit message prefix for documentation changes is `DOC` as per https://docs.silverstripe.org/en/contributing/issues_and_bugs/#labels

PR #643, opened 2024-12-03
DOC Add PHP 8.4 support

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/331

I don't think we should add a changelog entry for this as it's PHP 8.4 support is simply something that CMS 6 comes with.

If we did want to mention this in the changelog then it seems like we'd need to mention that PHP 8.1 and 8.2 are no longer supported i.e. it's a change from CMS 5

### DIFF
--- a/en/00_Getting_Started/00_Server_Requirements.md
+++ b/en/00_Getting_Started/00_Server_Requirements.md
@@ -12,7 +12,7 @@ the server to update templates, website logic, and perform upgrades or maintenan
 
 ## PHP
 
-- PHP 8.3
+- PHP 8.3 - 8.4
 - PHP extensions: `ctype`, `dom`, `fileinfo`, `hash`, `intl`, `mbstring`, `session`, `simplexml`, `tokenizer`, `xml`
 - PHP configuration: `memory_limit` with at least `48M`
 - PHP extension for image manipulation: Either `gd` or `imagick`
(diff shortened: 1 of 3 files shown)


GuySartorelli:
> I don't think we should add a changelog entry for this as it's PHP 8.4 support is simply something that CMS 6 comes with.

The changelog is meant to point out changes between CMS 5.4 and CMS 6.0 - not just the breaking changes. I think it should be included.

> If we did want to mention this in the changelog then it seems like we'd need to mention that PHP 8.1 and 8.2 are no longer supported i.e. it's a change from CMS 5

That's a good idea.

en/00_Getting_Started/00_Server_Requirements.md
```diff
@@ -12,7 +12,7 @@ the server to update templates, website logic, and perform upgrades or maintenan
 
 ## PHP
 
-- PHP 8.3
+- PHP 8.3 or newer
```
ChloeSartorelli:
```suggestion
- PHP 8.3 - 8.4
```

Or to match the way it is in the [CMS 5 docs](https://docs.silverstripe.org/en/5/getting_started/server_requirements/):
```suggestion
- PHP >=8.3, <=8.4
```

8.5 isn't supported in 6.0.0, and won't be until 6.2.0 - and then anything newer that comes out won't be supported for 6.x at all.

GuySartorelli (changes requested):
The "PHP requirements for older Silverstripe CMS releases" section needs to be updated

emteknetnz:
I've added a small section to the 6.0.0 changelog in the same section that talks about mysql support changing

GuySartorelli (approved):
LGTM

PR #654, opened 2024-12-16
DOC Update validate signature

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11457

### DIFF
--- a/en/08_Changelogs/6.0.0.md
+++ b/en/08_Changelogs/6.0.0.md
@@ -1042,6 +1042,7 @@ As part of these changes [`ArrayList::find()`](api:SilverStripe\Model\List\Array
 
 ### General changes {#api-general}
 
+- [`DataObject::validate()`](api:SilverStripe\ORM\DataObject::validate()) now has an explicit `ValidationResult` return type.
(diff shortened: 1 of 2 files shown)


en/08_Changelogs/6.0.0.md
```diff
@@ -928,6 +929,33 @@ The following classes have been changed when code was moved to the `silverstripe
...
+    public function validate(): ValidationResult
+    {
+        $result = parent::validate();
+        if ($this->Title === 'Bad Title') {
+            $result->addError('Title cannot be "Bad Title"');
+        }
+        return $result;
+    }
+}
+```
```
ChloeSartorelli: I don't think this is worth its own section. There's a _lot_ of signature changes happening that don't get their own sections.
It might make sense for this to be a sub-section of "Validation added to DBFields" though, since this is a result of that work.

Alternatively "Validation added to DBFields" could have a parent "Changes to validation" heading, and we could move "Changes to password validation" and any other validation-related changelog entries in there.

GuySartorelli (changes requested):

en/08_Changelogs/6.0.0.md
emteknetnz: DataObject validate should not be a subsection of DBField validation IMO. Also the DBField work is new functionality, this is just a signature change on something presumably fairly high use.

I think I'll just delete this and leave it in as a single line entry. PHP will tell people quicky enough how to update their method signatures.

GuySartorelli (approved):
LGTM

PR #677, opened 2025-01-16
DOC Show more information in ValidationException message

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11423

### DIFF
--- a/en/08_Changelogs/6.0.0.md
+++ b/en/08_Changelogs/6.0.0.md
@@ -50,6 +50,7 @@ title: 6.0.0 (unreleased)
   - [Update JS MIME type, remove `type` in `<script>` tags](#js-mime-type-update)
   - [`getSchemaDataDefaults()` now includes attributes](#formfield-schema-data)
   - [Remember me token rotation](#remember-me-token-rotation)
+  - [Validation exceptions shows additional info in some contexts](#validation-exception-info)
 - [Full list of removed and changed API (by module, alphabetically)](#api-removed-and-changed)
 
 ## Change to commercially supported modules {#changes-to-support}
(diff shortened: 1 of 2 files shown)


en/08_Changelogs/6.0.0.md
```diff
@@ -1165,6 +1166,14 @@ If you were calling `getSchemaData()` in your `getAttributes()` method in a `For
 
 The related `onAfterRenewToken` extension hook has been renamed to `onAfterRenewSession`, and is triggered at the same logical step in the session renewal process.
 
+### Validation exceptions shows additional info in some contexts {#validation-exception-info}
+
+[`ValidationResult`](api:SilverStripe\Core\Validation\ValidationResult) now includes additional information about validation exceptions including the field name that failed validation, and if applicable the DataObject class and record ID. This additional information will be outputted by [`ValidationException`](api:SilverStripe\Core\Validation\ValidationException) if the validation exception happened in a CLI context, or in an HTTP context if the current controller is an instance of [`DevelopmentAdmin`](api:SilverStripe\Dev\DevelopmentAdmin) or a subclass of it.
```
ChloeSartorelli:
```suggestion
[`ValidationResult`](api:SilverStripe\Core\Validation\ValidationResult) now includes additional information about validation exceptions including the field name that failed validation, and if applicable the model class and record ID. This additional information will be included in the validation message by [`ValidationException`](api:SilverStripe\Core\Validation\ValidationException) if the validation exception happened in a CLI context, or in an HTTP context if the current controller is an instance of [`DevelopmentAdmin`](api:SilverStripe\Dev\DevelopmentAdmin) or a subclass of it.
```
There's no reason to lock this down to `DataObject` specifically.

ChloeSartorelli: May need to update the conditions noted here depending on code changes in the code PR.

GuySartorelli (changes requested):
Probably worth adding a note that out of the box the only validation messages which have this additional context are ones produced by a call to `DataObject::validate()`.

Also add some docs around the new API in the validation documentation section.

en/08_Changelogs/6.0.0.md
emteknetnz: Updated

emteknetnz:
> Also add some docs around the new API in the validation documentation section.

I don't know how to do this. The most relevant section would be [here](https://docs.silverstripe.org/en/5/developer_guides/model/validation/#model-validation) 

We could add something like this, but it seems like far too much basically irrelevant detail

`DataObject::validate()` will also call `ValidationResult::setModelClass()` and `ValidationResult::setRecordID()` with `static::class` and `$this->ID` respectively, this is done to provide context to the validation result and will be used to update the error messages with the model class and record ID when in a CLI context, or when the current controller is `DevelopmentAdmin` i.e. the controller used when running `dev/build`.

GuySartorelli:
I'm not saying we should document what `DataObject` does - but we should document that developers can call `setModelClass()` and `setRecordID()` on their own validation result instances in order to provide that information which will then be included in the `ValidationException`.

I wonder if we should also provide `setModelClass()` and `setRecordID()` for the exception class itself, so that direct validation exceptions don't _have_ to have a validation result object to benefit from that? But that's just an idle thought that you can feel free to ignore if you wish.

emteknetnz:
Updated

GuySartorelli (approved):
LGTM, will merge after code PRs

PR #684, opened 2025-02-10
DOC Form sudo mode

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1898

### DIFF
--- a/en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
+++ b/en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
@@ -6,11 +6,11 @@ icon: key
 
 # Sudo mode
 
-Sudo mode represents a heightened level of permission in that you are more certain that the current user is actually the person whose account is logged in. This is performed by re-validating that the account's password is correct, and will then last for a certain amount of time (configurable) until it will be checked again.
+Sudo mode represents a heightened level of permission in that you are more certain that the current CMS user is actually the person whose account is logged in. This is performed by re-validating that the account's password is correct, and will then last for a certain amount of time (configurable) until it will be checked again.
 
-Sudo mode will automatically be enabled for the configured lifetime when a user logs into the CMS. Note that if the PHP session lifetime expires before the sudo mode lifetime, that sudo mode will also be cleared (and re-enabled when the user logs in again). If the user leaves their CMS open, or continues to use it, for an extended period of time with automatic refreshing in the background, sudo mode will eventually deactivate once the max lifetime is reached.
+Sudo mode is not active when a user logs in to the CMS. If the PHP session lifetime expires before the sudo mode lifetime, that sudo mode will also be cleared. If the user leaves their CMS open, or continues to use it, for an extended period of time with automatic refreshing in the background, sudo mode will eventually deactivate once the max lifetime is reached.
 
-## Configuring the lifetime
+## Configuring the sudo mode lifetime
 
 The default [`SudoModeServiceInterface`](api:SilverStripe\Security\SudoMode\SudoModeServiceInterface) implementation is [`SudoModeService`](api:SilverStripe\Security\SudoMode\SudoModeService), and its lifetime can be configured with YAML. You should read the lifetime value using `SudoModeServiceInterface::getLifetime()`.
 
@@ -19,31 +19,52 @@ SilverStripe\Security\SudoMode\SudoModeService:
   lifetime_minutes: 25
 ```
 
-## Enabling sudo mode for controllers
+## Sudo mode for models
 
-You can add the `SudoModeServiceInterface` singleton as a dependency to a controller that requires sudo mode for one of its actions:
+Models which are protected by sudo mode (usually a [`DataObject`](api:SilverStripe\ORM\DataObject) subclass) are still viewable without entering the password. Any actions that modify data, e.g. POSTing an edit form submission, will require the user to enter their password. This is done to balance usability with security.
+
+The following `DataObject` subclasses are protected by sudo mode out of the box:
+
+- [`Member`](api:SilverStripe\Security\Member)
+- [`Group`](api:SilverStripe\Security\Group)
+- [`PermissionRole`](api:SilverStripe\Security\PermissionRole)
+- [`PermissionRoleCode`](api:SilverStripe\Security\PermissionRoleCode)
+
+### Configuring sudo mode for your models
+
+To add sudo mode for a particular model, including your `DataObject` subclasses, simply set the [`require_sudo_mode`](api:SilverStripe\View\ViewableData->require_sudo_mode) configuration property to `true`, either directly on the class or via yml.
+
+> [!NOTE]
+> This will only add sudo mode to edit forms within the CMS interface. It will have no effect on forms outside of the CMS, such as custom forms in the frontend.
 
 ```php
-namespace App\Control;
+namespace App\Model;
 
-class MyController extends Controller
+use SilverStripe\ORM\DataObject;
+
+class Product extends DataObject
 {
-    private ?SudoModeServiceInterface $sudoModeService = null;
-
-    private static array $dependencies = ['SudoModeService' => '%$' . SudoModeServiceInterface::class];
-
-    public function setSudoModeService(SudoModeServiceInterface $sudoModeService): static
-    {
-        $this->sudoModeService = $sudoModeService;
-        return $this;
-    }
+    // ...
+    private static bool $require_sudo_mode = true;
 }
 ```
 
-Performing a sudo mode verification check in a controller action is simply using the service to validate the request:
+```yml
+SomeModule\Model\Player:
+  require_sudo_mode: true
+```
+
+> [!WARNING] If you add new fields to form that is protected by sudo mode, such as in an overridden [`LeftAndMain::getEditForm()`](api:SilverStripe\Admin\LeftAndMain::getEditForm()) method, it may mean having to call [`Form::requireSudoMode()`](api:SilverStripe\Forms\Form::requireSudoMode()) on the form to ensure the newly added fields are set to read-only while sudo mode is inactive.
+
+## Sudo mode for controller endpoints
+
+Performing a sudo mode verification check in a controller endpoint by using the sudo mode service to validate the request:
 
 ```php
 namespace App\Control;
+
+use SilverStripe\Core\Injector\Injector;
+use SilverStripe\Security\SudoMode\SudoModeServiceInterface;
 
 class MyController extends Controller
 {
@@ -51,7 +72,8 @@ class MyController extends Controller
 
     public function myAction(HTTPRequest $request): HTTPResponse
     {
-        if (!$this->sudoModeService->check($request->getSession())) {
+        $service = Injector::inst()->get(SudoModeServiceInterface::class);
+        if (!$service->check($request->getSession())) {
             return $this->httpError(403, 'Sudo mode is required for this action');
         }
         // ... continue with sensitive operations
@@ -59,14 +81,20 @@ class MyController extends Controller
 }
 ```
 
-## Using sudo mode in a react component
+## Sudo mode for react components
 
 The `silverstripe/admin` module defines a [React Higher-Order-Component](https://reactjs.org/docs/higher-order-components.html) (aka HOC) which can
 be applied to React components in your module or code to intercept component rendering and show a "sudo mode required"
 information and log in screen, which will validate, activate sudo mode, and re-render the wrapped component afterwards
 on success.
 
+This sudo mode differs from sudo mode for `DataObject`s in that it is not tied to a specific model, but rather to a
+specific react component.
+
 > [!WARNING]
+> Sudo mode for react components does not protect the data the component manages, or any endpoints the component uses, it simply requires the user to re-enter their password before the component is rendered.
+
+> [!IMPORTANT]
 > The `WithSudoMode` HOC is exposed via [Webpack's expose-loader plugin](https://webpack.js.org/loaders/expose-loader/). You will need to add it as a [webpack external](https://webpack.js.org/configuration/externals/) to use it. The recommended way to do this is via the [@silverstripe/webpack-config npm package](https://www.npmjs.com/package/@silverstripe/webpack-config) which handles all the external configuration for you.
 
 You can get the injector to apply the HOC to your component automatically using [injector transformations](/developer_guides/customising_the_admin_interface/reactjs_redux_and_graphql/#transforming-services-using-middleware):
--- a/en/08_Changelogs/5.4.0.md
+++ b/en/08_Changelogs/5.4.0.md
@@ -8,6 +8,7 @@ title: 5.4.0 (unreleased)
 
 - [Security considerations](#security-considerations)
 - [Features and enhancements](#features-and-enhancements)
+  - [Sudo mode for sensitive data](#form-sudo-mode)
   - [Logged warning if allowed hosts have not been configured](#allowed-hosts-warning)
   - [New `XssSanitiser` class](#new-xsssanitiser-class)
   - [Option to change `ClassName` column from enum to varchar](#classname-varchar)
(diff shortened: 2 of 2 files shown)


en/08_Changelogs/5.4.0.md
```diff
@@ -32,6 +33,26 @@ We have provided a severity rating of the vulnerabilities below based on the CVS
 
 ## Features and enhancements
 
+### Sudo mode for sensitive data {#form-sudo-mode}
+
+Some data managed by the CMS is always considered sensitive from a security point of view, such as member data and permissions assigned to groups. Data of this nature is now protected by default by "sudo mode" in the CMS. When a user tries to edit sensitive data, they will be prompted to enter their password to confirm their identity.
+
+This change was made to provide an extra layer of protection against cross site scripting (XSS) attacks, as well as people maliciously using someone elses computer left unattended in a logged-in state.
+
+Users will still be able to view sensitive data without entering their password as they could before. Previously "sudo mode" only protected a members MFA settings. Now the following [`DataObject`](api:SilverStripe\ORM\DataObject) subclasses are now protected by sudo mode:
```
ChloeSartorelli:
```suggestion
Users will still be able to view sensitive data without entering their password as they could before. Previously "sudo mode" only protected a member's MFA settings. Now the following [`DataObject`](api:SilverStripe\ORM\DataObject) subclasses are also protected by sudo mode:
```

en/08_Changelogs/5.4.0.md
```diff
@@ -32,6 +33,26 @@ We have provided a severity rating of the vulnerabilities below based on the CVS
...
+
+- [`Member`](api:SilverStripe\Security\Member)
+- [`Group`](api:SilverStripe\Security\Group)
+- [`PermissionRole`](api:SilverStripe\Security\PermissionRole)
+- [`PermissionRoleCode`](api:SilverStripe\Security\PermissionRoleCode)
+- [`SiteConfig`](api:SilverStripe\SiteConfig\SiteConfig)
+
+You can also add sudo mode to your own `DataObject` subclass by setting the [`DataObject.require_sudo_mode`](api:SilverStripe\ORM\DataObject->require_sudo_mode) configuration property to `true`.
```
ChloeSartorelli:
````suggestion
You can also add sudo mode to your own `DataObject` subclass by setting the [`DataObject.require_sudo_mode`](api:SilverStripe\ORM\DataObject->require_sudo_mode) configuration property to `true`.

```php
namespace App\Model;

use SilverStripe\ORM\DataObject;

class Product extends DataObject
{
    // ...
    private static bool $require_sudo_mode = true;
}
```
````
Just to avoid anyone misunderstanding to think they have to set it on `DataObject` directly.
Alternatively show the example in YAML instead.

File: en/08_Changelogs/5.4.0.md [hunk 3]
```diff
@@ -32,6 +33,26 @@ We have provided a severity rating of the vulnerabilities below based on the CVS
...
+
+Previously sudo mode was automatically activated when a user logged in, which largely negated the value the feature provided. This has now been changed so that this no longer happens and users will always have to enter their password to activate sudo mode.
```
ChloeSartorelli:
```suggestion
Previously sudo mode was automatically activated when a user logged in, which largely negated the value the feature provided. This has now been changed so that users will always have to enter their password to activate sudo mode the first time they reach a sensitive area of the CMS.

There is still a configurable grace period where, after entering your password to activate sudo mode, it will remain active for a short time.
```

en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
```diff
@@ -6,9 +6,28 @@ icon: key
...
 
-Sudo mode will automatically be enabled for the configured lifetime when a user logs into the CMS. Note that if the PHP session lifetime expires before the sudo mode lifetime, that sudo mode will also be cleared (and re-enabled when the user logs in again). If the user leaves their CMS open, or continues to use it, for an extended period of time with automatic refreshing in the background, sudo mode will eventually deactivate once the max lifetime is reached.
+Sudo mode is not active when a user logs in to the CMS. If the PHP session lifetime expires before the sudo mode lifetime, that sudo mode will also be cleared. If the user leaves their CMS open, or continues to use it, for an extended period of time with automatic refreshing in the background, sudo mode will eventually deactivate once the max lifetime is reached.
+
+Data protected by sudo mode is still viewable without entering the password, but any actions that modify data will require the user to enter their password. This is done to balance usability with security. The following [`DataObject`](api:SilverStripe\ORM\DataObject) subclasses are protected by sudo mode:
```
ChloeSartorelli: > Data protected by sudo mode is still viewable without entering the password

Is that true for the per-field sudo mode as well? I think it isn't, right? The HOC just replaces the field outright?

We need to remember (and point out) that there's now two ways to use sudo mode - on a whole data object, or just for a single form field - and they work subtly differently.

Might make sense to just outright separate them into two separate headings:

1. ``## Protecting an entire model``
1. ``## Protecting a single `FormField` ``

en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
```diff
@@ -6,9 +6,28 @@ icon: key
...
+## Adding or removing sudo mode for a `DataObject`
+
+To add sudo mode for a particular `DataObject`, simply set the [`DataObject.require_sudo_mode`](api:SilverStripe\ORM\DataObject->require_sudo_mode) configuration property to `true`, either via yml or directly on the class.
+
+While it's not recommended, to remove sudo mode for a particular `DataObject`, set the [`DataObject.require_sudo_mode`](api:SilverStripe\ORM\DataObject->require_sudo_mode) configuration property to `false` via yml.
+
+```yml
+SilverStripe\Security\Member:
+  require_sudo_mode: false
+```
```
ChloeSartorelli:
````suggestion
To add sudo mode for a particular `DataObject`, simply set the [`DataObject.require_sudo_mode`](api:SilverStripe\ORM\DataObject->require_sudo_mode) configuration property to `true`, either via yml or directly on the class.

```php
namespace App\Model;
use SilverStripe\ORM\DataObject;
class Product extends DataObject
{
    // ...
    private static bool $require_sudo_mode = true;
}
```

```yml
SomeModule\Model\Player:
  require_sudo_mode: true
```
````
I don't think we should tell people they can disable this and put that idea into their heads. Saying how to configure it in the broader sense (i.e. this is controlled by `DataObject.require_sudo_mode` which you can configure) should be enough to tell people how to turn it off without telling them "you can turn it off"

File: en/02_Developer_Guides/09_Security/04_Sudo_Mode.md [hunk 6]
```diff
@@ -19,39 +38,24 @@ SilverStripe\Security\SudoMode\SudoModeService:
   lifetime_minutes: 25
 ```
 
-## Enabling sudo mode for controllers
+## Enabling sudo mode for controller endpoints
```
ChloeSartorelli: Just as a sanity check in case you didn't already - please quickly check if there are any links in the docs to the old anchor, and update them to the new anchor if there are.

en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
```diff
@@ -19,39 +38,24 @@ SilverStripe\Security\SudoMode\SudoModeService:
...
-
-    private static array $dependencies = ['SudoModeService' => '%$' . SudoModeServiceInterface::class];
-
-    public function setSudoModeService(SudoModeServiceInterface $sudoModeService): static
-    {
-        $this->sudoModeService = $sudoModeService;
-        return $this;
-    }
-}
-```
```
ChloeSartorelli: Why is this example being removed?

GuySartorelli (changes requested):

en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
emteknetnz: The HOC is basically irrelevant at this point as it only used for the MFA field on the Member form, though the entire Member form is now protected.  You need to enable sudo mode on the form to see the HOC ... which won't be used because sudo mode is now activated

There may be some edge case where a project/module somehow uses the HOC for something else ... though my assumption is that there are zero instances of that so mentioning that here is just adding confusion for no benefit

en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
emteknetnz: Updated

en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
emteknetnz: Checked and did not find any

en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
emteknetnz: I really dislike the `$dependencies` way of doing it. It's long and unintuitive. I updated the example below to:

```php
$service = Injector::inst()->get(SudoModeServiceInterface::class);
if (!$service->check($request->getSession())) {
// ...
```

en/08_Changelogs/5.4.0.md
emteknetnz: Updated

en/08_Changelogs/5.4.0.md
emteknetnz: Updated

en/08_Changelogs/5.4.0.md
emteknetnz: Updated

en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
```diff
@@ -6,52 +6,69 @@ icon: key
...
+```
+
+```yml
+SomeModule\Model\Player:
+  require_sudo_mode: true
+```
 
-    private static array $dependencies = ['SudoModeService' => '%$' . SudoModeServiceInterface::class];
+> [!NOTE]
+> This will only add sudo mode to `DataObject` edit forms within the CMS interface. It will have no effect on forms outside of the CMS, such as custom forms in the frontend.
```
ChloeSartorelli:
```suggestion
> This will only add sudo mode to edit forms within the CMS interface. It will have no effect on forms outside of the CMS, such as custom forms in the frontend.
```

en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
```diff
@@ -6,52 +6,69 @@ icon: key
...
+- [`Group`](api:SilverStripe\Security\Group)
+- [`PermissionRole`](api:SilverStripe\Security\PermissionRole)
+- [`PermissionRoleCode`](api:SilverStripe\Security\PermissionRoleCode)
+- [`SiteConfig`](api:SilverStripe\SiteConfig\SiteConfig)
 
-## Enabling sudo mode for controllers
+## Adding or removing sudo mode for a `DataObject`
 
-You can add the `SudoModeServiceInterface` singleton as a dependency to a controller that requires sudo mode for one of its actions:
+To add sudo mode for a particular `DataObject`, simply set the [`DataObject.require_sudo_mode`](api:SilverStripe\ORM\DataObject->require_sudo_mode) configuration property to `true`, either directly on the class or via yml.
```
ChloeSartorelli:
```suggestion
To add sudo mode for a particular model, including your `DataObject` subclasses, simply set the [`require_sudo_mode`](api:SilverStripe\View\ViewableData->require_sudo_mode) configuration property to `true`, either directly on the class or via yml.
```

Hopefully this strikes a good balance between being accurate (the API doc link should be for where the config is defined) and not confusing (we don't actually mention `ViewableData` in the copy directly)

en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
ChloeSartorelli: The HOC is not irrelevant. There are plenty of scenarios where a project or module might want to protect a single sensitive form field with sudo mode. Just because we're not using it directly anymore doesn't make it useless. It's also _still there_ so should be documented regardless of our feelings about it until/if we decide to remove it, which would be a separate card.

> mentioning that here is just adding confusion for no benefit

I also don't see how there would be any confusion provided you use separate sections like I suggested above for the two use cases.

en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
ChloeSartorelli: I personally think we should keep it, and that this change is unrelated - but in the interest of moving forward I'll let this go

GuySartorelli (changes requested):

en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
emteknetnz: There's currently a "Using sudo mode in a react component" section lower down in this file, which is instructions to use the HOC, which seems like the "protecting a single FormField", scenario, but it's not really.  The MFA field is really a separate xhr field that not part of the form, i.e there's no underlying `<input>` that submitted as part of main form submission.  There's currently no way to have sudo mode on a single regular form field.

I don't think there needs to be any changes made here

en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
emteknetnz: Updated

en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
emteknetnz: Updated

en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
ChloeSartorelli: There's still a question outstanding which was the start of this whole thread:

>> Data protected by sudo mode is still viewable without entering the password
>
> Is that true for the per-field sudo mode as well? I think it isn't, right? The HOC just replaces the field outright?

If we're not splitting this up into clear sections that make it obvious when we're talking about the per-field implementation vs the entire form implementation, then we need to make it clear in the writing what applies to what. "sudo mode" is the name for both implementations, so we can't just blanket say "sudo mode won't hide data"

GuySartorelli:
There's a merge conflict to resolve here

en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
emteknetnz: I've mostly reorganized the existing headings to make the split clearer

en/08_Changelogs/5.4.0.md
ChloeSartorelli, line 1: Need to add a note that any end-to-end tests editing anything that's now protected with sudo mode will need to be updated to fill in the form or disable sudo, and link to the "disabling sudo mode" behat docs.

en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
```diff
@@ -6,67 +6,93 @@ icon: key
...
+> [!WARNING] If you add new fields to form that is protected by sudo mode, such as in an overridden [`LeftAndMain::getEditForm()`](api:SilverStripe\Admin\LeftAndMain::getEditForm()) method, it may mean having to call [`Form::requireSudoMode()`](api:SilverStripe\Forms\Form::requireSudoMode()) on the form to ensure the newly added fields are set to read-only while sudo mode is inactive.
+
+## Configuring the sudo mode lifetime
+
+The default [`SudoModeServiceInterface`](api:SilverStripe\Security\SudoMode\SudoModeServiceInterface) implementation is [`SudoModeService`](api:SilverStripe\Security\SudoMode\SudoModeService), and its lifetime can be configured with YAML. You should read the lifetime value using `SudoModeServiceInterface::getLifetime()`.
+
+```yml
+SilverStripe\Security\SudoMode\SudoModeService:
+  lifetime_minutes: 25
 ```
```
ChloeSartorelli: This section shouldn't be splitting up the various "sudo mode for" sections. Please move it to be either before or after that group of sections.

en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
```diff
@@ -6,67 +6,93 @@ icon: key
...
-    public function setSudoModeService(SudoModeServiceInterface $sudoModeService): static
-    {
-        $this->sudoModeService = $sudoModeService;
-        return $this;
-    }
-}
+> [!NOTE]
+> This will only add sudo mode to edit forms within the CMS interface. It will have no effect on forms outside of the CMS, such as custom forms in the frontend.
+
+> [!WARNING] If you add new fields to form that is protected by sudo mode, such as in an overridden [`LeftAndMain::getEditForm()`](api:SilverStripe\Admin\LeftAndMain::getEditForm()) method, it may mean having to call [`Form::requireSudoMode()`](api:SilverStripe\Forms\Form::requireSudoMode()) on the form to ensure the newly added fields are set to read-only while sudo mode is inactive.
```
ChloeSartorelli: IIRC our docs site doesn't like two callout blocks being right after one another. Consider putting one of these above the code blocks.

en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
```diff
@@ -6,67 +6,93 @@ icon: key
...
+Sudo mode is not active when a user logs in to the CMS. If the PHP session lifetime expires before the sudo mode lifetime, that sudo mode will also be cleared. If the user leaves their CMS open, or continues to use it, for an extended period of time with automatic refreshing in the background, sudo mode will eventually deactivate once the max lifetime is reached.
 
-## Configuring the lifetime
+## Sudo mode for models
 
-The default [`SudoModeServiceInterface`](api:SilverStripe\Security\SudoMode\SudoModeServiceInterface) implementation is [`SudoModeService`](api:SilverStripe\Security\SudoMode\SudoModeService), and its lifetime can be configured with YAML. You should read the lifetime value using `SudoModeServiceInterface::getLifetime()`.
+The data of models (usually a [`DataObject`](api:SilverStripe\ORM\DataObject) subclass) are protected by sudo mode is still viewable without entering the password, and any actions that modify data, e.g. POSTing an edit form submission, will require the user to enter their password. This is done to balance usability with security. The following `DataObject` subclasses are protected by sudo mode:
```
ChloeSartorelli:
```suggestion
Modules which are protected by sudo mode (usually a [`DataObject`](api:SilverStripe\ORM\DataObject) subclass) are still viewable without entering the password. Any actions that modify data, e.g. POSTing an edit form submission, will require the user to enter their password. This is done to balance usability with security.

The following `DataObject` subclasses are protected by sudo mode out of the box:
```
Just rewording/rearranging this to be a bit clearer

en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
```diff
@@ -6,67 +6,93 @@ icon: key
...
-SilverStripe\Security\SudoMode\SudoModeService:
-  lifetime_minutes: 25
-```
+- [`Member`](api:SilverStripe\Security\Member)
+- [`Group`](api:SilverStripe\Security\Group)
+- [`PermissionRole`](api:SilverStripe\Security\PermissionRole)
+- [`PermissionRoleCode`](api:SilverStripe\Security\PermissionRoleCode)
 
-## Enabling sudo mode for controllers
+### Configuring sudo mode for a `DataObject`
```
ChloeSartorelli:
```suggestion
### Configuring sudo mode for your models
```

GuySartorelli (changes requested):

en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
emteknetnz: Have moved it above the other sections

en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
emteknetnz: Updated

en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
emteknetnz: Done, the first word "Modules" I assumed you meant "Models"

en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
emteknetnz: Updated

emteknetnz:
> Need to add a note that any end-to-end tests editing

Done

en/02_Developer_Guides/09_Security/04_Sudo_Mode.md
ChloeSartorelli: Yes, sorry. Thanks for catching that.

GuySartorelli (approved):
LGTM

PR #687, opened 2025-02-11
DOC Document new filter by range functionality

GuySartorelli:
## Issue
- https://github.com/silverstripe/silverstripe-cms/issues/2949


en/02_Developer_Guides/00_Model/11_Scaffolding.md
```diff
@@ -25,6 +25,7 @@ use SilverStripe\ORM\DataObject;
 
 class MyDataObject extends DataObject
 {
+    // ...
```
ChloeSartorelli, line 28: Adding these where appropriate in this file to indicate we're skipping some best practices for the sake of brevity - we tend to do this with new code blocks, so this just gives consistency across this file.

en/02_Developer_Guides/00_Model/11_Scaffolding.md
```diff
@@ -433,15 +442,18 @@ use SilverStripe\ORM\DataObject;
...
     private static $searchable_fields = [
         'Name' => 'PartialMatchFilter',
-        'ProductCode' => NumericField::class,
+        'ProductCode' => [
+            'field' => NumericField::class,
+        ],
```
ChloeSartorelli, line 450: I don't know if this worked at some stage and broke, but it definitely doesn't work right now and shouldn't be documented as a thing that works.

en/02_Developer_Guides/00_Model/11_Scaffolding.md
```diff
@@ -515,7 +586,7 @@ class Player extends DataObject
 Use a single search field that matches on multiple database fields with `'match_any'`. This also supports specifying a `FormField` and a filter, though it is not necessary to do so.
 
 > [!CAUTION]
-> If you don't specify a `FormField`, you must use the name of a real database field as the array key instead of a custom name so that a default field class can be determined.
+> If you don't specify `field` or `dataType`, you must use the name of a real database field as the array key instead of a custom name so that a default field class can be determined.
```
ChloeSartorelli, line 589: `dataType` is new and was intended for use with the range stuff, but it also works here so makes sense to document this option.

en/08_Changelogs/6.0.0.md
```diff
@@ -596,6 +597,40 @@ SilverStripe\UserForms\Model\UserDefinedForm:
 
 </details>
 
+### Filter within a range with `searchable_fields` {#searchable-withinrange}
+
+Using the `searchable_fields` configuration, you can now declare that a field should be filtered using a range. This works out of the box with the numeric, date, datetime, and time fields that come in Silverstripe framework.
```
```suggestion
Using the `searchable_fields` configuration, you can now declare that a field should be filtered using a range. This works out of the box with the numeric, date, datetime, and time fields that come in the Silverstripe framework.
```

en/02_Developer_Guides/00_Model/11_Scaffolding.md
```diff
@@ -466,6 +479,62 @@ class MyDataObject extends DataObject
...
+namespace App\Model;
+
+use SilverStripe\ORM\DataObject;
+SilverStripe\ORM\Filters\WithinRangeFilter
+
+class MyDataObject extends DataObject
+{
+    // ...
+    private static array $db = [
+        'Price' => 'Currency',
```
emteknetnz, line 496: The currency filter currently seems very busted, even with the framework PR. There is a card open to fix at least some of that issues there, though the PR change this example use an Int instead so when people copy it they don't think that they've done something wrong

emteknetnz (changes requested):

en/02_Developer_Guides/00_Model/11_Scaffolding.md
ChloeSartorelli: I'd rather just fix the field. If it's just the validation thing, I'll pull that card into in progress and fix it. If there are more problems, please open cards for them.

emteknetnz: Noted on https://github.com/silverstripe/silverstripe-framework/pull/11602#issuecomment-2661684309 it's related to the framework PR, so should be fixed there

en/08_Changelogs/6.0.0.md
ChloeSartorelli: Done

emteknetnz (approved):

PR #690, opened 2025-02-12
DOC Document new versioned methods

GuySartorelli:
## Issue
- https://github.com/silverstripe/silverstripe-cms/issues/2949


emteknetnz (changes requested):
As per https://github.com/silverstripe/silverstripe-versioned/pull/489#pullrequestreview-2625574268 I do not think we should add any of these new methods

emteknetnz (approved):

PR #691, opened 2025-02-12
DOC Document new tinymce module

GuySartorelli:
Docs related to TinyMCE specifically have been either generalised or moved into the TinyMCE module.

A lot of the docs in the `HTMLEditorField` page have just been rearranged, so check if content already exists but just moved before suggesting changes.

## Issue
- https://github.com/silverstripe/silverstripe-htmleditor-tinymce/issues/2


en/02_Developer_Guides/03_Forms/Field_types/03_HTMLEditorField.md
```diff
@@ -42,236 +44,218 @@ class MyObject extends DataObject
...
-will use the configuration with the name 'myConfig'.
+There can be multiple configs, each with a unique string identifier, which should always be created and accessed using [`HTMLEditorConfig::get()`](api:SilverStripe\Forms\HTMLEditor\HTMLEditorConfig::get()). You can
+then set the currently active config using `HTMLEditorConfig::set_active()`.
 
-You can also specify which `HTMLEditorConfig` to use on a per field basis via the construct argument.
-This is particularly useful if you need different configurations for multiple `HTMLEditorField` on the same page or form.
+> [!WARNING]
+> The order in which the `_config.php` files are executed depends on the module names. Execution
+> order is alphabetical, so if you set an option in the `aardvark/_config.php` (i.e. the module name is simply `aardvark`),
+> this will be overridden when setting the same named option for the same named config in `vendor/silverstripe/admin/_config.php` (because the module name is `silverstripe/admin`) and your modification will disappear.
```
GuySartorelli, line 59: This warning just moved up from below.

en/02_Developer_Guides/03_Forms/Field_types/03_HTMLEditorField.md
```diff
@@ -444,32 +408,148 @@ elements (and back out in case an existing element gets edited).
...
+If you want your editor to be the default for all `HTMLEditorConfig`, set it through the injector configuration:
+
+```yml
+SilverStripe\Core\Injector\Injector:
+  SilverStripe\Forms\HTMLEditor\HTMLEditorConfig:
+    class: 'App\Forms\CustomHTMLEditorConfig'
 ```
 
-## Developing a wrapper to use a different WYSIWYG editors with `HTMLEditorField`
+If you want your editor to be the default specifically for the `cms` config, set the `HTMLEditorConfig.default_config_definitions` configuration like so:
```
ChloeSartorelli, line 449: This section is admittedly more for us than for anyone else, so we know what needs to be done when we find a suitable replacement for TinyMCE.
But hey, someone might want to slap CKEditor in there or something, and the system has been suitably abstracted enough for that to be doable now.

en/02_Developer_Guides/08_Performance/06_ORM.md
```diff
@@ -21,7 +21,7 @@ SilverStripe\Forms\TreeDropdownField:
 ```
 
 > [!TIP]
-> A very common use of `TreeDropdownField` is the "Insert Link" feature in the TinyMCE WYSIWYG. Setting this configuration to use another filter and adding an index on `Title` and `MenuTitle` for [`SiteTree`](api:SilverStripe\CMS\Model\SiteTree) can significantly improve performance.
+> A very common use of `TreeDropdownField` is the "Insert Link" feature used by supported HTML editors. Setting this configuration to use another filter and adding an index on `Title` and `MenuTitle` for [`SiteTree`](api:SilverStripe\CMS\Model\SiteTree) can significantly improve performance.
```
ChloeSartorelli, line 24: I'm making an assumption here that a replacement for TinyMCE will have similar plugins which use the same forms and modals our TinyMCE ones do. I think that's a safe assumption.

en/02_Developer_Guides/15_Customising_the_Admin_Interface/08_Typography.md
ChloeSartorelli, line 1: Moved wholesale to the new module

en/08_Changelogs/6.0.0.md
```diff
@@ -107,6 +109,17 @@ The code in `silverstripe/externallinks`, `silverstripe/security-report`, and `s
 
 CMS 6 compatible versions of `silverstripe/blog`, `silverstripe/subsites`, and `silverstripe/crontask` have been released with CMS 6.0.0, though they are not commercially supported.
 
+### New supported modules
+
+Some new modules have been created and are now covered by our [supported modules policy](/project_governance/supported_modules/). These modules will be supported for the full lifetime of Silverstripe CMS 6.
+
+> [!WARNING]
+> To continue using TinyMCE as an HTML editor in your existing project, you must add `silverstripe/htmleditor-tinymce` as an explicit dependency.
```
ChloeSartorelli, line 118: Intentionally putting this warning both here AND below so people have plenty of opportunity to see it.

en/08_Changelogs/6.0.0.md
```diff
@@ -1139,6 +1152,50 @@ SilverStripe\ORM\Connect\MySQLDatabase:
...
+See [defining HTML editor configurations](/developer_guides/forms/field_types/htmleditorfield/#defining-html-editor-configurations) for details about the new generalised API, and [optional features > TinyMCE HTML editor](/optional_features/htmleditor-tinymce/) for any TinyMCE-specific documentation.
+
+#### If you want TinyMCE 7 anyway
+
+If you want to use TinyMCE 7 in your own projects, and have done your due diligence to validate that the new license won't cause you any legal problems, you can theoretically do that.
+
+> [!WARNING]
+> Using TinyMCE 7 with Silverstripe CMS isn't supported and has not been tested.
+>
+> The plugins, skin, and configuration that come with `silverstripe/htmleditor-tinymce` are not guaranteed to work with TinyMCE 7.
```
ChloeSartorelli: This is a question that people are likely to ask - I think it's best to just answer it outright and say it's theoretically doable, but not supported.

en/08_Changelogs/6.0.0.md
```diff
@@ -1139,6 +1152,50 @@ SilverStripe\ORM\Connect\MySQLDatabase:
...
+
+Part of this change includes a new way to define rules such as which elements and attributes are allowed in HTML content. You can continue to set the `valid_elements` and `extended_valid_elements` configuration if you want to, but the new API is generic and will work for any `HTMLEditorConfig` implementation and can be defined in YAML configuration.
+
+See [defining HTML editor configurations](/developer_guides/forms/field_types/htmleditorfield/#defining-html-editor-configurations) for details about the new generalised API, and [optional features > TinyMCE HTML editor](/optional_features/htmleditor-tinymce/) for any TinyMCE-specific documentation.
+
+#### If you want TinyMCE 7 anyway
```
emteknetnz, line 1257: We don't support this, we haven't tested this, so I don't see why we're providing any instructions for how to do this. It's feels like it sending mixed messages, as well as needlessly making people read and comprehend more.

This should only be a single sentence saying that's it's unsupported, untested, though theoretically possible, however you're entirely on your own.

en/08_Changelogs/6.0.0.md
```diff
@@ -107,6 +109,17 @@ The code in `silverstripe/externallinks`, `silverstripe/security-report`, and `s
...
+### New supported modules
+
+Some new modules have been created and are now covered by our [supported modules policy](/project_governance/supported_modules/). These modules will be supported for the full lifetime of Silverstripe CMS 6.
+
+> [!WARNING]
+> To continue using TinyMCE as an HTML editor in your existing project, you must add `silverstripe/htmleditor-tinymce` as an explicit dependency.
+
+- [`silverstripe/startup-theme`](https://packagist.org/packages/silverstripe/startup-theme). See [new default front-end theme](#theme) below for more details.
+- [`silverstripe/template-engine` module](https://packagist.org/packages/silverstripe/template-engine). See [abstraction of template rendering](#view-layer-abstraction) for more details.
+- [`silverstripe/htmleditor-tinymce`](https://packagist.org/packages/silverstripe/htmleditor-tinymce). See [TinyMCE is in a separate module now](#tinymce) for more details.
```
emteknetnz: This should be the first point since it's what people need to pay the most attention to

en/08_Changelogs/6.0.0.md
```diff
@@ -1139,6 +1152,50 @@ SilverStripe\ORM\Connect\MySQLDatabase:
...
+
+Instead, we have moved all TinyMCE functionality used by Silverstripe CMS into a new module - `silverstripe/htmleditor-tinymce`. To continue using TinyMCE in your project, you must add that dependency in your `composer.json` file.
+
+Part of this change includes further abstracting the [`HTMLEditorConfig](api:SilverStripe\Forms\HTMLEditor\HTMLEditorConfig) class and related functionality, which makes it easier to replace TinyMCE with an alternative HTML editor implementation. We intend to replace TinyMCE as the default HTML editor for new projects with an alternative editor in a future minor release of Silverstripe CMS. Existing projects will be able to opt-in to the new editor when it is available, and all projects including new projects will be able to continue using TinyMCE for the lifetime of CMS 6.
+
+Part of this change includes a new way to define rules such as which elements and attributes are allowed in HTML content. You can continue to set the `valid_elements` and `extended_valid_elements` configuration if you want to, but the new API is generic and will work for any `HTMLEditorConfig` implementation and can be defined in YAML configuration.
```
```suggestion
Silverstripe CMS 6 continues to use TinyMCE 6, the same version used in CMS 5. We chose not to upgrade to TinyMCE 7 because that version of TinyMCE has `GPLv2+` license which is incompatible with our `BSD-3` license.

We plan to introduce a new default HTML editor in a future release of Silverstripe. Existing projects can opt-in when it becomes available, while both new and existing projects can continue using TinyMCE for the lifetime of CMS 6.

To facilitate swapping to a new editor, we have abstracted the HTMLEditorConfig class and related functionality. This change makes it possible to replace TinyMCE with a different HTML editor in the future.

Because TinyMCE won't be the only available HTML editor, we've created a new module `silverstripe/htmleditor-tinymce` and moved all TinyMCE functionaly out of core and in to there. To continue using TinyMCE in your project, you must add that dependency in your `composer.json` file.

Part of this change includes a new way to define rules such as which elements and attributes are allowed in HTML content. You can continue to set the `valid_elements` and `extended_valid_elements` configuration if you want to. The new API is generic and will work for any `HTMLEditorConfig` implementation and can be defined in YAML configuration.
```

emteknetnz (changes requested):

en/08_Changelogs/6.0.0.md
ChloeSartorelli: I'd prefer to give people a jump start if they want to do it. I don't think there's mixed messages - I've made it very clear this is untested and unsupported, but that at least these steps will be required if someone wants to attempt it.

en/08_Changelogs/6.0.0.md
ChloeSartorelli: Which "this" is "this"? Do you mean the third item in the list should be moved to the top? GitHub isn't good at telling saying whether you were intending to highlight a specific line sometimes 😅

emteknetnz: The `silverstripe/htmleditor-tinymce` should be the first bullet point as it's the one that requires the most action. People will pay more attention to things at the top of a list

en/08_Changelogs/6.0.0.md
emteknetnz: Yeah ... this new license really does complicate things and people may not realise what that entails until after they've upgraded. I don't think this is something we should really be encouraging.  Also we do intend to replace the editor which would be what we people should gravitate towards instead.

en/08_Changelogs/6.0.0.md
ChloeSartorelli: Done. They're in alpha order now which is also sensible.

en/08_Changelogs/6.0.0.md
ChloeSartorelli: Please suggest changes to each paragraph individually in future - I've asked for this before.

It's too difficult with GitHub's diffs to see what specific changes you're actually suggesting here, so I have to go line-by-line and check for myself.

It's also difficult to comment on and discuss specific changes when they're all together like this.

ChloeSartorelli: I've made some of these changes, altered some, and left some the same. If there are still changes you want please comment on each paragraph independently.

en/08_Changelogs/6.0.0.md
ChloeSartorelli: Is there another place you'd prefer I put this information for people who may want to upgrade to TinyMCE 7?
Or is there a way we can reduce the risk you perceive in providing it here? Additional warnings, hide it in a collapsible element, etc?

ChloeSartorelli: For now I've added a note into the warning about seeking legal advice before using `GPL` licensed code.

emteknetnz: Ha "seek legal advise" is a bit much

> Is there another place you'd prefer I put this information

I can't really think of anywhere official. I just don't think we should be including this information. Even with warnings, it seems odd documenting how to do something we don't stand behind and could simply break at any point.

ChloeSartorelli: > I just don't think we should be including this information.

Like I've said before, this is a question people are going to have. I don't want to be answering this question 20 different times for 20 different people - and I also don't want people floundering in the dark when we can at least give them a head start, because we do know that at least these two things will need to be done.

Your initial concern seemed to be that there would be confusion about it because we're saying we don't support it - so I added a really clear warning to remove that confusion.

Now your concern seems to be "it's not supported so don't say it" but I think it's sensible in a changelog (note specifically a changelog - I didn't put this in the main docs because that _would_ be weird) to say "here's some help for how to migrate away from this if you want to" when a change is made (or in this case when a change would normally be made, but wasn't).

> Ha "seek legal advise" is a bit much

I thought your concern was around confusion about the license, which is why I added that.
The language is based on the language TinyMCE use, to be fair. e.g https://github.com/tinymce/tinymce/issues/9453#issuecomment-2327646149
> The interpretation of this can vary based on your individual use case, so we recommend consulting with a legal expert to ensure compliance.

and https://www.tiny.cloud/blog/tinymce-7-release-highlights/
> You should consult your legal team for compliance assurance around all your software usage.

en/08_Changelogs/6.0.0.md
```diff
@@ -1217,6 +1230,54 @@ SilverStripe\Control\Session:
...
+```php
+// app/_config.php
+use SilverStripe\Core\Manifest\ModuleLoader;
+use SilverStripe\TinyMCE\TinyMCEConfig;
+
+$moduleManifest = ModuleLoader::inst()->getManifest();
+$module = $moduleManifest->getModule('silverstripe/htmleditor-tinymce');
+$editorConfig = TinyMCEConfig::get('cms');
+$editorConfig->setOption('skin_url', $module->getResource('client/dist/tinymce/skins/ui/silverstripe')->getURL());
+```
```
```suggestion
If you want to use TinyMCE 7 in your own projects, and have done your due diligence to validate that its [new license](https://www.tiny.cloud/docs/tinymce/latest/license-key/) won't cause you any legal problems, you can theoretically do that.

Configure [`TinyMCEConfig`](api:SilverStripe\TinyMCE\TinyMCEConfig) class to use your own copy of TinyMCE by setting the [`base_dir`](api:api:SilverStripe\TinyMCE\TinyMCEConfig->base_dir) configuration property to the directory which holds your TinyMCE JavaScript distribution files e.g. `path/to/tinymce7/directory`. Then define the URL for the silverstripe skin if you want to use it:

```php
// app/_config.php
use SilverStripe\Core\Manifest\ModuleLoader;
use SilverStripe\TinyMCE\TinyMCEConfig;

$url = ModuleLoader::getModule('silverstripe/htmleditor-tinymce')->getResource('client/dist/tinymce/skins/ui/silverstripe')->getURL();
TinyMCEConfig::get('cms')->setOption('skin_url', $url);
```

Compromise here is probably keep the information mostly infact, just condense it down

emteknetnz (changes requested):

GuySartorelli:
Rebased

en/08_Changelogs/6.0.0.md
ChloeSartorelli: I've tried for something inbetween what I had and what you have.

- I want to explicitly keep the note that this isn't supported and may not work (because I don't want people asking us to make it work if it doesn't). I've moved it out of the callout block to compromise with your desire not to draw too much attention to this section.
- I removed the "e.g. `path/to/tinymce7/directory`" reference because that's not meaningful without the yaml example.
- I was going to keep that `$url` assignment to two lines but that's just not worth arguing about, the information is there even if it maybe wouldn't pass a peer review of a code PR.

I didn't realise `ModuleLoader::getModule()` existed, that's a nice find.

ChloeSartorelli: > I was going to keep that `$url` assignment to two lines but that's just not worth arguing about, the information is there even if it maybe wouldn't pass a peer review of a code PR.

lol nevermind, that doesn't pass linting. Made it two lines.
https://github.com/silverstripe/developer-docs/actions/runs/13847698077/job/38749338946?pr=691

GuySartorelli:
Lint failure will be fixed by https://github.com/silverstripe/documentation-lint/pull/19

emteknetnz (approved):

PR #692, opened 2025-02-13
DOC Document warning for allowed hosts

GuySartorelli:
Also some reworking on the related configuration to improve clarity, make sure everything is fully documented, and updates to the reverse proxy docs which was just wrong in some cases.

## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11610

### DIFF
--- a/en/02_Developer_Guides/09_Security/05_Secure_Coding.md
+++ b/en/02_Developer_Guides/09_Security/05_Secure_Coding.md
@@ -660,45 +660,106 @@ This is a recommended option to secure any controller which displays
 or submits sensitive user input, and is enabled by default in all CMS controllers,
 as well as the login form.
 
-## Request hostname forgery {#request-hostname-forgery}
+## Request hostname forgery and host header injection attacks {#request-hostname-forgery}
 
-To prevent a forged hostname appearing being used by the application, Silverstripe CMS
-allows the configure of a whitelist of hosts that are allowed to access the system. By defining
-this whitelist in your `.env` file, any request presenting a `Host` header that is
-*not* in this list will be blocked with a HTTP 400 error:
+Ideally your hosting will reject invalid host headers. For example Apache allows you to define valid hosts as part of the virtual host configuration, and a Web Application Firewall (WAF) can also be configured to validate the host header. However if your hosting is set to allow *any* host header, your project might be vulnerable to host header injection attacks.
+
+To prevent a forged hostname being used by your application, Silverstripe CMS
+allows the configuration of an allow list of valid hosts that are allowed in the `Host` header. This is an extra layer of protection against this type of attack.
+By defining this allow list, any request presenting a `Host` header that is *not* in this list will be blocked with a HTTP 400 error.
+
+While you should have appropriate validation at a hosting level, it is best practice to also configure this in your project.
+
+> [!NOTE]
+> If this configuration is not set, a warning will be logged.
+> If you have implemented adequate protections in your hosting already and have legitimate reasons to not set this configuration,
+> you can explicitly set it to `"*"`. That will disable logging and declare that your project should explicitly allow any host
+> header without validating it.
+
+The main way to configure this is using the `SS_ALLOWED_HOSTS` environment variable:
 
 ```bash
-SS_ALLOWED_HOSTS="www.example.com,example.com,subdomain.example.com"
+SS_ALLOWED_HOSTS="example.com,www.example.com,subdomain.example.com"
 ```
 
-Please note that if this configuration is defined, you *must* include *all* subdomains (eg `<www>.`)
-that will be accessing the site.
+You can also configure the allow list with YAML configuration, which can be useful if you don't have full control over your hosting:
+
+```yml
+---
+after: requestprocessors
+---
+SilverStripe\Core\Injector\Injector:
+  SilverStripe\Control\Middleware\AllowedHostsMiddleware:
+    properties:
+      AllowedHosts:
+        - 'example.com'
+        - 'www.example.com'
+        - 'subdomain.example.com'
+```
+
+> [!WARNING]
+> Please note that you *must* include *all* subdomains (eg `www.`) that will be serving out content from the project using the same codebase.
+> For example if your CMS serves content on the following hosts, *all* of them must be added to the allow list:
+>
+> - `example.com` (primary domain)
+> - `www.example.com` (www subdomain)
+> - `blog.example.com` (subdomain for a blog)
+> - `example.org` (secondary domain serving the same content)
+
+Note that domains which only *redirect* to your project (i.e. with a `301` HTTP response) should not be added as allowed hosts.
+
+### Reverse proxies and forwarded hosts
 
 When Silverstripe CMS is run behind a reverse proxy, it's normally necessary for this proxy to
 use the `X-Forwarded-Host` request header to tell the webserver which hostname was originally
-requested. However, when Silverstripe CMS is not run behind a proxy, this header can still be
+requested. The [`TrustedProxyMiddleware`](api:SilverStripe\Control\Middleware\TrustedProxyMiddleware) ensures only trusted IP addresses are allowed to use
+that header to override the host header in Silverstripe CMS.
+
+Without that middleware, the `X-Forwarded-Host` header could be
 used by attackers to fool the server into mistaking its own identity.
 
 The risk of this kind of attack causing damage is especially high on sites which utilise caching
 mechanisms, as rewritten urls could persist between requests in order to misdirect other users
 into visiting external sites.
 
-In order to prevent this kind of attack, it's necessary to whitelist trusted proxy
-server IPs using the SS_TRUSTED_PROXY_IPS define in your `.env`.
+In order to prevent this kind of attack, if your project is behind a reverse proxy, it's necessary to define the trusted proxy
+server IP addresses in an allow list.
+
+The primary way to define this allow list is using the `SS_TRUSTED_PROXY_IPS` environment variable:
 
 ```bash
 SS_TRUSTED_PROXY_IPS="127.0.0.1,192.168.0.1"
 ```
 
-You can also whitelist subnets in CIDR notation if you don't know the exact IP of a trusted proxy.
-For example, some cloud provider load balancers don't have fixed IPs.
+You can also allow subnets in CIDR notation if you don't know the exact IP of a trusted proxy.
+For example, some cloud provider load balancers don't have fixed IP addresses.
 
 ```bash
 SS_TRUSTED_PROXY_IPS="10.10.0.0/24,10.10.1.0/24,10.10.2.0/24"
 ```
 
+You can also define the allow list using YAML configuration:
+
+```yml
+---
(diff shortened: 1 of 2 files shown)


en/02_Developer_Guides/09_Security/05_Secure_Coding.md
```diff
@@ -660,80 +660,114 @@ This is a recommended option to secure any controller which displays
...
 ```yml
-SilverStripe\Control\TrustedProxyMiddleware:
-  properties:
-    ProxyHostHeaders: X-Forwarded-Host
-    ProxySchemeHeaders: X-Forwarded-Protocol
-    ProxyIPHeaders: X-Forwarded-Ip
-```
-
-```bash
-SS_TRUSTED_PROXY_HOST_HEADER="HTTP_X_FORWARDED_HOST"
```
emteknetnz, line 712: The docs for the TRUSTED_PROXY env variables are being deleted but not added in somewhere else - is that intentional?

en/08_Changelogs/5.4.0.md
```diff
@@ -32,6 +33,20 @@ We have provided a severity rating of the vulnerabilities below based on the CVS
...
+
+Notably this is used in [`Director::host()`](api:SilverStripe\Control\Director::host()), which in turn is called by many methods including [`Director::hostName()`](api:SilverStripe\Control\Director::hostName()), [`Director::protocolAndHost()`](api:SilverStripe\Control\Director::protocolAndHost()), [`Director::is_site_url()`](api:SilverStripe\Control\Director::is_site_url()), and [`Director::absoluteURL()`](api:SilverStripe\Control\Director::absoluteURL()).
+
+Ideally your hosting will reject invalid host headers. For example Apache allows you to define valid hosts as part of the virtual host configuration, and a Web Application Firewall (WAF) can also be configured to validate the host header. However if your hosting is set to allow *any* host header, your project might be vulnerable to host header injection attacks.
+
+You can configure Silverstripe CMS to validate the host header, which is an extra layer of protection against this type of attack. While you should have appropriate validation at a hosting level, it is best practice to also configure this in your project.
```
emteknetnz:
```suggestion
You should configure Silverstripe CMS to validate the host header, which is an extra layer of protection against this type of attack. While you should have appropriate validation at a hosting level, it is best practice to also configure this in your project.
```

en/08_Changelogs/5.4.0.md
```diff
@@ -32,6 +33,20 @@ We have provided a severity rating of the vulnerabilities below based on the CVS
...
+
+This configuration has existed since 2016, but we've been alerted that many projects still have not configured their hosting nor their project to adequately validate host headers. To help prompt developers, we've added a warning which will be logged if the configuration is not set.
```
emteknetnz, line 51: > To help prompt developers, we've added a warning which will be logged if the configuration is not set.

This bit needs a lot more prominence, since it is the actual change.  If should basically be in its own paragraph and say "If you haven't configured SS_ALLOWED_HOSTS, or set some of the new YML (which you should either add an example of in this change log, or link direclty to an example in docs)", then yours logs will now fill up with warnings.

emteknetnz (changes requested):

en/02_Developer_Guides/09_Security/05_Secure_Coding.md
ChloeSartorelli: The code for those was removed when the middleware was added in https://github.com/silverstripe/silverstripe-framework/commit/ccc86306b6a0d6eb9bbb8b41b209b8595393d591

en/08_Changelogs/5.4.0.md
ChloeSartorelli: Done

en/08_Changelogs/5.4.0.md
ChloeSartorelli: It is already it's own paragraph? I've added _another_ paragraph afterward with wording similar to the language you've suggested.

> or  set some of the new YML (which you should either add an example of in this change log, or link direclty to an example in docs)

The YAML isn't new. There is a link to the example in the docs - that's the last line in this section. We usually only add code examples for _new_ API, but this API isn't new by any means.

en/08_Changelogs/5.4.0.md
```diff
@@ -32,6 +33,22 @@ We have provided a severity rating of the vulnerabilities below based on the CVS
...
+Ideally your hosting will reject invalid host headers. For example Apache allows you to define valid hosts as part of the virtual host configuration, and a Web Application Firewall (WAF) can also be configured to validate the host header. However if your hosting is set to allow *any* host header, your project might be vulnerable to host header injection attacks.
+
+You should configure Silverstripe CMS to validate the host header, which is an extra layer of protection against this type of attack. While you should have appropriate validation at a hosting level, it is best practice to also configure this in your project.
+
+This configuration has existed since 2016, but we've been alerted that many projects still have not configured their hosting nor their project to adequately validate host headers. To help prompt developers, we've added a warning which will be logged if the configuration is not set.
+
+If you haven't configured your allowed hosts, then the new warning will be logged once for each request.
+
+You can learn more about the relevant configuration in the [secure coding](/developer_guides/security/secure_coding/#request-hostname-forgery) documentation.
+
```
emteknetnz:
```suggestion
### Logged warning if allowed hosts have not been configured {#allowed-hosts-warning}

If your site does not have one of the following configured, then a warning will now be logged on every request:

- `SS_ALLOWED_HOSTS` environment variable
- [`AllowedHostsMiddleware.AllowedHosts`](api:SilverStripe\Control\Middleware\AllowedHostsMiddleware->AllowedHosts) property
- [`Director.alternate_base_url`](api:SilverStripe\Control\Director->alternate_base_url) property

If one of those is not set, then the "host" HTTP request header is used by Silverstripe CMS to determine what the host name is for your project. This is used when creating absolute URLs, e.g. for use in emails that link back to the site.

Ideally your hosting will reject invalid host headers. For example Apache allows you to define valid hosts as part of the virtual host configuration, and a Web Application Firewall (WAF) can also be configured to validate the host header. However if your hosting is set to allow *any* host header, your project might be vulnerable to host header injection attacks.

You should configure Silverstripe CMS to validate the host header, which is an extra layer of protection against this type of attack. While you should have appropriate validation at a hosting level, it is best practice to also configure this in your project.

This configuration has existed since 2016, but we've been alerted that many projects still have not configured their hosting nor their project to adequately validate host headers. To help prompt developers, we've added a warning which will be logged if the configuration is not set.

You can learn more about the relevant configuration in the [secure coding](/developer_guides/security/secure_coding/#request-hostname-forgery) documentation.

```

emteknetnz: The most important info should be up the top, in this case it's saying that if you are missing some configuration then you'll get logged warnings

emteknetnz (changes requested):

en/08_Changelogs/5.4.0.md
ChloeSartorelli: > ``[`Director.alternate_base_url`](api:SilverStripe\Control\Director->alternate_base_url) property``

Not relevant here. The warning is still logged if you have that set. All that changes is whether the host header is used by `Director::host()` or not.

> If one of those is not set, then the "host" HTTP request header is used by Silverstripe CMS to determine what the host name is for your project.

Not quite. The host header will still be used if you have allowed hosts set too - that's kinda the whole point.

I've moved that one bit to the top - the rest I've left the same as it already is. If you want to change a specific paragraph, please suggest changes to each paragraph individually so it's easy to comment and respond to. Changing all of it in a single diff is really hard to respond to - let alone see what specific changes you're suggesting.

emteknetnz: >  Not relevant here. The warning is still logged if you have that set.

OK, so if that makes it so that host header is not used by Director::host(), then won't that mean that host header injection isn't possible because the host headers aren't used? i.e. links in an email back to the site will just use alternate_base_url ... or in that instance do they still use the host header?

ChloeSartorelli: In theory they will use the alternate base URL instead of the host header _when using `Director::host()`_ - which is an important caveat.

If the host header is accessed anywhere else, an injection attack could still be possible, so I think we should continue to warn about not having allowed hosts config set even if `alternate_base_url` is set.

emteknetnz: I honestly think there's too much nuance in Director.alternate_base_url / Director::host() and it does more harm then good mentioning them. It's very distracting, particularly since they're at the top.

The key thing that people need to know is that you host header injection may be possible is some scenarios, therefore you need to set either an env variable or some yml configure in your application to limit allowed hosts

en/08_Changelogs/5.4.0.md
```diff
@@ -32,6 +33,25 @@ We have provided a severity rating of the vulnerabilities below based on the CVS
...
+### Logged warning if allowed hosts have not been configured {#allowed-hosts-warning}
+
+If your site does not have one of the following configured, then a warning will now be logged on every request:
+
+- `SS_ALLOWED_HOSTS` environment variable
+- [`AllowedHostsMiddleware.AllowedHosts`](api:SilverStripe\Control\Middleware\AllowedHostsMiddleware->AllowedHosts) property
+
+If [`Director.alternate_base_url`](api:SilverStripe\Control\Director->alternate_base_url) is not set, the "host" header is used by Silverstripe CMS to determine what the host name is for your project. This is useful when creating absolute URLs, e.g. for use in emails.
+
+Notably this is used in [`Director::host()`](api:SilverStripe\Control\Director::host()), which in turn is called by many methods including [`Director::hostName()`](api:SilverStripe\Control\Director::hostName()), [`Director::protocolAndHost()`](api:SilverStripe\Control\Director::protocolAndHost()), [`Director::is_site_url()`](api:SilverStripe\Control\Director::is_site_url()), and [`Director::absoluteURL()`](api:SilverStripe\Control\Director::absoluteURL()).
```
emteknetnz:
```suggestion
If [`Director.alternate_base_url`](api:SilverStripe\Control\Director->alternate_base_url) is set then it is often used instead instead of the "host" HTTP request header for determining the URL of the website. However there may be some scenarios where the host header is still used, so having `alternate_base_url` configured by itself will not prevent a warning be logged if allowed hosts are not configured.
```

Maybe this?

emteknetnz (changes requested):

en/08_Changelogs/5.4.0.md
ChloeSartorelli: I've removed the note about `alternate_base_url`

en/08_Changelogs/5.4.0.md
ChloeSartorelli: Based on https://github.com/silverstripe/developer-docs/pull/692#discussion_r1955319621 it sounds like referencing `alternate_base_url` was confusing, so I've removed that and left the rest.

I.e. it doesn't matter that sometimes some methods won't use the host header if `alternate_base_url` is set - but it is useful to know some of the common methods that _do usually_ use the host header.

emteknetnz (approved):

Issue #695, opened 2025-02-16
Improve docs about security best practices


GuySartorelli:
I'm writing a blog post about security best practices and have noticed a couple of things in our documentation around this topic are a bit out of date.

## PRs
- https://github.com/silverstripe/developer-docs/pull/696

GuySartorelli:
PR got merged

PR #696, opened 2025-02-16
DOC Update secure coding documentation to reflect best practices

GuySartorelli:
Two commits here - if you squash, make sure the commit message is broad enough to encompass both of them.

Commit 1: Update best practice for setting `CanonicalURLMiddleware.ForceSSL`

Commit 2: Backport of https://github.com/silverstripe/developer-docs/pull/692 minus the bit about the warning. The beta has been delayed and it's possible the blog post will be live before we swap over the source of documentation.

## Issue
- https://github.com/silverstripe/developer-docs/issues/695

### DIFF
--- a/en/02_Developer_Guides/09_Security/05_Secure_Coding.md
+++ b/en/02_Developer_Guides/09_Security/05_Secure_Coding.md
@@ -660,45 +660,100 @@ This is a recommended option to secure any controller which displays
 or submits sensitive user input, and is enabled by default in all CMS controllers,
 as well as the login form.
 
-## Request hostname forgery {#request-hostname-forgery}
+## Request hostname forgery and host header injection attacks {#request-hostname-forgery}
 
-To prevent a forged hostname appearing being used by the application, Silverstripe CMS
-allows the configure of a whitelist of hosts that are allowed to access the system. By defining
-this whitelist in your `.env` file, any request presenting a `Host` header that is
-*not* in this list will be blocked with a HTTP 400 error:
+Ideally your hosting will reject invalid host headers. For example Apache allows you to define valid hosts as part of the virtual host configuration, and a Web Application Firewall (WAF) can also be configured to validate the host header. However if your hosting is set to allow *any* host header, your project might be vulnerable to host header injection attacks.
+
+To prevent a forged hostname being used by your application, Silverstripe CMS
+allows the configuration of an allow list of valid hosts that are allowed in the `Host` header. This is an extra layer of protection against this type of attack.
+By defining this allow list, any request presenting a `Host` header that is *not* in this list will be blocked with a HTTP 400 error.
+
+While you should have appropriate validation at a hosting level, it is best practice to also configure this in your project.
+
+The main way to configure this is using the `SS_ALLOWED_HOSTS` environment variable:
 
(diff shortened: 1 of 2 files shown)


en/02_Developer_Guides/09_Security/05_Secure_Coding.md
```diff
@@ -660,115 +660,159 @@ This is a recommended option to secure any controller which displays
...
+
+In some circumstances, like connecting to a database on a remote host for example, you may wish to enable SSL encryption to ensure the protection of sensitive information and database access credentials.
+You can configure that by setting the following environment variables:
+
+| Name  | Description |
+| ----  | ----------- |
+| `SS_DATABASE_SSL_KEY` | Absolute path to SSL key file (optional - but if set, `SS_DATABASE_SSL_CERT` must also be set) |
+| `SS_DATABASE_SSL_CERT` | Absolute path to SSL certificate file (optional - but if set, `SS_DATABASE_SSL_KEY` must also be set) |
+| `SS_DATABASE_SSL_CA` | Absolute path to SSL Certificate Authority bundle file (optional) |
+| `SS_DATABASE_SSL_CIPHER` | Custom SSL cipher for database connections (optional) |
```
ChloeSartorelli, line 813: Makes sense to move this to be adjacent to the other bit of docs talking about TLS.

en/02_Developer_Guides/09_Security/05_Secure_Coding.md
```diff
@@ -660,115 +660,159 @@ This is a recommended option to secure any controller which displays
...
+
+## Secure sessions and cookies
```
ChloeSartorelli, line 815: Makes sense to separate this into its own section - secure cookies and TLS are sort-of related but I don't think enough so to join them together like that.

en/02_Developer_Guides/09_Security/05_Secure_Coding.md
```diff
@@ -660,115 +660,159 @@ This is a recommended option to secure any controller which displays
...
-Silverstripe CMS recommends the use of TLS (HTTPS) for your application, and you can easily force the use through the
-director function `forceSSL()`
+## TLS (aka SSL aka HTTPS)
 
-```php
-use SilverStripe\Control\Director;
+Silverstripe CMS recommends the use of TLS (HTTPS) for your application. You can configure this by setting the `ForceSSL` property on the [`CanonicalURLMiddleware`](api:SilverStripe\Control\Middleware\CanonicalURLMiddleware) singleton.
 
-if (!Director::isDev()) {
-    Director::forceSSL();
```
ChloeSartorelli, line 747: This API does exist but it's a lot less powerful than configuring via YAML because
- there are a lot of options on `CanonicalURLMiddleware` that can't be set via director (even the rest of this old docs has to instantiate a singleton of the middleware to set something)
- you ideally want everything configuring this in the same place - so if you have to set some of it in YAML and then some here, that's hard to notice and may cause problems
- I'd expect basically every project that configures `CanonicalURLMiddleware` uses YAML, following CWP's example.

Basically there's no good reason to be setting this in `_config.php`

emteknetnz (approved):

PR #733, opened 2025-04-29
DOC Document new non-blocking sessions solution

GuySartorelli:
## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11671

### DIFF
--- a/en/02_Developer_Guides/18_Cookies_And_Sessions/02_Sessions.md
+++ b/en/02_Developer_Guides/18_Cookies_And_Sessions/02_Sessions.md
@@ -100,9 +100,52 @@ including form and page comment information. None of this is vital but `clear_al
 $session->clearAll();
 ```
 
-## Cookies
+## Configuration
 
-### Samesite attribute
+### Session lifetime
+
+By default, the client-side session cookie will only last until the user closes their browser, and the session on the server-side will live for the number of seconds set in the [`session.gc_maxlifetime`](https://www.php.net/manual/en/session.configuration.php#ini.session.gc-maxlifetime) ini configuration since it was last touched.
+
+You can configure the session lifetime by setting the [`Session.timeout`](SilverStripe\Control\Session->timeout) configuration property to the number of seconds the session should be valid for since the last modified time. If one of the session handlers included with `silverstripe/framework` is used, this will be used for both setting the "expires" attribute on the client-side cookie, as well as for determining when the server-side record for storing the session is deleted.
+
+```yml
+SilverStripe\Control\Session:
+  timeout: 86400
+```
+
+> [!WARNING]
+> If you use a save handler which isn't included in `silverstripe/framework`, it may treat the session lifetime differently. You should check its documentation or implementation to understand how it works.
+
+### Save handler
+
+You can choose how session are handled by setting the [`Session.save_handler`](api:SilverStripe\Control\Session->save_handler) configuration property. By default this is set to [`FileSessionHandler`](api:SilverStripe\Control\SessionHandler\FileSessionHandler).
+
+If you want to use the default PHP file session handler instead, you can set the `Session.save_handler` configuration to `null`.
+
+```yml
+SilverStripe\Control\Session:
+  save_handler: null
+```
+
+You can also set it to the FQCN or injector service name of any session handler that implements [`SessionHandlerInterface`](https://www.php.net/manual/en/class.sessionhandlerinterface.php).
+
+#### `FileSessionHandler`
+
+The default file-based session handler for PHP holds a lock on the session file while the session is open. This means that multiple concurrent requests from the same user have to wait for one another to finish processing after a session has been started. This includes AJAX requests.
+
+To resolve this problem, Silverstripe CMS comes with [`FileSessionHandler`](api:SilverStripe\Control\SessionHandler\FileSessionHandler).
+
+Note that in edge case scenarios, for example if your application wants to modify a session value *based on the value that is already set* and must do so for each request, non-blocking sessions may cause unexpected results.
+
+`FileSessionHandler` differs from the default PHP file session handler in the following ways:
+
+1. It doesn't lock the session file, and therefore doesn't block concurrent requests.
+1. The [`Session.timeout`](api:SilverStripe\Control\Session->timeout) configuration property is used as the source of truth for the lifetime of session files (see [session lifetime](#session-lifetime) above).
+1. If there are problems reading or writing to session files, the [default logging service](/developer_guides/debugging/error_handling/) is used to log them.
+
+### Cookies
+
+#### Samesite attribute
 
 The session cookie is handled slightly differently than most cookies on the site, which provides the opportunity to handle the samesite attribute separately from other cookies. By default, it is set to `Strict` to prevent cross-site attacks.
 
@@ -122,7 +165,7 @@ SilverStripe\Core\Injector\Injector:
       TokenCookieSameSite: 'Lax'
 ```
 
-### Secure session cookie
+#### Secure session cookie
 
 A different `session_name` cookie is used when using the `https` protocol for security purposes. This uses the session_name `SECSESSID` for `https` connections instead of the default `PHPSESSID`. Doing so adds an extra layer of security to your session cookie since you no longer share `http` and `https` sessions.
 
@@ -144,7 +187,7 @@ SilverStripe\Core\Injector\Injector:
 
 Note that if you set `cookie_samesite` to `None` (which is *strongly* discouraged), the `cookie_secure` value will *always* be `true`.
 
-## Relaxing checks around user agent strings
+### Relaxing checks around user agent strings
 
 Out of the box, Silverstripe CMS will invalidate a user's session if the `User-Agent` header changes. This provides some supplemental protection against session high-jacking attacks.
 
--- a/en/08_Changelogs/6.1.0.md
+++ b/en/08_Changelogs/6.1.0.md
@@ -1,0 +1,51 @@
+---
+title: 6.0.0 (unreleased)
+---
+
+# 6.0.0 (unreleased)
+
+## Overview
+
+- [Features and enhancements](#features-and-enhancements)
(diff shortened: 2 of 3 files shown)


en/08_Changelogs/6.1.0.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Other new features
+
+## API changes
+
+### Deprecated API
+
+The following API has been deprecated in this release, and will be removed or replaced in a future major release.
+
+It's best practice to avoid usnig deprecated code where possible, but sometimes it's unavoidable. This API will all continue to be available at least until the next major release.
```
ChloeSartorelli: Blurb matches https://github.com/silverstripe/developer-docs/pull/734

en/02_Developer_Guides/18_Cookies_And_Sessions/02_Sessions.md
```diff
@@ -100,9 +100,52 @@ including form and page comment information. None of this is vital but `clear_al
...
 
-## Cookies
+## Configuration
 
-### Samesite attribute
+### Session lifetime
+
+By default, the session cookie will only last until the user closes their browser, and the session on the server-side will live for the number of seconds set in the [`session.gc_maxlifetime`](https://www.php.net/manual/en/session.configuration.php#ini.session.gc-maxlifetime) ini configuration since it was last touched.
+
+You can configure the session lifetime by setting the [`Session.timeout`](SilverStripe\Control\Session->timeout) configuration property to the number of seconds the session should be valid for since the last modified time.
```
emteknetnz:
```suggestion
By default, the client-side session cookie will only last until the user closes their browser, and the session on the server-side will live for the number of seconds set in the [`session.gc_maxlifetime`](https://www.php.net/manual/en/session.configuration.php#ini.session.gc-maxlifetime) ini configuration since it was last touched.

You can configure the session lifetime by setting the [`Session.timeout`](SilverStripe\Control\Session->timeout) configuration property to the number of seconds the session should be valid for since the last modified time. This will be used for both setting the "expires" attribute on the client-side cookie, as well as for determining when the server-side record for storing the session is deleted.
```

emteknetnz: I'm assuming what I've written above is correct, though double check what I'm saying is correct

en/02_Developer_Guides/18_Cookies_And_Sessions/02_Sessions.md
```diff
@@ -100,9 +100,52 @@ including form and page comment information. None of this is vital but `clear_al
...
+
+```yml
+SilverStripe\Control\Session:
+  timeout: 86400
+```
+
+> [!WARNING]
+> If you use a save handler which isn't included in `silverstripe/framework`, it may treat the session lifetime differently. You should check its docuemntation or implementation to understand how it works.
```
emteknetnz:
```suggestion
> If you use a save handler which isn't included in `silverstripe/framework`, it may treat the session lifetime differently. You should check its documentation or implementation to understand how it works.
```

en/02_Developer_Guides/18_Cookies_And_Sessions/02_Sessions.md
```diff
@@ -100,9 +100,52 @@ including form and page comment information. None of this is vital but `clear_al
...
+```yml
+SilverStripe\Control\Session:
+  save_handler: null
+```
+
+You can also set it to the FQCN or injector service name of any session handler that implements [`SessionHandlerInterface`](https://www.php.net/manual/en/class.sessionhandlerinterface.php).
+
+#### `FileSessionHandler`
+
+The default file-based session handler for PHP holds a lock on the session file while the session is open. In practice this means that multiple concurrent requests from the same user have to wait for one another to finish processing after a session has been started. This includes AJAX requests.
```
emteknetnz:
```suggestion
The default file-based session handler for PHP holds a lock on the session file while the session is open. In practice this means that multiple concurrent requests from the same user have to wait for one another to finish processing after a session has been started. This includes XHR requests.
```

emteknetnz: AJAX technically only refers to XHR requests that return HTML, wheras we also care about requests that return JSON

en/02_Developer_Guides/18_Cookies_And_Sessions/02_Sessions.md
```diff
@@ -100,9 +100,52 @@ including form and page comment information. None of this is vital but `clear_al
...
+
+To resolve this problem, Silverstripe CMS comes with [`FileSessionHandler`](api:SilverStripe\Control\SessionHandler\FileSessionHandler).
+
+Note that in edge case scenarios, for example if your application wants to modify a session value *based on the value that is already set* and must do so for each request, non-blocking sessions may cause unexpected results.
+
+`FileSessionHandler` differs from the default PHP file session handler in the following ways:
+
+1. It doesn't lock the session file, and therefore doesn't block concurrent requests.
```
emteknetnz, line 142:
```suggestion
1. It doesn't keep a lock on the session file, and therefore doesn't block concurrent requests.
```
I think this is correct? It does lock the file, it's just it unlocks it again when it's finished, which is practive happens very quickly

en/08_Changelogs/6.1.0.md
```diff
@@ -0,0 +1,51 @@
...
+  - [Non-blocking file-based sessions](#non-blocking-sessions)
+  - [Other new features](#other-new-features)
+- [API changes](#api-changes)
+- [Bug fixes](#bug-fixes)
+
+## Features and enhancements
+
+## Non-blocking file-based sessions {#non-blocking-sessions}
+
+The default file-based session handler for PHP holds a lock on the session file while the session is open. In practice this means that multiple concurrent requests from the same user have to wait for one another to finish processing after a session has been started. This includes AJAX requests.
```
```suggestion
The default file-based session handler for PHP holds a lock on the session file while the session is open. This means that multiple concurrent requests from the same user have to wait for one another to finish processing after a session has been started. This includes XHR requests.
```

en/08_Changelogs/6.1.0.md
```diff
@@ -0,0 +1,51 @@
...
+
+To resolve this problem, a new [`Session.save_handler`](api:SilverStripe\Control\Session->save_handler) configuration property has been added. By default this is set to [`FileSessionHandler`](api:SilverStripe\Control\SessionHandler\FileSessionHandler).
```
```suggestion
To resolve this problem, a new [`Session.save_handler`](api:SilverStripe\Control\Session->save_handler) configuration property has been added. By default this is set to [`FileSessionHandler`](api:SilverStripe\Control\SessionHandler\FileSessionHandler) which is *non-blocking* which means that multiple concurrent requests from the same user do *not* have to wait for one another to finish.
```

en/08_Changelogs/6.1.0.md
```diff
@@ -0,0 +1,51 @@
...
+
+Note that in edge case scenarios, for example if your application wants to modify a session value *based on the value that is already set* and must do so for each request, non-blocking sessions may cause unexpected results.
+
+If you want to use the default PHP file session handler instead, you can set the `Session.save_handler` configuration to `null`.
```
```suggestion
If you want to use the default blocking PHP file session handler instead, you can set the `Session.save_handler` configuration to `null`.
```

emteknetnz (changes requested):

en/02_Developer_Guides/18_Cookies_And_Sessions/02_Sessions.md
ChloeSartorelli: Done

en/02_Developer_Guides/18_Cookies_And_Sessions/02_Sessions.md
ChloeSartorelli: Done

en/02_Developer_Guides/18_Cookies_And_Sessions/02_Sessions.md
ChloeSartorelli: > AJAX technically only refers to XHR requests that return HTML

I don't think that's true? Do you have some documentation for that? I know the term does stand for "Asynchronous JavaScript And XML" which contains the term "XML", but its usage everywhere I can find has it just meaning essentially "an HTTP request fired off from client-side JavaScript".

It's worth noting that "XHR" is short for `XMLHttpRequest` (which also contains the term "XML") which is a specific JavaScript API, and we use `fetch` a lot rather than that specific API.

But ultimately both "XHR" and "AJAX" as terminology really mean the same thing and don't specify what is returned, to the best of my knowledge. I think AJAX is the more commonly used term so I would prefer to keep it here.

en/02_Developer_Guides/18_Cookies_And_Sessions/02_Sessions.md
ChloeSartorelli: It doesn't lock the file, or if it does that's an implementation detail of PHP or of the filesystem and not of the session handler.
Also if PHP or the filesystem does lock the file, it's done for such a small period of time that I don't think that technical distinction is helpful.

en/08_Changelogs/6.1.0.md
ChloeSartorelli: XHR vs AJAX discussion can be held in https://github.com/silverstripe/developer-docs/pull/733#discussion_r2074381494

ChloeSartorelli: Made the other change.

en/08_Changelogs/6.1.0.md
ChloeSartorelli: Done but without the emphasis around "not"

en/02_Developer_Guides/18_Cookies_And_Sessions/02_Sessions.md
emteknetnz: Opps, apologies got ajax confused with pjax.  Yes ajax is the superset of xhr so is correct.

emteknetnz (approved):

PR #737, opened 2025-05-12
DOC Document filesystem performance improvements

GuySartorelli:
## Issue

- https://github.com/silverstripe/silverstripe-assets/issues/681

### DIFF
--- a/en/02_Developer_Guides/14_Files/04_File_Storage.md
+++ b/en/02_Developer_Guides/14_Files/04_File_Storage.md
@@ -43,7 +43,7 @@ These records are composite fields which contain sufficient information useful t
 to store, manage, and  publish files. By default this composite field behind this field stores the following details:
 
 | Field name     | Description |
-| ----------     | -----------
+| ----------     | ----------- |
 | `Hash`         | The sha1 of the file content, useful for versioning (if supported by the backend) |
 | `Filename`     | The internal identifier for this file, which may contain a directory path (not including assets). Multiple versions of the same file will have the same filename. |
 | `Variant`      | The variant for this file. If a file has multiple derived versions (such as resized files or reformatted documents) then you can point to one of the variants here. |
@@ -53,6 +53,15 @@ storage backend to determine how variants are managed.
 
 Note that the storage backend used will not be automatically synchronised with the database. Only files which
 are loaded into the backend through the asset API will be available for use within a site.
+
+## Flysystem adapters
+
+The [Flysystem](https://flysystem.thephpleague.com/docs/) backend includes several optional adapters for storing files in different ways. By default the local filesystem adapter is used, which stores files in your local filesystem in the path that the `ASSETS_PATH` constant resolves to (usually `public/assets/` relative to your project root).
+
+If you chose to use a different adapter for your project, note that there may be performance implications for doing so:
+
+- We use glob pattern matching where possible to look for image variants. If the adapter you've chosen can support glob pattern matching, consider subclassing it and implementing the [`GlobContentLister`](api:SilverStripe\Assets\Flysystem\GlobContentLister) interface. This dramatically improves performance. The alternative behaviour is to iterate through all files in a folder to check if they're a variant file or not.
+- Flysystem doesn't implement a way to check if a folder is empty. [`Filesystem::isEmpty()`](api:SilverStripe\Assets\Flysystem\Filesystem::isEmpty()) tries to get around this by only checking the first item in a generator returned by the adapter, but some adapters will still fetch data about many more files than is necessary e.g. from a networked service.
 
 ## Public file paths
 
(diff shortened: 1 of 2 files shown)


en/02_Developer_Guides/14_Files/04_File_Storage.md
```diff
@@ -43,7 +43,7 @@ These records are composite fields which contain sufficient information useful t
 to store, manage, and  publish files. By default this composite field behind this field stores the following details:
 
 | Field name     | Description |
-| ----------     | -----------
+| ----------     | ----------- |
```
ChloeSartorelli, line 46: Unrelated fix to table syntax. It's rendering fine without this for now but other markdown parsers could be more strict if we change our doc renderer in the future.

en/08_Changelogs/6.1.0.md
```diff
@@ -7,14 +7,31 @@ title: 6.0.0 (unreleased)
...
   - [Other new features](#other-new-features)
 - [API changes](#api-changes)
 - [Bug fixes](#bug-fixes)
 
 ## Features and enhancements
 
-## Non-blocking file-based sessions {#non-blocking-sessions}
+### Performance improvements
+
+Development of this release was focused largely on improving performance. Some improvements were safe to include in patch releases, so were included in patches to Silverstripe CMS 5.4 and merged up, while others required adding new API or had an increased risk of regressions so were only included in this release.
```
```suggestion
Development of this release was focused largely on improving performance.
```

It's confusing talking about prior releases. People will wonder why these weren't just mentioned in prior changelogs (obviously you can't since they're patch releases, though unless you're very used to semver releases like we are people can easily misunderstand this). I'm not sure if any of the 5.4 performance related improvements are even really worth talking about.

en/08_Changelogs/6.1.0.md
```diff
@@ -7,14 +7,31 @@ title: 6.0.0 (unreleased)
...
+
+#### Improved performance working with assets {#filesystem-performance}
+
+In projects with thousands of files in the same folder, especially with images, performance can be slow for some operations. That performance hit scales with the number of files in the folder you're looking at. One reason for this is that [Flysystem](https://flysystem.thephpleague.com/docs/) (the filesystem abstraction layer) can't rely on the service it's interfacing with having a way to discover if a folder is empty, or find files matching a certain pattern. This means that any time we needed to do either of those, we had to loop over data for every file in the folder we're interested in.
```
```suggestion
In projects with thousands of files in the same folder, especially with images with multiple variants, performance can be slow for some operations. That performance hit scales with the number of files in the folder. One reason for this is that [Flysystem](https://flysystem.thephpleague.com/docs/) (the filesystem abstraction layer) can't assume the filesystem it's interfacing with has a way to discover if a folder is empty, or find files in a folder matching a certain pattern. This means that any time either of those operations was performed on a folder, Flysystem had to loop over every file it contained.
```

en/08_Changelogs/6.1.0.md
```diff
@@ -7,14 +7,31 @@ title: 6.0.0 (unreleased)
...
+
+To remedy this, we've made the following changes:
+
+- Reduced the number of times we check if a folder is empty by removing duplicate checks.
+- Added [`Filesystem::isEmpty()`](api:SilverStripe\Assets\Flysystem\Filesystem::isEmpty()) to check if a folder is empty. With the default local filesystem adapter, this means we check exactly one file to know if the folder is empty or not. With other adapters it depends on the implementation (e.g. the AWS S3 adapter will likely still fetch a full page of up to 1000 files).
+- Added [`GlobContentLister`](api:SilverStripe\Assets\Flysystem\GlobContentLister) and [`GlobbableFileIDHelper`](api:SilverStripe\Assets\FilenameParsing\GlobbableFileIDHelper) interfaces to enable looking for image variants using a glob pattern. This dramatically improves performance with the default local filesystem adapter, but by default will have no effect for other adapters.
+    If you are using a different adapter that might support globbing such as using FTP (for specific FTP servers), consider implementing the `GlobContentLister` interface.
```
emteknetnz, line 31: Using FTP as a filesystem? For a Silverstripe install? That seems very strange, is there a different example you could give that's a little more standard?

emteknetnz (changes requested):
I'm part way reviewing this, will submit for now. Depending on what we do on the main PR may change this PR quite a bit

en/08_Changelogs/6.1.0.md
ChloeSartorelli: I think it's worth keeping but I don't have the energy to argue about it, so I'll remove it.

en/08_Changelogs/6.1.0.md
ChloeSartorelli: Changed "folder you're looking at" to "folder you're dealing with" because just ending the sentence at "files in the folder" makes my brain go "which folder, though?"

Made all the other changes.

en/08_Changelogs/6.1.0.md
ChloeSartorelli: The official adapters are:

- Local
- FTP
- InMemory
- Read-Only
- Path Prefixing
- AWS S3
- AsyncAws S3
- Azure Blob Storage
- Google Cloud Storage
- MongoDB GridFS
- SFTP (V2)
- SFTP (V3)
- WebDAV
- ZipArchive

If another of those stands out as a better example which can support glob pattern matching I'll be happy to swap it out. I think SFTP can handle glob pattern matching.

emteknetnz (approved):

PR #739, opened 2025-05-14
DOC Add note about monolog added context

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11699

### DIFF
--- a/en/06_Upgrading/07_Deprecations.md
+++ b/en/06_Upgrading/07_Deprecations.md
@@ -73,10 +73,13 @@ SilverStripe\Core\Injector\Injector:
 > ```bash
 > SS_ERROR_LOG="./silverstripe.log"
 > ```
->
-> You don't need any of the YAML configuration above if you are using the `SS_ERROR_LOG` environment variable - but you can use a combination of the environment variable and YAML configuration if you want to configure multiple error log files.
->
-> You will also need to make sure the user running the PHP process has write access to the log file, wherever you choose to put it.
+
+You don't need any of the YAML configuration above if you are using the `SS_ERROR_LOG` environment variable - but you can use a combination of the environment variable and YAML configuration if you want to configure multiple error log files.
+
+You will also need to make sure the user running the PHP process has write access to the log file, wherever you choose to put it.
+
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


en/06_Upgrading/07_Deprecations.md
```diff
@@ -73,10 +73,13 @@ SilverStripe\Core\Injector\Injector:
...
 > SS_ERROR_LOG="./silverstripe.log"
 > ```
->
-> You don't need any of the YAML configuration above if you are using the `SS_ERROR_LOG` environment variable - but you can use a combination of the environment variable and YAML configuration if you want to configure multiple error log files.
->
-> You will also need to make sure the user running the PHP process has write access to the log file, wherever you choose to put it.
+
+You don't need any of the YAML configuration above if you are using the `SS_ERROR_LOG` environment variable - but you can use a combination of the environment variable and YAML configuration if you want to configure multiple error log files.
+
+You will also need to make sure the user running the PHP process has write access to the log file, wherever you choose to put it.
```
emteknetnz, line 79: I've moved this bit out of the rather large !WARNING block above mainly to prevent linting errors because of the newly added !NOTE block below

ChloeSartorelli: Probably better this way anyway tbh.

en/06_Upgrading/07_Deprecations.md
```diff
@@ -73,10 +73,13 @@ SilverStripe\Core\Injector\Injector:
...
+
+> [!NOTE]
+> When deprecation notices are logged with monolog there will be additional context added by monolog about the original of the `user_error()` called in `Deprecation.php` which is unrelated the actual deprecated code, for example `{"code":16384,"message":"...","file":"/var/www/vendor/silverstripe/framework/src/Dev/Deprecation.php","line":355}`. This additional context should simply be ignored.
```
ChloeSartorelli:
```suggestion
> When deprecation notices are logged with monolog there will be additional context added by monolog about the origin of the `user_error()` called in `Deprecation.php` which is unrelated the actual deprecated code, for example `{"code":16384,"message":"...","file":"/var/www/vendor/silverstripe/framework/src/Dev/Deprecation.php","line":355}`. This additional context should simply be ignored.
```

GuySartorelli (changes requested):
Minor typo to fix.

en/06_Upgrading/07_Deprecations.md
emteknetnz: Updated

GuySartorelli (approved):
LGTM

Issue #741, opened 2025-05-15
Drop support for silverstripe/hybridsessions


GuySartorelli:
As per discussion in https://github.com/silverstripe/silverstripe-framework/issues/11719#issuecomment-2881819664, `silverstripe/hybridsessions` will only be supported for 6.0. Once 6.1 is released, we won't be supporting the module anymore as it will be redundant to the session handlers provided in framework from that version onwards.

This will need to be an exception to our [policy for supported modules](https://docs.silverstripe.org/en/6/project_governance/supported_modules/) which currently states
>  These modules will be supported for the lifetime of Silverstripe CMS 6 in the provided major versions.

That doesn't come into effect until after 6.0 goes stable. That means we have until June to make a documented exception for this module.

## Important note

> [!IMPORTANT]
> Subsequent discussion has lead to dropping support from 6.0 instead as that's easier to communicate and easier to maintain.

## Acceptance criteria

- ~~CMS 6 supported modules documentation clearly states that the hybridsessions module is an exception which is only supported for CMS 6.0 (aka `silverstripe/hybridsessions` 4.0)~~
- [x] silverstripe/hybridsessions is marked as CMS 6 compatible without commercial support

## PRs

- https://github.com/silverstripe/developer-docs/pull/742

PR #742, opened 2025-05-15
DOC Remove commercial support for silverstripe/hybridsessions

GuySartorelli:
## Issues

- https://github.com/silverstripe/developer-docs/issues/741


emteknetnz (changes requested):
Wouldn't it be cleaner to just not support it in the first place and only provide a CMS 6 compatible version? That way we don't need a special condition* just for this one module.

GuySartorelli:
1. That's not what has been approved.
2. That would mean for a period of time there would be no supported non-blocking session handler that doesn't require specific infrastructure.

SimulatedPanda:
All good to remove support from 6.0

GuySartorelli:
Updated to remove support in 6.0 instead. I've also updated https://github.com/silverstripeltd/product-issues/issues/887

emteknetnz (approved):

PR #749, opened 2025-05-26
DOC Document gridfield bugfix that is worth noting

GuySartorelli:
This is the sort of thing that someone could _potentially_ be relying on, although it was always incorrect behaviour. Seems sensible to document this change so anyone relying on it can hopefully see it (or if they don't see it, we can point at it if they raise it as a bug).

## Issue

- https://github.com/silverstripe/silverstripe-framework/issues/11739


emteknetnz (changes requested):
I've got existing PRs to fix the linting issues [here](https://github.com/silverstripe/.github/issues/410), so removing those fixes from this PR?

The `_config.php` one in particular is problematic because it'll just happen again for the stable release I believe since it's auto-generated from commits

GuySartorelli:
Dropped the second commit. CI failures for unrelated linting are now expected.

emteknetnz (approved):

PR #753, opened 2025-05-29
DOC Create unified release policy

emteknetnz:
Issue ~~https://github.com/silverstripeltd/product-issues/issues/905~~
https://github.com/silverstripeltd/product-issues/issues/908


GuySartorelli:
I think we should add redirects for the old pages to redirect to the new page.

It _might_ also be worth updating the logic for when you swap pages, so that if you're viewing the CMS 5 policies and swap to v6 docs you get redirected to the new page, but that's more effort and less value so I'm not as concerned about that scenario.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,80 @@
+---
+title: Release policy
+summary: Outline the release lifecycle and support commitment for Silverstripe CMS
+icon: code-branch
+---
+
+# Silverstripe CMS release policy
+
+This document outlines the unified release policy for Silverstripe CMS and its commercially supported modules. Its purpose is to provide clear guidance on the types of releases, their cadence, scope, and support lifecycle, ensuring predictability and stability for our users and developers.
```
ChloeSartorelli:
```suggestion
This document outlines the release policy for Silverstripe CMS. It applies to all [commercially supported modules](/project_governance/supported_modules/).

Community modules are not covered by this policy. Modules in the silverstripe GitHub organisation that are not commercially supported are updated on a best effort basis.

The purpose of this policy is to provide clear guidance on the types of releases, their cadence, scope, and support lifecycle, ensuring predictability and stability for our users and developers.
```
"Unified" is a good word to use in our comms about this because it's describing the change that's happening, but it doesn't really fit in the policy itself. Please remove it anywhere it might be in this document.

Other changes in the suggestion reflect wording from the "Scope of the policy" section from both existing policies.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,80 @@
...
+
```
ChloeSartorelli, line 30:
```suggestion

> [!NOTE]
> Refer to our [definition of public API](/project_governance/public_api/).

```
This is in both existing policies.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,80 @@
...
+## Release types and cadence
+
+Silverstripe CMS adheres to [Semantic versioning](https://semver.org/) for its release philosophy.
+
+### Major releases `(x.0.0)`
+
+Major releases are the cornerstone for introducing significant evolution and breaking changes to the Silverstripe CMS platform.
```
ChloeSartorelli: Where did this language come from? This is meant to be a blending of the two policies - I'd expect most if not all of the wording to come from them.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,80 @@
...
+- **Scope:** This policy applies to all Silverstripe CMS commercially supported modules. At launch, the list of commercially supported modules for that specific major release line will be explicitly defined.
+
+- **Pre-release period:** A stable major release is always preceded by a **beta period of at least three months** to allow for extensive testing and feedback.
+
+- **Dependency alignment:** At launch, Silverstripe CMS major releases aim to:
+
+  - Support all PHP versions currently in full support.
+  - Have all PHP and JavaScript dependencies on their latest stable versions.
+
+- **Upgrade path:** Silverstripe CMS commits to providing a clear upgrade path to new major releases, including comprehensive documentation on all breaking changes and migration guides.
```
ChloeSartorelli: This seems to all be new?

GuySartorelli (changes requested):
I stopped reviewing this when I realised that most of the content appears to be new, instead of coming from the existing policies.

We might need to discuss what we're expecting here. My expectation was that wording would be almost exclusively from the two existing policies, with the only changes being:
1. they're in the same document
2. some words are updated to reflect the changes in the support timeline page

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Updated

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Updated

en/12_Project_Governance/05_Release_policy.md
emteknetnz: It's new language. Have rewritten

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Correct

emteknetnz:
> My expectation was that wording would be almost exclusively from the two existing policies ...

The problem with that approach is that we'll basically just end up with the two existing documents together in the same document, one of top of the each other, to make one really long document.  I found the existing documents to be too wordy so I reduced the number of words so that it reads much clearer IMO

There's no actual change it what the policies state (unless I made a mistake writing the new one), it's just less words (and slightly different language)

emteknetnz:
PR for redirects https://github.com/silverstripe/doc.silverstripe.org/pull/334

en/12_Project_Governance/05_Release_policy.md
ChloeSartorelli: Not updated?

en/12_Project_Governance/05_Release_policy.md
ChloeSartorelli: Not added?

en/12_Project_Governance/05_Release_policy.md
ChloeSartorelli: >> My expectation was that wording would be almost exclusively from the two existing policies ...
>
> The problem with that approach is that we'll basically just end up with the two existing documents together in the same document, one of top of the each other, to make one really long document. I found the existing documents to be too wordy so I reduced the number of words so that it reads much clearer IMO
>
> There's no actual change it what the policies state (unless I made a mistake writing the new one), it's just less words (and slightly different language)

........ that's gonna make this really hard to review but I'll try. I suspect a lot of my review comments will be effectively a rewrite that brings this more in-line with the combined language of the current policies.

GuySartorelli (changes requested):

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,80 @@
...
+icon: code-branch
+---
+
+# Silverstripe CMS release policy
+
+This document outlines the unified release policy for Silverstripe CMS and its commercially supported modules. Its purpose is to provide clear guidance on the types of releases, their cadence, scope, and support lifecycle, ensuring predictability and stability for our users and developers.
+
+## Release types and cadence
+
+Silverstripe CMS adheres to [Semantic versioning](https://semver.org/) for its release philosophy.
```
ChloeSartorelli: This probably belongs above, along with the link to the definition of public API.

ChloeSartorelli: In this section add the following which is language blended from the two existing policies:

```suggestion
Silverstripe CMS aims to deliver regular releases at predefined intervals with a clear support timeline. This is to allow Silverstripe CMS project owners to plan upgrades ahead of time.

Note that the release cadence and pre-release time frames are indicative only. We may alter them in response to external constraints.
```

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Communication and documentation
+
+For all release types, Silverstripe CMS is committed to clear and timely communication:
+
+- **Release announcements:** All major, minor, and patch releases will be announced through official Silverstripe channels.
+
+- **Release notes:** Comprehensive release notes detailing new features, improvements, bug fixes, and security advisories will be published for every release.
+
+- **Upgrade guides:** For major releases, detailed upgrade guides and documentation on breaking changes will be provided to assist users in migrating to the new version.
```
ChloeSartorelli: Was there anything about communication or documentation in our existing policies? Other than a single sentence in the major release policy (below for convenience) I couldn't see anything.

> This includes clear documentation on what breaking changes have been made to each area of Silverstripe CMS.

ChloeSartorelli: Remember the goal here is to combine the existing policies. Anything that wasn't already part of those policies should be done separately and requires additional scrutiny.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Release types and cadence
+
+Silverstripe CMS adheres to [Semantic versioning](https://semver.org/) for its release philosophy.
+
+### Major releases `(x.0.0)`
+
+Major releases are the cornerstone for introducing significant evolution and breaking changes to the Silverstripe CMS platform.
+
+- **Purpose:** To deliver breaking changes and architectural updates that cannot be shipped in minor releases due to backward compatibility constraints.
```
ChloeSartorelli:
```suggestion
```
Not needed - this is the main blurb of the section as above.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,80 @@
...
+
+- **Backward compatibility:** Major releases are **not backward compatible**.
+
+- **Frequency:** New Silverstripe CMS major versions are released every **two years**.
+
+- **Tagging period:** New major releases are typically tagged as stable between **April and June** of their release year.
+
+- **Scope:** This policy applies to all Silverstripe CMS commercially supported modules. At launch, the list of commercially supported modules for that specific major release line will be explicitly defined.
```
ChloeSartorelli:
```suggestion
```
The scope is defined at the start of the policy and applies to the whole policy. There's no need to have a separate scope here.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,80 @@
...
+
+Major releases are the cornerstone for introducing significant evolution and breaking changes to the Silverstripe CMS platform.
+
+- **Purpose:** To deliver breaking changes and architectural updates that cannot be shipped in minor releases due to backward compatibility constraints.
+
+- **Backward compatibility:** Major releases are **not backward compatible**.
+
+- **Frequency:** New Silverstripe CMS major versions are released every **two years**.
+
+- **Tagging period:** New major releases are typically tagged as stable between **April and June** of their release year.
```
ChloeSartorelli:
```suggestion
- **Cadence:** New Silverstripe CMS major versions are released every two years between April and June of their release year.
```

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,80 @@
...
+
+- **Scope:** This policy applies to all Silverstripe CMS commercially supported modules. At launch, the list of commercially supported modules for that specific major release line will be explicitly defined.
+
+- **Pre-release period:** A stable major release is always preceded by a **beta period of at least three months** to allow for extensive testing and feedback.
```
ChloeSartorelli:
```suggestion
- **Pre-release period:** A stable major release is preceded by a **beta period of at least three months** to allow for extensive testing and feedback.
```

ChloeSartorelli: This is missing some context from the Major Release Policy:

> The purpose of the beta period is to:
> 
> - allow early adopters to plan their upgrade
> - identify bugs in the major release
> - help community modules add support for the new major release early on.
>
> Once a new major enters the beta period, it should be compatible with the dependencies intended to ship in the stable release. Substantial breaking changes should be avoided post-beta.
>
> Prior to the beta period, alpha releases are published at the discretion of the development team.

I recommend having a section outside of the "major" and "minor" sections that explains pre-releases and their intended purpose, since that applies to both types of release.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,80 @@
...
+
+- **Backward compatibility:** Patch releases are **backward compatible** within their major and minor release line. They will not introduce new features or breaking changes.
+
+- **Frequency:** Patch releases are shipped on an as-needed basis in response to identified bugs or security vulnerabilities. They do not follow a fixed schedule.
+
+## Support lifecycle
+
+Silverstripe CMS maintains a clear support lifecycle for versions that have a stable releases to provide long-term stability and planning for users. Each minor release version progresses through defined support phases:
+
+- **Full support:** Receives bug fixes and updates for all indentified security vulnerabilies.
```
ChloeSartorelli:
```suggestion
- **Full support:** Receives bugfixes that do not change existing public API and security patches for identified security vulnerabilities at all impact levels.
```
Language taken from "Bug and security fixes" section of the minor release policy - though I personally prefer the longer form from that policy especially as it details explicitly what is _not_ included in this support level

```suggestion
### Full Support

A Silverstripe CMS minor release line enters the *full support* phase once it is tagged stable.

A minor release in the *full support* phase receives:

- bugfixes that do not change existing public API
- security patches for vulnerabilities at all impact levels.

It does **not** receive:

- new features
- new public API.

A minor release line with a standard support length stays in the *full support* phase until the next minor release is tagged (approximately 6 months).

A minor release line with extended support stays in the *full support* phase for one year.
```

Note that this more verbose version has several advantages:

1. It's (mostly) straight from the existing policies so we _know_ it's not changing our actual or percieved commitments
2. It lists in clear bullet points exactly what IS and what ISN'T included in this support phase.
3. It indicates the transition points leading into and out of this support phase
4. It provides an anchor link

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,80 @@
...
+
+Patch releases address critical issues swiftly to maintain stability and security.
+
+- **Purpose:** To deliver bug fixes and security updates.
+
+- **Backward compatibility:** Patch releases are **backward compatible** within their major and minor release line. They will not introduce new features or breaking changes.
+
+- **Frequency:** Patch releases are shipped on an as-needed basis in response to identified bugs or security vulnerabilities. They do not follow a fixed schedule.
+
+## Support lifecycle
```
ChloeSartorelli: This whole section specifically mentions "minor release" so should either be moved into that section or else should also talk about major and patch releases.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,80 @@
...
+
+Silverstripe CMS maintains a clear support lifecycle for versions that have a stable releases to provide long-term stability and planning for users. Each minor release version progresses through defined support phases:
+
+- **Full support:** Receives bug fixes and updates for all indentified security vulnerabilies.
+
+- **Partial support:** Receives no bug fixes. Receives security updates for high and critical security vulnerabilities only (defined as those with a **CVSS score of 7.0 or above**). Low and medium level security vulnerabilities are not fixed.
```
ChloeSartorelli:
```suggestion
### Partial support

At the end of the *full support* phase, a Silverstripe CMS minor release line transitions to the *partial support* phase.

Releases in  *partial support* receive security updates for high and critical security vulnerabilities only (defined as those with a CVSS score of 7.0 or above).

These releases **do not** receives bug fixes or patches for security vulnerabilitties with a low or medium severity rating.

A minor release line with a standard support length stays in the *partial support* phase for 6 months.

A minor release line with extended support stays in the *partial support* phase for approximately one year. It exits the *partial support* phase once the next major release has entered *full support*.
```
This suggestion assumes we go with the heading style in the above suggestion which allows us to be more explicit about what the policy covers.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,80 @@
...
+
+- **End-of-life (EOL):** These minor versions receive no further support, bug fixes, or security updates. Users are strongly encouraged to upgrade to a supported version.
```
ChloeSartorelli:
```suggestion
### End-of-life

Once the *partial support* phase for a minor release line has ended, it goes *end-of-life*.

These minor versions receive no further support, bug fixes, or security updates. Users are strongly encouraged to upgrade to a supported version.
```

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,80 @@
...
+
+Most minor releases start with **six months** of full support which is followed by **six months** of partial support. The last minor release of major version, typically the X.4 release, has an extended support length which is **one year**.of full support which is followed by **one year** of partial support.
```
ChloeSartorelli:
```suggestion
```
I think it's cleaner to outline this information inside sections for each support phase.

The fact that the last minor release of a given major is the one with extended support should either be at the start of the lifecycle section instead of the end. I'll add a suggestion there with some wording.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,80 @@
...
+
+- **Purpose:** To deliver bug fixes and security updates.
+
+- **Backward compatibility:** Patch releases are **backward compatible** within their major and minor release line. They will not introduce new features or breaking changes.
+
+- **Frequency:** Patch releases are shipped on an as-needed basis in response to identified bugs or security vulnerabilities. They do not follow a fixed schedule.
+
+## Support lifecycle
+
+Silverstripe CMS maintains a clear support lifecycle for versions that have a stable releases to provide long-term stability and planning for users. Each minor release version progresses through defined support phases:
```
ChloeSartorelli:
```suggestion
Silverstripe CMS maintains a clear support lifecycle for versions that have a stable releases to provide long-term stability and planning for users. Each minor release version progresses through defined support phases.

The last minor release of major release line has an extended support length, which provides one year of *full support* following by approximately one year of *partial support*.
```

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,80 @@
...
+- **Tagging period:** New major releases are typically tagged as stable between **April and June** of their release year.
+
+- **Scope:** This policy applies to all Silverstripe CMS commercially supported modules. At launch, the list of commercially supported modules for that specific major release line will be explicitly defined.
+
+- **Pre-release period:** A stable major release is always preceded by a **beta period of at least three months** to allow for extensive testing and feedback.
+
+- **Dependency alignment:** At launch, Silverstripe CMS major releases aim to:
+
+  - Support all PHP versions currently in full support.
+  - Have all PHP and JavaScript dependencies on their latest stable versions.
```
ChloeSartorelli: Is this meant to cover everything that's in "PHP support commitments" and "dependency management" in the current major release policy?

If so the following is missing from the "PHP support commitments" section:

> The Silverstripe CMS release cycle is built around these assumptions:
> 
> - new PHP releases are shipped at the end of November on an annual basis
> - PHP releases are in full support for two years with one year of limited support.

> PHP versions in limited support are not supported at launch by new Silverstripe CMS major releases.

> Following the initial launch of a Silverstripe CMS major release, the development team aims to add forward compatibility for the next PHP release. e.g: Silverstripe CMS 7 at launch will support PHP 8.5 and PHP 8.6. CMS 6 should receive official support for an eventual PHP 8.5 in early 2026.

> Support for end-of-life PHP releases is not dropped within a Silverstripe CMS major release line, unless it's necessary to address vulnerabilities or high impact bugs.

> Major release lines not in active development do not receive official support for later PHP releases.

And the following is missing form the "dependency management" section:

> For dependencies offering a *Long Term Support* version, the latest LTS of that dependency is preferred.
>
> A list of *fixed dependencies* for each Silverstripe CMS major release line is explicitly defined by the development team. *Fixed dependencies* do not change for the lifetime of a Silverstripe CMS major release line.
> 
> Support for newer major releases of fixed dependencies is not added to Silverstripe CMS major release line, even if the currently used version is end-of-life. A Silverstripe CMS major release line only supports one major version of each fixed dependency.
> 
> The only exceptions are:
> 
> - if the upgrade is necessary to fix a high impact bug or security vulnerability
> - if the upgrade is necessary to add support for a new PHP version to a Silverstripe CMS major release in active development.
> 
> Non-fixed dependencies may be upgraded to new major releases or swapped out altogether within a Silverstripe CMS release line.


My recommendation is to just add the full "PHP support commitments" and "dependency management"  sections.

emteknetnz:
Ah sorry didn't push

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Hadn't pushed, updated. It's in the same place as the major/minor policies i.e. below "Community modules are not covered by this policy ..."

en/12_Project_Governance/05_Release_policy.md
ChloeSartorelli: The note about commercially supported module list changing at launch was also missing some context from the "Commercially supported module" section. I recommend adding that full section into the policy so we don't lose context.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,87 @@
...
+
+- **Upgrade path:** Silverstripe CMS commits to providing a clear upgrade path to new major releases, including comprehensive documentation on all breaking changes and migration guides.
```
ChloeSartorelli:
```suggestion
- **Upgrade path:** Silverstripe CMS commits to providing a [clear upgrade path](/upgrading) to new major releases. This includes clear documentation on all breaking changes and a changelog.
```
Adding link from the existing policy and using some more of the existing language.

I'm not sure what you mean by migration guide - what migration guides did we provide for CMS 6? Unless you mean the changelog? I don't think we should be calling the changelog a migration guide.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,87 @@
...
+> [!NOTE]
+> Refer to our [definition of public API](/project_governance/public_api/).
+
+The purpose of this policy is to provide clear guidance on the types of releases, their cadence, scope, and support lifecycle, ensuring predictability and stability for our users and developers.
+
+## Release types and cadence
+
+Silverstripe CMS adheres to [Semantic versioning](https://semver.org/) for its release philosophy.
+
+### Major releases `(x.0.0)`
```
ChloeSartorelli: The following major release policy information is missing:

- At any given time, only two major release lines are officially supported.
- symfony support commitments
- context from the "Active development" section (it looks like you might be changing the context of the policy from talking about a major release *line* to just the major x.0.0 release itself? Whichever of those is intended needs to be very explicitly stated)
- At any given time, only two major release lines are officially supported.


Other things are also missing but are explicitly pointed out in other comments.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,87 @@
...
+# Silverstripe CMS release policy
+
+This document outlines the release policy for Silverstripe CMS. It applies to all [commercially supported modules](/project_governance/supported_modules/).
+
+Community modules are not covered by this policy. Modules in the silverstripe GitHub organisation that are not commercially supported are updated on a best effort basis.
+
+> [!NOTE]
+> Refer to our [definition of public API](/project_governance/public_api/).
+
+The purpose of this policy is to provide clear guidance on the types of releases, their cadence, scope, and support lifecycle, ensuring predictability and stability for our users and developers.
```
ChloeSartorelli: I think these should be the other way around but I'm not gonna hold up merging for it.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Release types and cadence
+
+Silverstripe CMS adheres to [Semantic versioning](https://semver.org/) for its release philosophy.
+
+### Major releases `(x.0.0)`
+
+Major releases include architectural updates, allow breaking changes and major dependency changes.
```
ChloeSartorelli:
```suggestion
The main function of major releases is to ship breaking changes that can not be shipped in minor releases and update third-party dependencies. These changes are primarily aimed at making sure the major releases can be supported for the full duration of its lifecycle.
```
From the existing policy - updated a bit to include third party dependencies which comes from the list "Most of the changes shipped in a major release will:"

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,87 @@
...
+  - Support all PHP versions currently in full support.
+  - Have all PHP and JavaScript dependencies on their latest stable versions.
+
+- **Upgrade path:** Silverstripe CMS commits to providing a clear upgrade path to new major releases, including comprehensive documentation on all breaking changes and migration guides.
+
+### Minor releases `(x.y.0)`
+
+Minor releases provide ongoing value and improvements within a stable major release line. They are backwards compatible within the major release line and designed to be easy to upgrade to.
+
+- **Purpose:** To ship new features, enhancements, performance improvements, and bug fixes that cannot be shipped in a patch releae due to API additions.
```
ChloeSartorelli:
```suggestion
Minor releases provide ongoing value and improvements within a stable major release line. This is where we ship new features and enhancements, as well as bug fixes that are deemed too risky to be shipped in a patch release due to API changes.
```

en/12_Project_Governance/05_Release_policy.md
ChloeSartorelli, line 1: We're missing the "Security patch windows" section from the minor release policy.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,87 @@
...
+
+- **Frequency:** Minor releases are shipped every **six months** within an active major release line, typically targeted for **April and October**. There will be **five** minor releases per major release.
+
+- **Pre-release period:** Minor releases are typically preceded by a **beta period of approximately six weeks** and a **release candidate (RC) period of approximately two weeks**. A feature freeze is implemented after the beta phase.
+
+### Patch releases `(x.y.z)`
+
+Patch releases address critical issues swiftly to maintain stability and security.
+
+- **Purpose:** To deliver bug fixes and security updates.
```
ChloeSartorelli:
```suggestion
- **Purpose:** To deliver bug fixes
```
Security patches should be treated in their own section since those do have specific release windows and have a different scope.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Minor releases `(x.y.0)`
+
+Minor releases provide ongoing value and improvements within a stable major release line. They are backwards compatible within the major release line and designed to be easy to upgrade to.
+
+- **Purpose:** To ship new features, enhancements, performance improvements, and bug fixes that cannot be shipped in a patch releae due to API additions.
+
+- **Backward compatibility:** Minor releases are **backward compatible** within their major release line. They will not introduce breaking changes.
+
+- **Frequency:** Minor releases are shipped every **six months** within an active major release line, typically targeted for **April and October**. There will be **five** minor releases per major release.
```
ChloeSartorelli:
```suggestion
- **Frequency:** Minor releases are shipped every **six months** within an active major release line around **April and October**.
```

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,87 @@
...
+
+- **Pre-release period:** Minor releases are typically preceded by a **beta period of approximately six weeks** and a **release candidate (RC) period of approximately two weeks**. A feature freeze is implemented after the beta phase.
```
ChloeSartorelli:
```suggestion
- **Pre-release schedule:** Approximately 6 weeks prior to the anticipated stable minor release, a beta minor release is tagged. Approximately 2 weeks prior to the anticipated stable minor release, a release candidate is tagged.
```
Calling these "periods" is weird - we tend to talk about these as releases (beta release, rc release) not periods of time. The existing minor release policy also talks about these as releases. We should probably maintain that.

Note that I have removed context about the feature freeze.
As I've suggested in anotehr comment, I think there should be a separate dedicated section talking about pre-releases which can include information like feature freezes and the purpose of pre-releases. The minor-release specific information really just needs to mention when these releases are in relation to the stable release.

en/12_Project_Governance/05_Release_policy.md
ChloeSartorelli: The existing policy currently says "Silverstripe CMS **_aims to ship_** a new minor releases every 6 months" (emphasis added).

This  "aims to" technically gives us an out if for some reason we want to skip a release or sneak an extra one in. We should probably get product owner approval before changing that language.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,87 @@
...
+- **Dependency alignment:** At launch, Silverstripe CMS major releases aim to:
+
+  - Support all PHP versions currently in full support.
+  - Have all PHP and JavaScript dependencies on their latest stable versions.
+
+- **Upgrade path:** Silverstripe CMS commits to providing a clear upgrade path to new major releases, including comprehensive documentation on all breaking changes and migration guides.
+
+### Minor releases `(x.y.0)`
+
+Minor releases provide ongoing value and improvements within a stable major release line. They are backwards compatible within the major release line and designed to be easy to upgrade to.
```
ChloeSartorelli:
```suggestion
Minor releases provide ongoing value and improvements within a stable major release line. They are backwards compatible within the major release line and deliver new features and new public API in a backward compatible way.
```
Let's not use the "easy" word.
Suggested wording comes from the existing policy.

GuySartorelli (dismissed):
Once we've got this into a state we both agree on, we should get someone outside of this team - preferably someone without much technical background (perhaps a few people) to give it a read and make sure they understand it.

I'm thinking that for example `x.0.0` and `x.y.z` won't be as obvious to non-developers as they are to you and I.

## Review cycle

This pass was mostly about checking if the content reflects what's in the existing policies. I started trying to suggest structural changes as well but it got mentally challenging very quickly.
There are likely structural changes I will suggest, but those will be in a separate review pass.

I recommend we get the content broadly correct first, then make sure the structure is easily readable and has content in a logical order and format, and then get someone(s) outside the team to take a look.

GuySartorelli:
There's also a linting failure

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Updated, though I've left in the semver bit, I think it's useful

I've also updated the "Release types and cadence" to "Release type definitions"

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Mostly removed, added a little saying that there will be changelogs

en/12_Project_Governance/05_Release_policy.md
emteknetnz: I've gone the other way and removed the blurbs

en/12_Project_Governance/05_Release_policy.md
emteknetnz: > scope

Removed

> The note about commercially supported module list changing at launch was also missing some ...

I'm not sure what this is referring to

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Might have forgotten to push

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Updated

en/12_Project_Governance/05_Release_policy.md
emteknetnz: I don't think we need this, we're kind of just defining what a beta release is, which isn't unique to us

> Substantial breaking changes should be avoided post-beta.

This isn't even helpful in this context IMO, the word *should* doesn't provide any guarantees. It's like the sentence is aimed for internal use (as in we shouldn't change deps), rather than external people reading this.

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Have gone the heading and most of the short recommendation. I've also separately bolded "minor release version" above to clarify what this section refers to

I'd argue the existing long form isn't technically correct, we can ship "new features" in patches if they don't add any new API

en/12_Project_Governance/05_Release_policy.md
emteknetnz: I think it's helpful being in its own section, reason being whenever everyone things about a release "release", they're exclusively referring to minor releases. This includes majors, because we're still just releasing a minor e.g. "6.0".

The new support-timeline table is explicitly build around minor releases. A large part of the problem with the old support-timeline page was it tried to communicate support lengths for both minor and major releases, and it was extremely confusing.  The new support timeline simply doesn't communicate what the support length for a major is (because, it kind of doesn't exist), so I think it makes sense to only talk about minors in a "Support lifecycle" section.

If we moved this to section to inside the "Minor releases" section, then the question becomes "why isn't it in the patch and major releases" sections? We know they're not there because it's non-sensical, though others might not

I've added a note to the bottom saying that major and patch releases don't have support lifestyles associated with them

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Yup individual headings make sense

> It exits the *partial support* phase once the next major release

This bit I don't like, because we don't release majors, we only release minors (i.e. 6.0 is a minor), have adjusted words a bit

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Updated

en/12_Project_Governance/05_Release_policy.md
emteknetnz: I've changed the formatting for this + the other support levels so it matches up with the release definitions e.g. bolded bullets for "Security support" etc, to make it consistent and extra clear

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Have changed to the bolded bullets style

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Have changed to the bolded bullets style

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Have moved to each support phase

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Added para, though have changed words around a bit

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Have incorporated into a "Documentation" bit instead of "Upgrade path"

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Added

en/12_Project_Governance/05_Release_policy.md
emteknetnz: > At any given time, only two major release lines are officially supported.

Have added sentence to "Support lifecycle" section

> symfony support commitments

Added as sub-heading under "Dependency management" 

> Active development

This was a very confusing concept that previously applied to both the next-minor and the current major and sometimes the next major (I think??)

The old major policy had this "Only one major release line is in active development at any given time." - which is just wrong, or at least very unintuitive? We worked on CMS 5.4 and 6.0 at the same time.

I've simply remove references to "Active development", I don't think it's a helpful concept

On the new support-timeline it's been changed to "In development" and refers exclusively to up-coming minor releases, not majors

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Updated

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Done, though added to the "purpose" bolded bullet as I've gotten rid of the blurbs

emteknetnz:
@GuySartorelli I've eneded up rewritting large parts of this and have deviated somewhat from some of comments I've previously made, I've kind of given up trying to record all the various changes I've made, sorry.

What I'm trying to do is be very precise with language and talking about "versions" rather than "releases"

For instead using "minor version release" when specifically talking about a release, and "minor version" when talking about a specific version. I've removed the words "release line" since they have "release" in there which is confusing and very overused, "version" is a better word.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,176 @@
...
+
+Because of the use of semantic versioning for version numbering, there are three version release types which allow for a different levels of changes:
+
+- [Major version release](#major-version-release)
+- [Minor version release](#minor-version-release)
+- [Patch version release](#patch-version-release)
+
+### Major version (not a release)
+
+When the first minor version release (defined below) of a major version is performed, it gets the status of being both a major release, and a minor release. For instance the "CMS 6.0" minor release is also called the "CMS 6 major release"
```
emteknetnz: This bit is needs to change, however I'm not sure how to write this

We have probably too many concepts floating around:

- Semantic versions
- CMS versions
- CMS version releases
- Major releases ... which aren't actually a thing (only minor releases are things)

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,176 @@
...
+The purpose of this policy is to provide clear guidance on the types of versions and version releases, including their cadence, scope, and support lifecycle, ensuring predictability and stability for our users and developers.
+
+> [!NOTE]
+> Refer to our [definition of public API](/project_governance/public_api/).
+
+## Verion numbering
+
+### Semantic versioning
+
+Silverstripe CMS adheres to [Semantic versioning](https://semver.org/) for its version numbering of individual modules such as `silverstripe/framework`
```
emteknetnz: I'm not happy with this section, particularly as it's talking about modules rather than the CMS

However the concept of semantic versioning is core to releases and API/dependencies that are allowed to change within each release, so we probably need to clear definition to provide context?

It gets really confusing because I talk about "CMS versions" below which is what people in the real world actually refer to

emteknetnz:
@GuySartorelli This PR is very much in a draft state and definitely need some help to get it to make sense, it feels like we have too many concepts that need to be communicated

en/12_Project_Governance/05_Release_policy.md
ChloeSartorelli: I'm not saying to delete the semver bit, just move it:
> This probably belongs above, along with the link to the definition of public API.

My two comments were separate - GitHub just doesn't let me _make them_ separate.

GuySartorelli:
> it feels like we have too many concepts that need to be communicated

I agree. That's why my preference would have been to just combine the two policies first and include _all_ of the information - and only look at simplifying or removing information as a second step. It's very difficult to do both at the same time.

I haven't looked at the changes this review round pending a discussion, but I'm reassigning to you for the changes in the other PRs in the meantime.

GuySartorelli:
Note that the unified policy has been completely redone. Any comments before this one should be ignored.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,191 @@
...
+Regular Silverstripe CMS releases are performed at predefined intervals with a clear support timeline. This allows Silverstripe CMS project owners to plan upgrades ahead of time.
+
+> [!NOTE]
+> The cadence and pre-release timeframes are indicative only. We may alter those in response to external constraints.
+
+### Silverstripe CMS major release
+
+A new Silverstripe CMS major version series (e.g. "CMS 6") is introduced by its first Silverstripe CMS release (e.g. "CMS 6.0").
+
+The main purpose of a major release is to deliver significant architectural updates and changes that are not backward compatible. This ensures the Silverstripe CMS major version series can be supported for its full lifecycle, including updates to third-party dependencies.
```
ChloeSartorelli:
```suggestion
The main purpose of a major release is to deliver significant architectural updates and changes that are not backward compatible, including updates to third-party dependencies. This ensures the Silverstripe CMS major version series can be supported for its full lifecycle.
```
Otherwise it sounds like "updates to third-party dependencies" is part of what happens in the full lifecycle, but while that _can_ happen for non-fixes dependencies, we ideally aim to keep dependency upgrades limited to major releases unless we _have_ to do them in minors.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,191 @@
...
+
+At launch, new Silverstripe CMS major versions may contain new features not present in the previous major version series. However, shipping new features is a secondary concern, as new features will usually ship in Silverstripe CMS minor releases.
```
ChloeSartorelli:
```suggestion
At launch, new Silverstripe CMS major versions may contain new features not present in the previous major version series. However, shipping new features is a secondary concern, as new features will usually ship in minor releases.
```
Just feels weird to repeat "Silverstripe CMS" so much, and "Silverstripe CMS minor release" wasn't a specifically defined concept - though it does reflect back to "Silverstripe CMS release".....

Ultimately I don't feel strongly about this, so up to you.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,191 @@
...
+
+- **Backward compatibility:** No - projects will require some development work to upgrade to a new major release
+- **Cadence:** Every two years, between April and June
+- **Pre-release schedule:**
+  - **Alpha**: At the discretion of the development team
+  - **Beta**: At least 3 months prior to the anticipated Silverstripe CMS major release. At this point, dependencies should not change
```
ChloeSartorelli: The existing major release does say this about beta releases:
> Substantial breaking changes should be avoided post-beta

I recommend we just slap that in here:
```suggestion
  - **Beta**: At least 3 months prior to the anticipated Silverstripe CMS major release. At this point, dependencies should not change, and substantial breaking changes should be avoided.
```

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,191 @@
...
+  - **RC**: Approximately 4 weeks prior to the anticipated Silverstripe CMS major release. At this point, no more breaking changes can be made unless they are to fix a regression
```
ChloeSartorelli: Note that the existing major release policy is quiet on the subject of release candidates - but given this specifically says "approximately" and we have a note that "the cadence and pre-release timeframes are indicative only", I think this addition is in line with our actual commitments and isn't contradicting the existing policy.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Definitions
+
+- **Silverstripe CMS version**: The version number corresponding to a Silverstripe CMS release, e.g. "6.1"
+- **Silverstripe CMS release**: A combined release of commercially supported modules which together make up Silverstripe CMS. Also known as a "stable release"
+- **Silverstripe CMS major version series**: All Silverstripe CMS versions that start with the same whole number, e.g. "6.0", "6.1", and "6.2" are all in the same "CMS 6" major release series
+
+### Pre-release definitions
+
+- **Alpha**: A pre-release that may be performed prior to a "beta pre-release". Its purpose is to allow for early exploration and feedback on foundational changes, new architectural approaches, or significant new features
```
ChloeSartorelli:
```suggestion
- **Alpha**: A pre-release that may be performed prior to a "beta" pre-release. Its purpose is to allow for early exploration and feedback on foundational changes, new architectural approaches, or significant new features
```

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,191 @@
...
+- **Beta**: A pre-release is performed prior to a "RC pre-release". Its purpose is to facilitate testing and feedback, allow early adopters to plan their upgrade, to identify bugs, and to help community modules dependent on Silverstripe CMS to upgrade ahead of the stable release
```
ChloeSartorelli:
```suggestion
- **Beta**: A pre-release which is performed prior to a "RC" pre-release. Its purpose is to facilitate testing and feedback, allow early adopters to plan their upgrade, to identify bugs, and to help community modules dependent on Silverstripe CMS to upgrade ahead of the stable release
```

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,191 @@
...
+- **Release candidate (aka RC)**: A pre-release that is performed prior to a stable release. A copy of the release candidate (which also includes any undisclosed security patches) will be sent to an external auditor for a security review
```
ChloeSartorelli:
```suggestion
- **Release candidate (aka RC)**: A pre-release that is performed prior to a stable release. A copy of the release candidate (which also includes any undisclosed security patches) will be sent to an external auditor for a security review. This pre-release should be very similar to the final stable release and is a last chance to find any regressions before then.
```
Just noticed we have a "purpose" for the other two but not this one. I'm not strongly tied to this wording so feel free to change it how you feel is best.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,191 @@
...
+  - **Alpha**: At the discretion of the development team
+  - **Beta**: At least 3 months prior to the anticipated Silverstripe CMS major release. At this point, dependencies should not change
+  - **RC**: Approximately 4 weeks prior to the anticipated Silverstripe CMS major release. At this point, no more breaking changes can be made unless they are to fix a regression
+- **Documentation:** A changelog is published listing significant new features and breaking changes. A [clear upgrade path](/upgrading) is available
+
+### Silverstripe CMS minor release
+
+A Silverstripe CMS minor release is the standard recurring release of Silverstripe CMS (e.g. "CMS 6.1", "CMS 6.2"). Minor release upgrades are generally more straightforward to manage than major release upgrades.
+
+- **Purpose:** Ship new features and enhancements
```
ChloeSartorelli: I don't feel strongly either way about this - but it feels a little odd having a "purpose" here when we removed that from the major release table-like-structure.

Maybe turn this into a sentence and slap it into the above paragraph? But again I don't feel strongly either way, it can stay as is if you prefer.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,191 @@
...
+## Types of Silverstripe CMS releases
+
+Regular Silverstripe CMS releases are performed at predefined intervals with a clear support timeline. This allows Silverstripe CMS project owners to plan upgrades ahead of time.
+
+> [!NOTE]
+> The cadence and pre-release timeframes are indicative only. We may alter those in response to external constraints.
+
+### Silverstripe CMS major release
+
+A new Silverstripe CMS major version series (e.g. "CMS 6") is introduced by its first Silverstripe CMS release (e.g. "CMS 6.0").
```
ChloeSartorelli:
```suggestion
A new Silverstripe CMS major version series (e.g. "CMS 6") is introduced by its first Silverstripe CMS release (e.g. "CMS 6.0.0").
```
I think technically since we defined the "version" as x.y, the release would be identified by its full tag name.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,191 @@
...
+- **Cadence:** Every two years, between April and June
+- **Pre-release schedule:**
+  - **Alpha**: At the discretion of the development team
+  - **Beta**: At least 3 months prior to the anticipated Silverstripe CMS major release. At this point, dependencies should not change
+  - **RC**: Approximately 4 weeks prior to the anticipated Silverstripe CMS major release. At this point, no more breaking changes can be made unless they are to fix a regression
+- **Documentation:** A changelog is published listing significant new features and breaking changes. A [clear upgrade path](/upgrading) is available
+
+### Silverstripe CMS minor release
+
+A Silverstripe CMS minor release is the standard recurring release of Silverstripe CMS (e.g. "CMS 6.1", "CMS 6.2"). Minor release upgrades are generally more straightforward to manage than major release upgrades.
```
ChloeSartorelli:
```suggestion
A Silverstripe CMS minor release is the standard recurring release of Silverstripe CMS (e.g. "CMS 6.1.0", "CMS 6.2.0"). Minor release upgrades are generally more straightforward to manage than major release upgrades.
```
I think technically since we defined the "version" as x.y, the release would be identified by its full tag name.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,191 @@
...
+
+- **Purpose:** Ship new features and enhancements
+- **Backward compatibility:** Yes - within their Silverstripe CMS major version series
+- **Cadence:** Every six months within an active Silverstripe CMS major version series. Minor releases are targeted for April and October each year
```
ChloeSartorelli:
```suggestion
- **Cadence:** Every six months within the most recent stable Silverstripe CMS major version series. Minor releases are targeted for April and October each year
```
We haven't defined "active", so I think it pays to be a little more verbose about what we mean here. Theoretically since CMS 5 is still in some level of support, it could be interpreted as being "active".

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,191 @@
...
+- **Pre-release schedule:**
+  - **Alpha**: Typically not released
```
ChloeSartorelli:
```suggestion
  - **Alpha**: Typically not performed
```
Since we've decided these aren't "releases" for the purposes of this document

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,191 @@
...
+  - **Beta**: Approximately 6 weeks prior to the anticipated Silverstripe CMS minor release. At this point, no new features or new public API should be added
+  - **RC**: Approximately 2 weeks prior to the Silverstripe CMS release. At this point, only bug fixes should be added
```
ChloeSartorelli:
```suggestion
  - **RC**: Approximately 2 weeks prior to the Silverstripe CMS minor release. At this point, only bug fixes should be added
```
To match with language in the beta bit above.
I'd also prefer to remove "Silverstripe CMS" from both of these, and maybe change to "stable minor release" - but I don't feel strongly about that.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,191 @@
...
+> [!NOTE]
+>
+> While each Silverstripe CMS release usually only gets one of each type of pre-stable release, additional pre-stable releases may be tagged if required.
+>
+> Patch releases for individual modules such as `silverstripe/framework` are not considered a "Silverstripe CMS release", however they are an important part of meeting our support commitments.
+
+## Support lifecycle
+
+> [!INFO]
+> For a detailed timeline of current and planned support durations for each Silverstripe CMS version, refer to the [support timeline](https://www.silverstripe.org/software/roadmap).
```
ChloeSartorelli:
```suggestion
> For a detailed timeline of current and planned support durations for each Silverstripe CMS version, refer to the [support timeline](https://www.silverstripe.org/software/roadmap#support-timeline).
```
I _assume_ that will be the anchor once the page has been updated.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,191 @@
...
+
+A clear support lifecycle for each Silverstripe CMS version is defined. This provides long-term stability and planning for users.
+
+At any given time, only two major version series are officially supported. For example, CMS 5.4, CMS 6.0, and CMS 6.1 can all be supported at once as they span a total of two major version series, however, CMS 5.4, CMS 6.0, and CMS 7.0 cannot all be supported at once.
+
+Each Silverstripe CMS version has a *standard* support length, except for the last minor release of a Silverstripe CMS major version series, which has an  *extended* support length, e.g. CMS 5.4 has extended support. Details about how this affects release timelines is detailed further below.
```
ChloeSartorelli:
```suggestion
Each Silverstripe CMS version has a *standard* support length, except for the last minor version of a Silverstripe CMS major version series, which has an  *extended* support length, e.g. CMS 5.4 has extended support. Details about how this affects release timelines is detailed further below.
```

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,191 @@
...
+
+Support commitments for security and bug fixes are met through patch releases on individual supported modules and are detailed further in a section below.
+
+Each Silverstripe CMS version progresses through three defined support phases.
+
+### Full support
+
+A Silverstripe CMS version enters the *full support* phase after the corresponding release for has been performed.
```
ChloeSartorelli:
```suggestion
A Silverstripe CMS version enters the *full support* phase after the corresponding release has been performed.
```

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,191 @@
...
+Any security fixes that are shipped at the same time as a new Silverstripe CMS release will be included in that release.
+
+## PHP support commitments
+
+The Silverstripe CMS release cycle is built around these assumptions:
+
+- New PHP releases are shipped at the end of November on an annual basis.
+- PHP releases are in full support for two years, followed by two years of support for critical security fixes only.
+
+At launch, a Silverstripe CMS major release supports all PHP versions in full support. PHP versions in limited support are not supported at launch by the new Silverstripe CMS major version series.
```
ChloeSartorelli:
```suggestion
At launch, a Silverstripe CMS major release supports all PHP versions in full support. PHP versions in limited support are not supported by the new Silverstripe CMS major version series.
```
Just removing the "at launch" bit - we wouldn't add backwards support after launch i.e. we won't add PHP 7.4 support for CMS 6.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,191 @@
...
+- **Cadence:** Every six months within an active Silverstripe CMS major version series. Minor releases are targeted for April and October each year
+- **Pre-release schedule:**
+  - **Alpha**: Typically not released
+  - **Beta**: Approximately 6 weeks prior to the anticipated Silverstripe CMS minor release. At this point, no new features or new public API should be added
+  - **RC**: Approximately 2 weeks prior to the Silverstripe CMS release. At this point, only bug fixes should be added
+- **Documentation:** A changelog is published listing significant new features
+
+> [!NOTE]
+>
+> While each Silverstripe CMS release usually only gets one of each type of pre-stable release, additional pre-stable releases may be tagged if required.
```
ChloeSartorelli: This should be moved to be above the "Silverstripe CMS major release" heading, since it doesn't only apply to minor releases.

Also, "tagged" should be replaced with "performed" per previous discussion about avoiding that word.

en/12_Project_Governance/05_Release_policy.md
```diff
@@ -0,0 +1,191 @@
...
+>
+> Patch releases for individual modules such as `silverstripe/framework` are not considered a "Silverstripe CMS release", however they are an important part of meeting our support commitments.
```
ChloeSartorelli:
```suggestion
```
This is explicitly mentioned below in the dedicated patch releases section.

GuySartorelli (changes requested):
Looks great! A few suggestions below but only a handful of them really affect the meaning of the document.

GuySartorelli:
Once these changes are made, it'll be ready for Jenn to approve.
Things for Jenn to be aware of:
1. This is intended to blend the [major release policy](https://docs.silverstripe.org/en/6/project_governance/major_release_policy/) and [minor release policy](https://docs.silverstripe.org/en/6/project_governance/minor_release_policy/) together.
1. It's not intended to actually change anything about our commitments, or about how we communicate those commitments, except that:
    1. The minor release policy says "Once a release candidate is tagged, only critical impact bug fixes can be added to the release" - this has been changed here to remove the word "critical" to better match reality.
1. Wording has been updated to better clarify exactly what our commitments are. The _interpretation_ of those commitments should not have changed, though.

GuySartorelli:
Just realised - this (and the redirections) should target `5.4` since CMS 5 is still in support and we don't want two separate active sets of release policies.

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Updated

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Updated

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Updated

en/12_Project_Governance/05_Release_policy.md
emteknetnz: We defined Silverstripe CMS version as x.y, and we defined a release as corresponding to a version, we should refer to it as x.y that rather than x.y.z?

I can see the argument, because we have a 6.1.0 changelog, not a 6.1 changelog.  Though our support timeline we mention the 6.1 version

I think it's going to be less confusing for people to call it a 6.1 release

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Updated

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Updated

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Updated

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Discuss in https://github.com/silverstripe/developer-docs/pull/753#discussion_r2244183910

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Moved to blurb

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Updated

en/12_Project_Governance/05_Release_policy.md
ChloeSartorelli: > - **Silverstripe CMS version**: The version number corresponding to a Silverstripe CMS release, e.g. "6.1"
> - **Silverstripe CMS release**: A combined release of commercially supported modules which together make up Silverstripe CMS. Also known as a "stable release"

The "Silverstripe CMS version" is x.y e.g. "6.1" - the release doesn't have that definition.

The release is "a combined release ... also known as a stable release" and specifically in the document we use it to mean the first stable tagged release for a Silverstripe CMS version, e.g. major or minor release. Those would identified by their full tag name if someone asked us to point out out to them (e.g. 5.4.0 was a minor _release_ and 5.4 is the minor _version_ of that release - the version may have additional patch _releases_ e.g. 5.4.1, 5.4.2)

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Updated

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Updated to use "prior to the stable release". I've done this for all the pre-releases for both major in minor. "stable release" makes sense if context

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Updated

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Removed

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Yup

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Updated

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Updated

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Updated

en/12_Project_Governance/05_Release_policy.md
emteknetnz: OK will go with x.y.z

en/12_Project_Governance/05_Release_policy.md
emteknetnz: Updated

emteknetnz:
Closed in favour of https://github.com/silverstripe/developer-docs/pull/788

Issue #689, opened 2025-02-12
Review CMS 6 changelog and docs prior to stable release


GuySartorelli:
We should give the CMS 6 changelog a thorough review before we produce the stable release

## Acceptance Criteria

- [x] Review structure of changelog and move things around as necessary to provide a clear flow
  - e.g. maybe have non-breaking changes first
  - e.g. maybe group the validation changes together
- ~~Ensure the automated list of changes is accurate~~ already done as part of reviewing changelogs at release time
- [x] Ensure instructions are clear
- [x] Remove anything that's unnecessary
- [x] Make sure we've deleted all the CMS 5 changelogs
- [x] Final check for renamed classes e.g. `ViewableData` should be `ModelData`
- [x] Double check supported modules list is correct
- [x] Double check fixed dependencies list is correct
- [x] Somewhere in the changelog it mentions that https://www.npmjs.com/package/@silverstripe/webpack-config got a new major version and should be updated if it's being used.
- ~~The "GraphQL removed from the CMS" section mentions that several webpack externals were removed - modules that were relying on them will need to include the dependencies directly instead if they want to keep using GraphQL in the CMS.~~ There's a separate section for that instead.
- [x] Mark webauthn + hybridsessions modules as being CMS 6 compatible but not supported
- [x] Note that the old way to add a history tab (using GraphQL) no longer works - just add `HistoryViewerField` to a new tab instead.

## PRs

- https://github.com/silverstripe/developer-docs/pull/757
- https://github.com/silverstripe/developer-docs/pull/760
- https://github.com/silverstripe/developer-docs/pull/761

PR #757, opened 2025-06-04
Restructure CMS 6.0 changelog

GuySartorelli:
Two commits intentionally - do not squash.

## Issue

- https://github.com/silverstripe/developer-docs/issues/689


en/08_Changelogs/6.0.0.md
```diff
@@ -4,61 +4,63 @@ title: 6.0.0 (unreleased)
 
 # 6.0.0 (unreleased)
 
+Silverstripe CMS 6.0.0 is a major release. That means it contains many breaking changes and dependency upgrades. We tried our best to minimise upgrades woes, but you should allocate additional time to account for regression testing.
+
+This changelog provides a list of changes between Silverstripe CMS 5.4 and 6.0.
```
ChloeSartorelli, line 9: Similar to the start of the [5.0.0 changelog](https://docs.silverstripe.org/en/5/changelogs/5.0.0/)

en/08_Changelogs/6.0.0.md
```diff
@@ -4,61 +4,63 @@ title: 6.0.0 (unreleased)
...
+
 ## Overview
 
 - [Change to commercially supported modules](#changes-to-support)
   - [Modules losing commercial support](#modules-losing-support)
+  - [`silverstripe/campaign-admin` no longer part of `silverstripe/recipe-cms`](#campaign-admin)
   - [New supported modules](#new-supported-modules)
+  - [New default front-end theme](#theme)
+  - [TinyMCE is in a separate module now](#tinymce)
```
ChloeSartorelli: Moved things that are directly related to changes in supported modules into this section. If you need to add or exclude deps in `composer.json`, it goes here.

en/08_Changelogs/6.0.0.md
```diff
@@ -4,61 +4,63 @@ title: 6.0.0 (unreleased)
...
+- [Dependency changes](#dependency-changes)
+  - [PHP version support](#php-version-support)
+  - [MySQL 5 no longer supported](#mysql-5-support)
+  - [`intervention/image` has been upgraded from v2 to v3](#intervention-image)
+  - [symfony dependencies have been upgraded from v6 to v7](#symfony)
+  - [JavaScript dependencies](#javascript-dependencies)
```
ChloeSartorelli: Moved dependency changes to be directly below the supported modules section, because then we have all the `composer.json` stuff packed together.

en/08_Changelogs/6.0.0.md
```diff
@@ -4,61 +4,63 @@ title: 6.0.0 (unreleased)
...
 
 - [Change to commercially supported modules](#changes-to-support)
   - [Modules losing commercial support](#modules-losing-support)
+  - [`silverstripe/campaign-admin` no longer part of `silverstripe/recipe-cms`](#campaign-admin)
   - [New supported modules](#new-supported-modules)
+  - [New default front-end theme](#theme)
+  - [TinyMCE is in a separate module now](#tinymce)
+- [Dependency changes](#dependency-changes)
+  - [PHP version support](#php-version-support)
+  - [MySQL 5 no longer supported](#mysql-5-support)
```
ChloeSartorelli, line 21: These feel like they belong in dependency changes since they're stuff you deal with before you start diving into actual code

en/08_Changelogs/6.0.0.md
```diff
@@ -4,61 +4,63 @@ title: 6.0.0 (unreleased)
...
+  - [Changes to the templating/view layer](#view-layer)
+  - [Changes to `LeftAndMain` and its subclasses](#leftandmain-refactor)
   - [`FormField` classes now use `FieldValidator` for validation](#formfield-validation)
   - [`CMSMain` search filter changes](#cmsmain-search-filter)
   - [Most extension hook methods are now protected](#hooks-protected)
-  - [Changes to some extension hook names](#hooks-renamed)
   - [`FormField::Value()` split into two methods](#formfield-value)
   - [`Controller::has_curr()` removed](#controller-has-curr)
   - [Strict typing for `Factory` implementations](#factory-strict-typing)
-  - [Elemental `TopPage` class names changed](#elemental-top-page)
```
ChloeSartorelli: 1. Removed "Elemental `TopPage` class names changed" section and put that content straight into "Many renamed classes".
1. Moved "changes to some extension hook names" right under "Many renamed classes" so the renaming stuff is grouped
1. Moved "Changes to the templating/view layer" and "Changes to `LeftAndMain` and its subclasses" into the API changes section because they're _more_ about changes to the API than they are about functional enhancements

en/08_Changelogs/6.0.0.md
```diff
@@ -4,61 +4,63 @@ title: 6.0.0 (unreleased)
...
+  - [Changes to scaffolded form fields](#scaffolded-fields)
+  - [`SiteTree` uses form field scaffolding](#sitetree-scaffolding)
+  - [MySQL now defaults to utf8mb4](#mysql-utf8mb4)
+  - [Run `CanonicalURLMiddleware` in all environments by default](#url-middleware)
+  - [Session and authentication cookie changes](#session-cookie-changes)
+  - [`DBDecimal` default value](#dbdecimal-default-value)
+  - [`RedirectorPage` validation](#redirectorpage-validation)
+  - [Update JS MIME type, remove `type` in `<script>` tags](#js-mime-type-update)
+  - [`getSchemaDataDefaults()` now includes attributes](#formfield-schema-data)
+  - [Remember me token rotation](#remember-me-token-rotation)
```
ChloeSartorelli: 1. Renamed "Other changes" to "Breaking changes" so it was clearer what sort of thing is in here. I've also added a paragraph to the start of the section indicating the scope of the section
2. Moved to be inbetween "Features and enhancements" and "Bug fixes" mostly to get it out from under "API changes". I want "API changes" to be right above the "Full list of removed and changed API" since that's API with API.
3. "`$CurrentPageURL` template support removed" section removed - it's just a bullet point under "API changes > general changes" now.
4. Moved "Changes to scaffolded form fields" and "`SiteTree` uses form field scaffolding" into here because these are kinda miscellaneous-feeling breaking changes.
5. Moved "Run `CanonicalURLMiddleware` in all environments by default" here because it's not _really_ an enhancement, but it _is_ a breaking change

en/08_Changelogs/6.0.0.md
```diff
@@ -4,61 +4,63 @@ title: 6.0.0 (unreleased)
...
+  - [MySQL 5 no longer supported](#mysql-5-support)
+  - [`intervention/image` has been upgraded from v2 to v3](#intervention-image)
+  - [symfony dependencies have been upgraded from v6 to v7](#symfony)
+  - [JavaScript dependencies](#javascript-dependencies)
 - [Upgrade tools](#upgrade-tools)
 - [Features and enhancements](#features-and-enhancements)
   - [Validation added to DBFields](#dbfield-validation)
-  - [Changes to `sake`, `BuildTask`, CLI interaction in general](#cli-changes)
+  - [Changes to password validation](#password-validation)
+  - [Changes to `sake`, `BuildTask`, and CLI interaction in general](#cli-changes)
```
ChloeSartorelli: 1. Moved validation next to validation
2. Added "and" to "Changes to `sake`, `BuildTask`, and CLI interaction in general" to match actual heading

en/08_Changelogs/6.0.0.md
```diff
@@ -4,61 +4,63 @@ title: 6.0.0 (unreleased)
...
   - [Read-only replica database support](#db-read-only-replicas)
+  - [Improvements to template functionality](#improvements-to-template-functionality)
```
ChloeSartorelli: Separated this out from "Changes to the templating/view layer" because these are enhancements that aren't _really_ related to the whole separation of view vs model.

en/08_Changelogs/6.0.0.md
```diff
@@ -496,12 +717,83 @@ What's more, the `PHPFilesAdapter` used as a fallback wasn't enabled for CLI, wh
 
 We've made the following changes to resolve this problem:
 
+- There is no default in-memory cache used by `DefaultCacheFactory` (APCu cache used to be used by default).
 - The `PHPFilesAdapter` will only be used if it's available for both the webserver *and* CLI. Otherwise, `FilesystemAdapter` will be used for both.
-- There is no default in-memory cache used by `DefaultCacheFactory`.
```
ChloeSartorelli, line 500: Added emphasis to the main breaking change by moving to the top and saying what the default used to be

en/08_Changelogs/6.0.0.md
```diff
@@ -4,61 +4,63 @@ title: 6.0.0 (unreleased)
 
 # 6.0.0 (unreleased)
 
+Silverstripe CMS 6.0.0 is a major release. That means it contains many breaking changes and dependency upgrades. We tried our best to minimise upgrades woes, but you should allocate additional time to account for regression testing.
+
+This changelog provides a list of changes between Silverstripe CMS 5.4 and 6.0.
+
 ## Overview
 
 - [Change to commercially supported modules](#changes-to-support)
```
```suggestion
- [Changes to commercially supported modules](#changes-to-support)
```

en/08_Changelogs/6.0.0.md
emteknetnz: I don't think "Breaking changes" is helpful here, because we have "API changes" as well which are also breaking. It implies that the API changes are not breaking, when they are.

I think "Other changes" is better.

emteknetnz (changes requested):
Seems like the PR is mostly about just reordering existing content without changing much of it. I'm basing this of the number of lines added and removed being fairly close when accounting for the CMS 5 changelogs being removed.

I feel the changelog in this PR reads somewhat negatively, because at the top it has several "things that are going away" e.g. modules loosing commercial support.

I think the changelog should lead with "new features" which is what people will be most looking forward to seeing. As much as we go on about "majors are for breaking changes", people are always asking about "what new features are in CMS 6", so we may as well arrange things to meet other peoples expectations.

Here's an alternative arrangement:

Note that within "Change to commercially supported modules" I've reoorded things so that the "new" stuff is up the top and the "going away" stuff is down the bottom

```
- Features and enhancements
    - Validation added to DBFields
    - Changes to password validation
    - Changes to sake, BuildTask, and CLI interaction in general
    - Read-only replica database support
    - Improvements to template functionality
    - Performance improvements for site tree rendering
    - Changes to default cache adapters
    - Filter within a range with searchable_fields
    - Status flags in the CMS
    - New Versioned methods
    - Other new features
- Other changes
    - Changes to scaffolded form fields
    - SiteTree uses form field scaffolding
    - MySQL now defaults to utf8mb4
    - Run CanonicalURLMiddleware in all environments by default
    - Session and authentication cookie changes
    - DBDecimal default value
    - RedirectorPage validation
    - Update JS MIME type, remove type in <script> tags
    - getSchemaDataDefaults() now includes attributes
    - Remember me token rotation
- Change to commercially supported modules
    - New supported modules
    - New default front-end theme
    - TinyMCE is in a separate module now
    - silverstripe/campaign-admin no longer part of silverstripe/recipe-cms
    - Modules losing commercial support
- Dependency changes
    - PHP version support
    - MySQL 5 no longer supported
    - intervention/image has been upgraded from v2 to v3
    - symfony dependencies have been upgraded from v6 to v7
    - JavaScript dependencies
- Upgrade tools
- Bug fixes
- API changes
    - Many renamed classes
    - Changes to some extension hook names
    - GraphQL removed from the CMS
    - Changes to the templating/view layer
    - Changes to LeftAndMain and its subclasses
    - FormField classes now use FieldValidator for validation
    - CMSMain search filter changes
    - Most extension hook methods are now protected
    - FormField::Value() split into two methods
    - Controller::has_curr() removed
    - Strict typing for Factory implementations
    - List interface changes
    - General changes
- Full list of removed and changed API (by module, alphabetically)
```

en/08_Changelogs/6.0.0.md
ChloeSartorelli: I added a blurb to the section to help alleviate that confusion - but how about "Other breaking changes"? That's what we used in CMS 5.

I don't like just "Other changes" because it sounds like the stuff in there is less important than the rest of the changelog.

GuySartorelli:
See https://github.com/silverstripe/developer-docs/pull/757#discussion_r2127511948 in response to "breaking changes" vs "other changes"

> Seems like the PR is mostly about just reordering existing content without changing much of it.

Yes, I figured restructuring should be its own PR, and updating content should be separate, or else it would be a nightmare to review.

> I feel the changelog in this PR reads somewhat negatively, because at the top it has several "things that are going away" e.g. modules loosing commercial support.
> I think the changelog should lead with "new features" which is what people will be most looking forward to seeing. As much as we go on about "majors are for breaking changes", people are always asking about "what new features are in CMS 6", so we may as well arrange things to meet other peoples expectations.

I would prefer to keep the changes to supported modules at the top, because:
1. that will be directly below the "modules included in this release" which keeps things logically grouped
2. that and the other dependency changes are the sort of thing developers should ideally look at the earliest point of scoping an upgrade. "What's changed in my dependencies, and what additional things outside of my own code do I need to worry about?"

I don't think people scoping and performing upgrades from CMS 5 to CMS 6 will care that much about new features - they are likely to want to perform a like-for-like upgrade which means they need to know about breaking changes and information pertinent to their upgrade primarily.

I also don't think there's anything "negative" about the ordering of the changelog. It's known that a major release includes breaking changes. Besides which, the changelog isn't a marketing tool, it's an upgrade tool. IMO the order should be whatever's most helpful to people using the changelog to upgrade their projects, but within a logical structure (e.g. don't put features and enhancements _last_ because the API stuff should ideally come just before the full list of API changes, which groups nicely with the commits list).

> Note that within "Change to commercially supported modules" I've reoorded things so that the "new" stuff is up the top and the "going away" stuff is down the bottom

Similarly to my thoughts about the order as a whole - I think the order here should be "what is most useful for upgrades" which ideally puts the changes to existing support first.

en/08_Changelogs/6.0.0.md
emteknetnz: We did all these things to make things better right? So they're basically all enhancements, it's just they happen to be breaking

Maybe just list them under the "Features and enhancements" section?

emteknetnz:
OK I guess if we assume the target audience is developers scoping a CMS 6 upgrade then that order makes sense

GuySartorelli:
Yeah I think we have to assume that - the blog post is for a less technical audience, but the changelog kinda has to be for developers or else we will have to start having discussions like "is that too technical a detail? Will non-technical people understand this?" which starts to diminish the usefulness of the changelog.

en/08_Changelogs/6.0.0.md
ChloeSartorelli: Yeah that makes sense I guess. I was originally trying to keep some separation between the not-so-breaking things and the "these are just plain breaking changes" things. But any distinction here will have exceptions no matter how you try to cut it. I have no qualms with combining them.

en/08_Changelogs/6.0.0.md
ChloeSartorelli: Done, good catch

GuySartorelli:
I've rearranged stuff within the now-combined features and enhancements section in a way that I think makes sense - with things more likely to affect upgrades higher in the list, and things less likely to affect upgrades lower in the list (but also with things that are thematically similar close to each other).

GuySartorelli:
I have now also:

1. Added missing links to the following sections:
    - "Support for PHPUnit 11"
    - "`getCMSValidator` method no longer supported"
1. Moved "New `SingleRecordAdmin` class and changes to `SiteConfig`" out of the "Changes to `LeftAndMain` and its subclasses" section and into "Features and enhancements" (felt like it deserved its own section, and that seems like the best place for it)
1. Moved "Most extension hook methods are now protected" up to be next to "Changes to some extension hook names" since they're thematically linked
1. Added links between the "`FormField` classes now use `FieldValidator` for validation" and "validation added to `DBFields`" sections - anyone dealing with validation in the CMS should be aware that these are connected - though their positions in the list I think are appropriate as one is a new feature while the other is kinda an API change for an existing feature. Happy to move "`FormField` classes now use `FieldValidator` for validation" into features and enhancements though if you prefer.

emteknetnz (approved):

Issue #762, opened 2025-06-05
Review performance docs section


emteknetnz:
Review the [performance](https://docs.silverstripe.org/en/5/developer_guides/performance/) section in developer docs to ensure it's accurate and make improvements to its informatio

## Acceptance criteria
- [x] A PR to upgrade content is raised
- [x] A PR to tidy up writing style is raised

## PRs
- https://github.com/silverstripe/developer-docs/pull/764
- https://github.com/silverstripe/developer-docs/pull/766
- https://github.com/silverstripe/developer-docs/pull/769

GuySartorelli:
PR merged, reassigning to Steve for next AC

GuySartorelli:
@emteknetnz PR is still in draft

emteknetnz:
Marked as ready for review

GuySartorelli:
@emteknetnz I didn't click on to the fact that your first PR changed the URL segments of some pages - please make sure you update links to those pages (e.g. `/developer_guides/performance/caching/` is referenced a few times)

GuySartorelli:
PRs merged. Reassigning to Steve for https://github.com/silverstripe/developer-docs/issues/762#issuecomment-2978861582

emteknetnz:
Have create new PR to fix links

GuySartorelli:
PRs merged

PR #766, opened 2025-06-16
DOC Tidy up writing style of performance section

emteknetnz:
Issue https://github.com/silverstripe/developer-docs/issues/762

### DIFF
--- a/en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
+++ b/en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
@@ -6,33 +6,31 @@ icon: tachometer-alt
 
 # Partial template caching
 
-Partial template caching is a feature that allows caching of rendered portions of a template so that it does not need to be recomputed on each request.
+Partial template caching allows you to cache rendered portions of a template, so that it does not need to be recomputed on each request.
 
-The [Partial template caching](../templates/partial_template_caching) page in the templates section talks in-depth about how to use partial template caching.
+See [Partial template caching](../templates/partial_template_caching) in the templates section for detailed usage instructions.
 
-This page will discuss some of the performance considerations.
+This page discusses some of the performance considerations.
 
 ## Cache block conditionals
 
-Use conditions whenever possible. The cache tag supports defining conditions via either `if` or `unless` keyword.
-Those are optional, however are highly recommended.
+Use conditions whenever possible. The cache tag supports optional `if` or `unless` conditions, which are highly recommended.
 
 > [!WARNING]
-> Avoid performing heavy computations in conditionals, as they are evaluated for every template rendering.
+> Avoid heavy computations in conditionals, as they are evaluated every time the template needs to be rendered.
 
 If you cache without conditions:
 
-- your cache backend will always be queried for the cache block (on every template render)
-- your cache may be cluttered with heaps of redundant and useless data (especially the default filesystem backend)
+- The cache backend will be queried for the cache block on every template render even if it's not needed.
+- Your cache may be cluttered with redundant data, especially with the default filesystem backend.
 
-As an example, if you use `$DataObject->ID` as a key for the block, consider adding a condition that ID is greater than zero:
+For example, if you use `$DataObject->ID` as a key for the block, consider adding a condition that the ID is greater than zero:
 
 ```ss
 <% cached $MenuItem.ID if $MenuItem.ID > 0 %>
 ```
 
-To cache the contents of a page for all anonymous users, but dynamically calculate the contents for logged in members,
- use something like:
+To cache content for anonymous users, but dynamically calculate it for logged-in members, use:
 
 ```ss
 <% cached unless $CurrentUser %>
@@ -40,24 +38,23 @@ To cache the contents of a page for all anonymous users, but dynamically calcula
 
 ## Cache invalidation strategies
 
-### Aggregates based invalidation {#aggregates}
+### Aggregates-based invalidation {#aggregates}
 
 Sometimes you may want to invalidate cache when any object in a set changes, or when objects in a relationship
 change. To do this, you may use [DataList](api:SilverStripe\ORM\DataList) aggregate methods (which we call Aggregates).
 These perform SQL aggregate queries on sets of [DataObject](api:SilverStripe\ORM\DataObject)s.
 
-Here are some useful methods of the [DataList](api:SilverStripe\ORM\DataList) class:
+Some useful methods of the [`DataList`](api:SilverStripe\ORM\DataList) class for this purpose include:
 
-- `int count()` : Return the number of items in this DataList
-- `mixed max(string $fieldName)` : Return the maximum value of the given field in this DataList
-- `mixed min(string $fieldName)` : Return the minimum value of the given field in this DataList
-- `mixed avg(string $fieldName)` : Return the average value of the given field in this DataList
-- `mixed sum(string $fieldName)` : Return the sum of the values of the given field in this DataList
+- `int count()`: Returns the number of items in this DataList.
+- `mixed max(string $fieldName)`: Returns the maximum value of the given field in this DataList.
+- `mixed min(string $fieldName)`: Returns the minimum value of the given field in this DataList.
+- `mixed avg(string $fieldName)`: Returns the average value of the given field in this DataList.
+- `mixed sum(string $fieldName)`: Returns the sum of the values of the given field in this DataList.
 
-To construct a `DataList` over a `DataObject`, we have a global template variable called `$List`.
+To construct a `DataList` over a `DataObject`, you can use the global template variable `$List`.
 
-For example, if we have a menu, we may want that menu to update whenever *any* page is edited, but would like to cache it
-otherwise. By using aggregates, we do that like this:
+For example, to cache a menu but invalidate that cache whenever any page is edited:
 
 ```ss
 <% cached
@@ -70,21 +67,16 @@ otherwise. By using aggregates, we do that like this:
 > [!NOTE]
 > The use of the fully qualified classname is necessary.
 >
-> The use of both `.max('LastEdited')` and `.count()` makes sure we check for any object
-> edited or deleted since the cache was last built.
+> Using both `.max('LastEdited')` and `.count()` ensures we check for any object edited or deleted since the cache was last built.
 
-The cache for this will update whenever a page is added, removed or edited.
+In this example, the cache will update whenever a page is added, removed or edited.
 
 > [!WARNING]
-> Be careful using aggregates. Remember that the database is usually one of the performance bottlenecks.
-> Keep in mind that every key of every cached block is recalculated for every template render, regardless of caching
-> result. Aggregating SQL queries usually produce more load on the database than simple select queries,
-> especially if you query records by Primary Key or join tables using database indices properly.
+> Be careful when using aggregates. The database is often a performance bottleneck. Keep in mind that every key of every cached block is recalculated for every template render, regardless of caching result. Aggregating SQL queries can put more load on the database than simple select queries, especially if you query records by Primary Key or join tables using database indices properly.
 >
-> Sometimes it may be cheaper to not cache altogether, rather than cache a block using a bunch of heavy aggregating SQL
-> queries.
+> Consider whether not caching at all is cheaper than caching a block using heavy aggregating SQL queries.
 >
-> Let us consider two versions:
+> Consider the following two versions:
 >
 > ```ss
 > # Version 1 (bad)
@@ -112,20 +104,17 @@ The cache for this will update whenever a page is added, removed or edited.
 > [controller method](../templates/partial_template_caching/#cache-key-calculated-in-controller).
 > [Object Caching](../templates/caching/#object-caching) only works for single variables and not for chained expressions.
 
-### Time based invalidation
+### Time-based invalidation
 
-In some situations it's more important to be fast than to always be showing the latest data. By constructing the cache
-key to invalidate less often than the data updates you can ensure rendering time is constant no matter how often the
-data updates.
+In some situations it's more important to be fast than to always be showing the latest data. One way to achieve this is to construct the cache key to invalidate less often than the data updates. This ensures constant rendering time, regardless of how often the data updates.
 
-For instance, if we show some blog statistics, but are happy having them be slightly stale, we could do
+For instance, if you show blog statistics, but are happy with slightly stale data:
 
 ```ss
 <% cached 'blogstatistics', $Blog.ID %>
 ```
 
-which will invalidate after the cache lifetime expires. If you need more control than that (cache lifetime is
-[configurable only on a site-wide basis](/developer_guides/templates/partial_template_caching/#cache-storage)), you could add a special function to your controller:
+This doesn't require any additional SQL queries and will invalidate after the cache lifetime expires. Cache lifetime is [configurable only on a site-wide basis](/developer_guides/templates/partial_template_caching/#cache-storage). For more control, add a function to your controller:
 
 ```php
 namespace App\Model;
@@ -144,7 +133,7 @@ class MyObject extends DataObject
 }
 ```
 
-and then use it in the cache key
+Then use it in the cache key:
 
 ```ss
 <% cached 'blogstatistics', $Blog.ID, $BlogStatisticsCount %>
@@ -152,14 +141,9 @@ and then use it in the cache key
 
 ## Cache backend
 
-The template engine uses [Injector](../extending/injector) service `Psr\SimpleCache\CacheInterface.cacheblock` as
-caching backend. The default definition of that service is very conservative and relies on the server filesystem.
-This is the most common denominator for most of the applications out there. However,
-this is not the most robust nor performant cache implementation. If you have a better solution
-available on your platform, you should consider tuning that setting for your application.
-All you need to do to swap the cache backend for partial template cache blocks is to redefine this service for the Injector.
+The template engine uses [Injector](../extending/injector) service `Psr\SimpleCache\CacheInterface.cacheblock` as its caching backend. The default definition of that service is conservative, and relies on the server filesystem. While this is the most common denominator, it's not the most robust or performant implementation. If you have a better solution available on your platform, you should consider tuning this setting. To swap the cache backend for partial template cache blocks, redefine this service for the Injector.
 
-Here's an example of how it could be done:
+For example:
 
 > [!NOTE]
 > For the below example to work it is necessary to have the Injector service `App\Cache\Service.memcached` defined somewhere in the configs.
@@ -176,6 +160,4 @@ SilverStripe\Core\Injector\Injector:
 ```
 
 > [!WARNING]
-> The default filesystem cache backend does not support auto cleanup of the residual files with expired cache records.
-> If your project relies on Template Caching heavily (e.g. thousands of cache records daily), you may want to keep an eye on the
-> filesystem storage. Sooner or later its capacity may be exhausted.
+> The default filesystem cache backend does not support auto cleanup of the residual files with expired cache records. If your project relies on Template Caching heavily (e.g. thousands of cache records daily), you may want to monitor the filesystem storage. If you don't, its capacity may eventually be exhausted.
--- a/en/02_Developer_Guides/08_Performance/01_Object_Caching.md
+++ b/en/02_Developer_Guides/08_Performance/01_Object_Caching.md
@@ -7,9 +7,12 @@ icon: tachometer-alt
 
 ## Overview
 
-The framework uses caches to store infrequently changing values.
-By default, the storage mechanism chooses the most performant adapter available
-(PHP opcache, APC, or filesystem). Other cache backends can be configured.
+Object caching stores infrequently changing values to improve performance.
+
+By default, the storage mechanism chooses the most performant filesystem adapter available
+(PHP opcache, or filesystem) but does not use in-memory cache unless you configure it. Other cache backends can be configured.
+
+Object caching is useful for situations where fetching or generating data can take a long time, such as expensive database queries or API calls to external services.
 
 The most common caches are manifests of various resources:
 
@@ -32,11 +35,10 @@ For long-lived server instances, this can become a capacity issue over time - se
 
 ## Configuration
 
-We are using the [PSR-16](https://www.php-fig.org/psr/psr-16/) standard ("SimpleCache")
+We use the [PSR-16](https://www.php-fig.org/psr/psr-16/) standard ("SimpleCache")
 for caching, through the [symfony/cache](https://symfony.com/doc/current/components/cache.html) library.
 
-Note that this library describes usage of [PSR-6](https://www.php-fig.org/psr/psr-6/) by default,
-though Silverstripe wraps these in a PSR-16 interface using the [Psr16Cache](https://github.com/symfony/cache/blob/6.1/Psr16Cache.php) class.
+Silverstripe provides a PSR-16 interface via the [`Psr16Cache`](https://github.com/symfony/cache/blob/6.1/Psr16Cache.php) class which wraps `symfony/cache`'s [PSR-6](https://www.php-fig.org/psr/psr-6/) compliant adapters.
 
 Cache objects are configured via YAML
 and Silverstripe CMS's [dependency injection](/developer_guides/extending/injector) system.
@@ -56,7 +58,7 @@ SilverStripe\Core\Injector\Injector:
 > Please read the [versioned cache segmentation](#versioned-cache-segmentation) section for more information.
 
 Cache objects are instantiated through a [CacheFactory](api:SilverStripe\Core\Cache\CacheFactory),
-which determines which cache adapter is used (see [cache adapters](cache-adapters) for details).
+which determines which cache adapter is used (see [cache adapters](cache_adapters) for details).
 This factory allows us you to globally define an adapter for all cache instances.
 
 ```php
--- a/en/02_Developer_Guides/08_Performance/02_Cache_Adapters.md
+++ b/en/02_Developer_Guides/08_Performance/02_Cache_Adapters.md
@@ -6,20 +6,30 @@ icon: tachometer-alt
 
 # Cache adapters
 
-We use the `smyfony/cache` library which supports various [cache adapters](https://github.com/symfony/cache/tree/6.1/Adapter).
+Cache adapters determine how cached data is stored and retrieved. Choosing the right adapter can significantly impact your application's performance.
+
+We use the `symfony/cache` library which supports various [cache adapters](https://github.com/symfony/cache/tree/6.1/Adapter).
 
 Silverstripe CMS tries to provide a sensible default cache implementation for your system
 through the [`DefaultCacheFactory`](api:SilverStripe\Core\Cache\DefaultCacheFactory) implementation.
 
+The `DefaultCacheFactory` chooses:
+
 - `PhpFilesAdapter` (PHP with [opcache](https://php.net/manual/en/book.opcache.php) enabled).
-    This cache has relatively low [memory defaults](https://php.net/manual/en/opcache.configuration.php#ini.opcache.memory-consumption).
-    We recommend increasing it for large applications, or enabling the
-    [`file_cache` fallback](https://php.net/manual/en/opcache.configuration.php#ini.opcache.file-cache).
-    You must have [`opcache.enable_cli`](https://www.php.net/manual/en/opcache.configuration.php#ini.opcache.enable-cli) set to `true`
-    to use this cache adapter. This is so that your cache is shared between CLI and the webserver.
+  This cache has relatively low [memory defaults](https://php.net/manual/en/opcache.configuration.php#ini.opcache.memory-consumption).
+  We recommend increasing it for large applications, or enabling the
+  [`file_cache` fallback](https://php.net/manual/en/opcache.configuration.php#ini.opcache.file-cache).
+  You must have [`opcache.enable_cli`](https://www.php.net/manual/en/opcache.configuration.php#ini.opcache.enable-cli) set to `true`
+  to use this cache adapter. This is so that your cache is shared between CLI and the webserver.
 - `FilesystemAdapter` if the above isn't available
 
 ## Adding an in-memory cache adapter
+
+The in-memory cache adapters provided by default are:
+
+- Memcached
+- Redis
+- APCu
 
 The cache adapter needs to be available before configuration has been loaded, so we use an environment variable to determine which class will be used to instantiate the in-memory cache adapter. The class must be referenced using its Fully Qualified Class Name (including namespace), and must be an instance of [`InMemoryCacheFactory`](api:SilverStripe\Core\Cache\InMemoryCacheFactory).
 
--- a/en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
+++ b/en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
@@ -6,10 +6,9 @@ icon: tachometer-alt
 
 # HTTP cache headers
 
-By default, Silverstripe CMS sends headers which signal to HTTP caches
-that the response should be not considered cacheable.
-HTTP caches can either be intermediary caches (e.g. CDNs and proxies), or clients (e.g. browsers).
-The cache headers sent are `Cache-Control: no-cache, must-revalidate`;
+HTTP cache headers instruct HTTP caches (e.g. browsers, CDNs, and proxies) how to store and serve responses. Proper configuration can significantly improve website performance.
+
+By default, Silverstripe CMS sends headers that prevent caching: `Cache-Control: no-cache, must-revalidate`.
 
 > [!WARNING]
 > HTTP caching can be a great way to speed up your website, but needs to be properly applied.
@@ -19,10 +18,7 @@ The cache headers sent are `Cache-Control: no-cache, must-revalidate`;
 
 ### Overview
 
-In order to support developers in making safe choices around HTTP caching,
-we're using a `HTTPCacheControlMiddleware` class to control if a response
-should be considered public or private. This is an abstraction on the
-`HTTPResponse->addHeader()` low-level API.
+Silverstripe CMS provides `HTTPCacheControlMiddleware` to help developers manage HTTP cache headers safely. This class helps prevent accidental caching of private content such as draft content or member specific content. It's an abstraction on top of the `HTTPResponse->addHeader()` low-level API.
 
 The `HTTPCacheControlMiddleware` API makes it easier to express your caching preferences
 without running the risk of overriding essential core safety measures.
@@ -42,44 +38,37 @@ The [`HTTPCacheControlMiddleware`](api:SilverStripe\Control\Middleware\HTTPCache
 (deprecated) caching methods in the `HTTP` helper class.
 It comes with methods which let developers safely interact with the `Cache-Control` header.
 
-### DisableCache()
+### `publicCache()`
 
-Simple way to set cache control header to a non-cacheable state.
-Use this method over `privateCache()` if you are unsure about caching details.
-Takes precedence over unforced `enableCache()`, `privateCache()` or `publicCache()` calls.
+Sets cache control headers to a cacheable state. Will use the `public` directive and remove the `private` if present as that is a contradictory directive.
 
-Removes all state and replaces it with `no-cache, no-store, must-revalidate`. Although `no-store` is sufficient
-the others are added under [recommendation from Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#Examples)
+### `privateCache()`
 
-Does not set `private` directive, use `privateCache()` if this is explicitly required
-([details](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching#public_vs_private))
+Sets cache control headers to a mostly non-cacheable state and uses the `private` directive.
 
-### EnableCache()
-
-Simple way to set cache control header to a cacheable state.
-Use this method over `publicCache()` if you are unsure about caching details.
-
-Removes the `no-store` directive unless a `max-age` is set; other directives will remain in place.
-Use alongside `setMaxAge()` to activate caching.
-
-Does not set `public` directive. Usually, `setMaxAge()` is sufficient. Use `publicCache()` if this is explicitly required
-([details](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching#public_vs_private))
-
-### PrivateCache()
-
-Advanced way to set cache control header to a non-cacheable state.
 Indicates that the response is intended for a single user and must not be stored by a shared cache.
+
 A private cache (e.g. Web Browser) may store the response. Also removes `public` as this is a contradictory directive.
 
 > [!WARNING]
 > Private caching will cache at the client level, not at the a user level. While it stops shared servers (like CDNs) from caching your personalized data, it doesn't stop other people using the same browser from seeing each others cached content.
 > For example, if multiple family members share a computer and browser, one person might accidentally see cached information meant for another family member.
 
-### PublicCache()
+Use `disableCache()` instead if you are unsure about caching details. Takes precedence over unforced `enableCache()`, `privateCache()` or `publicCache()` calls.
 
-Advanced way to set cache control header to a cacheable state.
-Indicates that the response may be cached by any cache (e.g. CDNs, Proxies, Web browsers).
-Also removes `private` as this is a contradictory directive
+### `enableCache()`
+
+Sets cache control headers to a cacheable state, though without the `public` directive that `publicCache()` will set.
+
+Removes the `no-store` directive unless a `max-age` is set; other directives will remain in place. Use alongside `setMaxAge()` to activate caching.
+
+Unlike `publicCache()`, does not set `public` directive. Usually, `setMaxAge()` is sufficient. Use `publicCache()` if this is explicitly required.
+
+### `disableCache()`
+
+Sets cache control headers to a non-cacheable state.
+
+Removes all state and replaces it with `no-cache, no-store, must-revalidate`.
 
 ### Priority
 
@@ -253,3 +242,15 @@ SilverStripe\Control\Middleware\HTTPCacheControlMiddleware:
   defaultState: 'enabled'
   defaultForcingLevel: 0
 ```
+
+## Debugging HTTP cache headers
+
+You can use browser developer tools or command-line tools like `curl` to inspect HTTP cache headers and verify that they are configured correctly.
+
+For example, using `curl`:
+
+```bash
+curl -I https://example.com/mypage
+```
+
+The output will show the HTTP headers, including `Cache-Control`, `Vary`, and `Age`.
--- a/en/02_Developer_Guides/08_Performance/04_Static_Publishing.md
+++ b/en/02_Developer_Guides/08_Performance/04_Static_Publishing.md
@@ -7,13 +7,14 @@ summary: Export web pages as static HTML
 
 Static publishing is the process of exporting website pages as static HTML files which are then served directly. This means that page content does not need to be recomputed on each request. This can have the following benefits:
 
-- Static HTML files load much faster than dynamically generated pages as they don't require any server-side processing.
-- Static publishing reduces the load on the server, as it doesn't need to connect to the database or render templates for each request.
-- Static publishing can lower hosting costs because there is less need for server resources.
+- Faster load times: Static HTML files load much faster than dynamically generated pages.
+- Reduced server load: No database connection or template rendering is required for each request.
+- Lower hosting costs: Less server resources are needed to serve statically cached content.
 
 By publishing the page as HTML it's possible to run Silverstripe CMS from behind a corporate firewall, on a low performance server and a large amount of traffic without expensive hardware.
 
-On the flip side, static publishing is more complex than simply dynamically regenerating page content on every request because it requires a separate process to generate and manage the static files. Also, if static publishing happens on demand (i.e. on page save/publish) rather than on a queue, then speed of CMS operations may degrade.
+> [!WARNING]
+> Static publishing is more complex than regenerating page content every request because it requires a separate process to generate and manage the static files. Also, if static publishing happens on demand (i.e. on page save/publish) rather than on a queue, then speed of CMS operations may degrade.
 
 Static publishing is available through the [silverstripe/staticpublishqueue](https://github.com/silverstripe/silverstripe-staticpublishqueue) module.
 
--- a/en/02_Developer_Guides/08_Performance/06_ORM.md
+++ b/en/02_Developer_Guides/08_Performance/06_ORM.md
@@ -5,9 +5,20 @@ summary: Configuration and tips for improving ORM performance
 
 # ORM performance
 
+## Identifying ORM performance bottlenecks
+
+ORM performance issues can arise from various factors, including:
+
+- Inefficient database queries
+- Lack of proper indexes
+- Large datasets
+- Complex relationships
+
+You can use the [`showqueries` variable tool](/developer_guides/debugging/url_variable_tools/#database) on a dev environment to identify slow running database queries.
+
 ## Indexes
 
-You can define indexes for your ORM queries using the `$indexes` configuration property in your `DataObject` subclasses. See the [Indexes](/developer_guides/model/indexes) section for more information.
+Adding indexes to frequently queried fields can significantly improve performance. You can define indexes for your ORM queries using the `$indexes` configuration property in your `DataObject` subclasses. See the [Indexes](/developer_guides/model/indexes) section for more information.
 
 ### `TreeDropdownField` `SearchFilter` configuration {#treedropdownfield}
 
@@ -21,7 +32,7 @@ SilverStripe\Forms\TreeDropdownField:
 ```
 
 > [!TIP]
-> A very common use of `TreeDropdownField` is the "Insert Link" feature used by supported HTML editors. Setting this configuration to use another filter and adding an index on `Title` and `MenuTitle` for [`SiteTree`](api:SilverStripe\CMS\Model\SiteTree) can significantly improve performance.
+> A common use of `TreeDropdownField` is the "Insert Link" feature used by supported HTML editors. Setting this configuration to use another filter and adding an index on `Title` and `MenuTitle` for [`SiteTree`](api:SilverStripe\CMS\Model\SiteTree) can significantly improve performance.
 
 See [SearchFilter Modifiers](/developer_guides/model/searchfilters/) for more information about search filters.
 
@@ -81,6 +92,6 @@ SilverStripe\ORM\FieldType\DBPolymorphicForeignKey:
 If find the ORM is making needlessly inefficient SQL queries for a particular use case, then you can use raw SQL.
 
 > [!WARNING]
-> Using raw SQL queries can make your code less more difficult to maintain. Only use raw SQL when the ORM is a clear bottleneck.
+> Using raw SQL queries can make your code less more difficult to maintain. Only use raw SQL when the ORM is a clear bottleneck. Consider carefully if this approach is needed.
 
 Refer to the [raw SQL](/developer_guides/model/data_model_and_orm/#raw-sql) section for details about how to make raw SQL queries.
--- a/en/02_Developer_Guides/08_Performance/08_Read_only_database_replicas.md
+++ b/en/02_Developer_Guides/08_Performance/08_Read_only_database_replicas.md
@@ -5,7 +5,11 @@ summary: Using read-only database replicas to improve performance
 
 # Read-only database replicas
 
-Read-only replicas are additional databases that are used to offload read queries from the primary database, which can improve performance by reducing the load on the primary database.
+Read-only replicas are additional databases that are used to offload read queries from the primary database, which can improve performance by reducing the load on the primary database. This is particularly useful for high-traffic websites with a large number of read operations.
+
+Read-only replicas allow you to scale your database infrastructure horizontally to handle increased traffic. Also, if the primary database fails, read-only replicas can continue to serve read requests, ensuring that your website remains available.
+
+## Configuration
 
 Read-only replicas are configured by adding environment variables that match the primary environment variable and suffixing `_REPLICA_<replica-number>` to the variable name, where `<replica_number>` is the replica number padding by a zero if it's less than 10, for example `SS_DATABASE_SERVER` becomes `SS_DATABASE_SERVER_REPLICA_01` for the first replica, or `SS_DATABASE_SERVER_REPLICA_12` for the 12th replica. Replias must be numbered sequentially starting from `01`.
 
@@ -63,3 +67,9 @@ DB::withPrimary(function () {
 
 // Code here can use a replica
 ```
+
+## Monitoring replica lag
+
+Replica lag is the delay between changes on the primary database being copied to the replicas. It's important to monitor replica lag to ensure data consistency.
+
+Specific tools to monitor replica lag will depend on the database system being used and your hosting service.
--- a/en/02_Developer_Guides/08_Performance/09_Other_Optimisations.md
+++ b/en/02_Developer_Guides/08_Performance/09_Other_Optimisations.md
@@ -47,6 +47,58 @@ Here are some strategies for cache warming:
 - Use a cron job to periodically regenerate cache entries.
 - Warm the cache whenever data is changed or added, such as by implementing the `onAfterWrite()` extension hook on your `DataObject`.
 
+For example, to use a [`BuildTask`](api:SilverStripe\Dev\BuildTask) to warm the cache for popular pages:
+
+```php
+namespace App\BuildTasks;
+
+use GuzzleHttp\Client;
+use SilverStripe\CMS\Model\SiteTree;
+use SilverStripe\Core\Injector\Injector;
+use SilverStripe\Dev\BuildTask;
+use SilverStripe\PolyExecution\PolyOutput;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+
+/**
+ * Make HTTP requests to popular pages on the website to warm any caches
+ * This would need to be run after a deployment has fully completed to ensure that
+ * any new assets such as JavaScript bundles are cached in intermediatory caches such as CDNs.
+ */
+class CacheWarmerTask extends BuildTask
+{
+    protected static string $commandName = 'page-cache-warmer';
+
+    protected string $title = 'Cache warmer';
+
+    protected static string $description = 'Warms the cache for the most popular pages.';
+
+    private const POPULAR_PAGES = [
+        'home',
+        'about-us',
+        'contact-us',
+    ];
+
+    protected function execute(InputInterface $input, PolyOutput $output): int
+    {
+        $pages = SiteTree::get()->filter([
+            'URLSegment' => CacheWarmerTask::POPULAR_PAGES,
+        ]);
+        foreach ($pages as $page) {
+            $guzzle = Injector::inst()->get(Client::class);
+            $url = $page->AbsoluteLink();
+            $response = $guzzle->get($url);
+            if ($response->getStatusCode() == 200) {
+                $output->writeln("Warmed caches for $url");
+            } else {
+                $output->writeln("<error>Error when trying to warm cache for $url</error>");
+            }
+        }
+        return Command::SUCCESS;
+    }
+}
+```
+
 ## Enable opcode cache
 
 Enable an [opcode](https://en.wikipedia.org/wiki/Opcode) cache such as [OPcache for PHP]([OPcache](https://www.php.net/manual/en/book.opcache.php)) to increase the performance of PHP code.
@@ -61,7 +113,7 @@ You can use a third-party Content Delivery Network (CDN) to cache and serve stat
 - Decreases traffic to your main server, improving its performance.
 - Can provide additional security features like DDoS protection
 
-CDNs will typically respect existing [HTTP cache headers](http-cache-headers) which you can configure within your application.
+CDNs will typically respect existing [HTTP cache headers](http_cache_headers) which you can configure within your application.
 
 ## Use a reverse proxy
 
--- a/en/02_Developer_Guides/08_Performance/index.md
+++ b/en/02_Developer_Guides/08_Performance/index.md
@@ -7,4 +7,13 @@ icon: tachometer-alt
 
 # Performance
 
+Making your Silverstripe CMS applications fast and scalable is crucial for providing a great user experience and reducing server costs. This section provides guidance on various performance optimization techniques, ranging from caching strategies to database optimizations.
+
+## Where to start
+
+- Use profiling tools to identify the slowest parts of your application.
+- Implement caching strategies for frequently accessed data and static assets.
+- Focus on quick-wins first such as adding appropriate indexes.
+- Continuously monitor your application's performance and make adjustments as needed.
+
 [CHILDREN Exclude=How_Tos]


en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
```diff
@@ -6,58 +6,53 @@ icon: tachometer-alt
 
 # Partial template caching
 
-Partial template caching is a feature that allows caching of rendered portions of a template so that it does not need to be recomputed on each request.
+Partial template caching allows you to cache rendered portions of a template, avoiding recomputation on each request.
```
ChloeSartorelli:
```suggestion
Partial template caching allows you to cache rendered portions of a template, so that it does not need to be recomputed on each request.
```

I think the style this PR is aiming for is prioritising clear wording - if that's the goal, I think the old wording of this one part was clearer (thinking primarily of people for whom English isn't their first language) even though it's more verbose.

This suggestion is very subjective so feel free to reject with a reason and I won't challenge it.

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
```diff
@@ -6,58 +6,53 @@ icon: tachometer-alt
...
 
-The [Partial template caching](../templates/partial_template_caching) page in the templates section talks in-depth about how to use partial template caching.
+See [Partial template caching](../templates/partial_template_caching) in the templates section for detailed usage instructions.
 
-This page will discuss some of the performance considerations.
+This page discusses performance considerations.
```
ChloeSartorelli:
```suggestion
This page discusses some of the performance considerations.
```
Unless we're sure this is _all_ of the performance considerations, I'd prefer to keep this worded in a way that implies it may not be a complete set.

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
```diff
@@ -6,58 +6,53 @@ icon: tachometer-alt
...
 
 ## Cache block conditionals
 
-Use conditions whenever possible. The cache tag supports defining conditions via either `if` or `unless` keyword.
-Those are optional, however are highly recommended.
+Use conditions whenever possible. The cache tag supports optional `if` or `unless` conditions, which are highly recommended.
 
 > [!WARNING]
-> Avoid performing heavy computations in conditionals, as they are evaluated for every template rendering.
+> Avoid heavy computations in conditionals, as they are evaluated for every template rendering.
```
ChloeSartorelli:
```suggestion
> Avoid heavy computations in conditionals, as they are evaluated every time the template needs to be rendered.
```

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
```diff
@@ -6,58 +6,53 @@ icon: tachometer-alt
...
 
 If you cache without conditions:
 
-- your cache backend will always be queried for the cache block (on every template render)
-- your cache may be cluttered with heaps of redundant and useless data (especially the default filesystem backend)
+- The cache backend will be queried for the cache block on every template render.
```
ChloeSartorelli:
```suggestion
- The cache backend will be queried for the cache block on every template render even if it's not needed.
```

I read this thinking "wait but why _wouldn't_ it be queried?"... forgetting that if there is a condition and the condition is false, the cache won't be needed and therefore won't be queried :sweat_smile:

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
```diff
@@ -6,58 +6,53 @@ icon: tachometer-alt
...
+- Your cache may be cluttered with redundant data, especially with the default filesystem backend.
 
-As an example, if you use `$DataObject->ID` as a key for the block, consider adding a condition that ID is greater than zero:
+For example, if you use `$DataObject->ID` as a key for the block, add a condition that the ID is greater than zero:
```
ChloeSartorelli:
```suggestion
For example, if you use `$DataObject->ID` as a key for the block, consider adding a condition that the ID is greater than zero:
```
You don't _have_ to add a condition just because you're querying for ID - you might know it's always going to be >0 so the condition would be redundant.

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
```diff
@@ -6,58 +6,53 @@ icon: tachometer-alt
...
 
 ## Cache invalidation strategies
 
-### Aggregates based invalidation {#aggregates}
+### Aggregates-based invalidation {#aggregates}
 
-Sometimes you may want to invalidate cache when any object in a set changes, or when objects in a relationship
-change. To do this, you may use [DataList](api:SilverStripe\ORM\DataList) aggregate methods (which we call Aggregates).
-These perform SQL aggregate queries on sets of [DataObject](api:SilverStripe\ORM\DataObject)s.
+To invalidate the cache when any object in a set changes, or when objects in a relationship change, use [DataList](api:SilverStripe\ORM\DataList) aggregate methods (Aggregates). These perform SQL aggregate queries on sets of [DataObject](api:SilverStripe\ORM\DataObject)s.
```
ChloeSartorelli: I prefer the old wording because:

1. Not all data is in a `DataList`
2. There are other ways to solve this problem, such as generating a cache-key on CRUD (as with https://github.com/silverstripe-terraformers/keys-for-cache)

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
```diff
@@ -6,58 +6,53 @@ icon: tachometer-alt
...
 
-Here are some useful methods of the [DataList](api:SilverStripe\ORM\DataList) class:
+Useful methods of the [DataList](api:SilverStripe\ORM\DataList) class:
```
ChloeSartorelli:
```suggestion
Some useful methods of the [`DataList`](api:SilverStripe\ORM\DataList) class for this purpose include:
```
or just keep the old wording

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
```diff
@@ -6,58 +6,53 @@ icon: tachometer-alt
...
-- `mixed avg(string $fieldName)` : Return the average value of the given field in this DataList
-- `mixed sum(string $fieldName)` : Return the sum of the values of the given field in this DataList
+- `int count()`: Returns the number of items in this DataList.
+- `mixed max(string $fieldName)`: Returns the maximum value of the given field in this DataList.
+- `mixed min(string $fieldName)`: Returns the minimum value of the given field in this DataList.
+- `mixed avg(string $fieldName)`: Returns the average value of the given field in this DataList.
+- `mixed sum(string $fieldName)`: Returns the sum of the values of the given field in this DataList.
 
-To construct a `DataList` over a `DataObject`, we have a global template variable called `$List`.
+To construct a `DataList` over a `DataObject`, use the global template variable `$List`.
```
ChloeSartorelli:
```suggestion
To construct a `DataList` over a `DataObject`, you can use the global template variable `$List`.
```
That's one way - probably not even the best practice these days (normally you'd make the list in PHP and get it into the template with a getter method - but since this is just about writing style and not best practices, just the suggestion above is fine)

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
```diff
@@ -6,58 +6,53 @@ icon: tachometer-alt
...
 
-For example, if we have a menu, we may want that menu to update whenever *any* page is edited, but would like to cache it
-otherwise. By using aggregates, we do that like this:
+For example, to update a menu whenever any page is edited, and have it cached otherwise:
```
ChloeSartorelli:
```suggestion
For example, to cache a menu but invalidate that cache whenever any page is edited:
```

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
```diff
@@ -70,21 +65,16 @@ otherwise. By using aggregates, we do that like this:
 > [!NOTE]
 > The use of the fully qualified classname is necessary.
 >
-> The use of both `.max('LastEdited')` and `.count()` makes sure we check for any object
-> edited or deleted since the cache was last built.
+> Using both `.max('LastEdited')` and `.count()` ensures we check for any object edited or deleted since the cache was last built.
 
-The cache for this will update whenever a page is added, removed or edited.
+The cache will update whenever a page is added, removed or edited.
```
ChloeSartorelli:
```suggestion
In this example, the cache will update whenever a page is added, removed or edited.
```

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
```diff
@@ -70,21 +65,16 @@ otherwise. By using aggregates, we do that like this:
...
 
 > [!WARNING]
-> Be careful using aggregates. Remember that the database is usually one of the performance bottlenecks.
-> Keep in mind that every key of every cached block is recalculated for every template render, regardless of caching
-> result. Aggregating SQL queries usually produce more load on the database than simple select queries,
-> especially if you query records by Primary Key or join tables using database indices properly.
+> Be careful when using aggregates. The database is often a performance bottleneck. Every key of every cached block is recalculated for every template render, regardless of caching result. Aggregating SQL queries can put more load on the database than simple select queries, especially if you query records by Primary Key or join tables using database indices properly.
```
ChloeSartorelli:
```suggestion
> Be careful when using aggregates. The database is often a performance bottleneck. Keep in mind that every key of every cached block is recalculated for every template render, regardless of caching result. Aggregating SQL queries can put more load on the database than simple select queries, especially if you query records by Primary Key or join tables using database indices properly.
```
I like this wording from the original - for me it invokes a call to action. It's like "whenever you're using aggregates, keep this in mind" as opposed to just providing me this information as a one-time thing that I'll forget later.

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
```diff
@@ -112,20 +102,17 @@ The cache for this will update whenever a page is added, removed or edited.
 > [controller method](../templates/partial_template_caching/#cache-key-calculated-in-controller).
 > [Object Caching](../templates/caching/#object-caching) only works for single variables and not for chained expressions.
 
-### Time based invalidation
+### Time-based invalidation
 
-In some situations it's more important to be fast than to always be showing the latest data. By constructing the cache
-key to invalidate less often than the data updates you can ensure rendering time is constant no matter how often the
-data updates.
+To prioritise speed over always showing the latest data, construct the cache key to invalidate less often than the data updates. This ensures constant rendering time, regardless of how often the data updates.
```
ChloeSartorelli:
```suggestion
In some situations it's more important to be fast than to always be showing the latest data. One way to achieve this is to construct the cache key to invalidate less often than the data updates. This ensures constant rendering time, regardless of how often the data updates.
```
I prefer the "in some situations" language - to me it helps me think about what my own code is doing, and consider if this is a situation that applies for me.

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
```diff
@@ -112,20 +102,17 @@ The cache for this will update whenever a page is added, removed or edited.
...
-For instance, if we show some blog statistics, but are happy having them be slightly stale, we could do
+For instance, if you show blog statistics, but are happy with slightly stale data:
 
 ```ss
 <% cached 'blogstatistics', $Blog.ID %>
 ```
 
-which will invalidate after the cache lifetime expires. If you need more control than that (cache lifetime is
-[configurable only on a site-wide basis](/developer_guides/templates/partial_template_caching/#cache-storage)), you could add a special function to your controller:
+This will invalidate after the cache lifetime expires. For more control (cache lifetime is [configurable only on a site-wide basis](/developer_guides/templates/partial_template_caching/#cache-storage)), add a function to your controller:
```
ChloeSartorelli:
```suggestion
This doesn't require any additional SQL queries and will invalidate after the cache lifetime expires. Cache lifetime is [configurable only on a site-wide basis](/developer_guides/templates/partial_template_caching/#cache-storage). For more control, add a function to your controller:
```

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
```diff
@@ -144,22 +131,17 @@ class MyObject extends DataObject
...
 
 ## Cache backend
 
-The template engine uses [Injector](../extending/injector) service `Psr\SimpleCache\CacheInterface.cacheblock` as
-caching backend. The default definition of that service is very conservative and relies on the server filesystem.
-This is the most common denominator for most of the applications out there. However,
-this is not the most robust nor performant cache implementation. If you have a better solution
-available on your platform, you should consider tuning that setting for your application.
-All you need to do to swap the cache backend for partial template cache blocks is to redefine this service for the Injector.
+The template engine uses [Injector](../extending/injector) service `Psr\SimpleCache\CacheInterface.cacheblock` as a caching backend. The default definition of that service is conservative, and relies on the server filesystem. While this is the most common denominator, it's not the most robust or performant implementation. If you have a better solution, consider tuning this setting. To swap the cache backend for partial template cache blocks, redefine this service for the Injector.
```
ChloeSartorelli:
```suggestion
The template engine uses [Injector](../extending/injector) service `Psr\SimpleCache\CacheInterface.cacheblock` as its caching backend. The default definition of that service is conservative, and relies on the server filesystem. While this is the most common denominator, it's not the most robust or performant implementation. If you have a better solution available on your platform, you should consider tuning this setting. To swap the cache backend for partial template cache blocks, redefine this service for the Injector.
```

The "available on your platform" part is important here - without it, it's implying "if you can think of a better way of doing this" - with it, it's implying "if your hosting platform has something like redis or memcached, use that"

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
```diff
@@ -176,6 +158,4 @@ SilverStripe\Core\Injector\Injector:
 ```
 
 > [!WARNING]
-> The default filesystem cache backend does not support auto cleanup of the residual files with expired cache records.
-> If your project relies on Template Caching heavily (e.g. thousands of cache records daily), you may want to keep an eye on the
-> filesystem storage. Sooner or later its capacity may be exhausted.
+> The default filesystem cache backend does not support auto cleanup of the residual files with expired cache records. If your project relies on Template Caching heavily (e.g. thousands of cache records daily), monitor the filesystem storage, as its capacity may eventually be exhausted.
```
ChloeSartorelli:
```suggestion
> The default filesystem cache backend does not support auto cleanup of the residual files with expired cache records. If your project relies on Template Caching heavily (e.g. thousands of cache records daily), you may want to monitor the filesystem storage. If you don't, its capacity may eventually be exhausted.
```
I prefer "you may want to" instead of telling people what to do, though I don't feel strongly about that.
I do feel a little stronger about adding "If you don't," instead of "as"

en/02_Developer_Guides/08_Performance/01_Object_Caching.md
```diff
@@ -7,10 +7,16 @@ icon: tachometer-alt
 
 ## Overview
 
-The framework uses caches to store infrequently changing values.
+Object caching stores infrequently changing values to improve performance.
 By default, the storage mechanism chooses the most performant adapter available
 (PHP opcache, APC, or filesystem). Other cache backends can be configured.
```
ChloeSartorelli:
```suggestion
By default, the storage mechanism chooses the most performant filesystem adapter available
(PHP opcache, or filesystem) but does not use in-memory cache unless you configure it. Other cache backends can be configured.
```
Oops, I missed that reference to APCu when removing that default for CMS 6.

en/02_Developer_Guides/08_Performance/01_Object_Caching.md
```diff
@@ -7,10 +7,16 @@ icon: tachometer-alt
...
 
+Object caching is particularly useful for:
+
+- Storing the results of expensive database queries.
+- Caching data retrieved from external APIs.
+- Reducing the load on frequently accessed resources.
```
ChloeSartorelli: The classic AI list of 3 :rofl:

I don't feel strongly about this suggestion so feel free to leave it if you prefer the list.

```suggestion
Object caching is useful for situations where fetching or generating data can take a long time, such as expensive database queries or API calls to external services.
```

en/02_Developer_Guides/08_Performance/01_Object_Caching.md
```diff
@@ -32,11 +38,8 @@ For long-lived server instances, this can become a capacity issue over time - se
 
 ## Configuration
 
-We are using the [PSR-16](https://www.php-fig.org/psr/psr-16/) standard ("SimpleCache")
-for caching, through the [symfony/cache](https://symfony.com/doc/current/components/cache.html) library.
-
-Note that this library describes usage of [PSR-6](https://www.php-fig.org/psr/psr-6/) by default,
-though Silverstripe wraps these in a PSR-16 interface using the [Psr16Cache](https://github.com/symfony/cache/blob/6.1/Psr16Cache.php) class.
+We use the [PSR-16](https://www.php-fig.org/psr/psr-16/) standard ("SimpleCache")
+for caching, through the [symfony/cache](https://symfony.com/doc/current/components/cache.html) library. Silverstripe provides a PSR-16 interface via the [Psr16Cache](https://github.com/symfony/cache/blob/6.1/Psr16Cache.php) class which wraps [PSR-6](https://www.php-fig.org/psr/psr-6/) usage in symfony/cache.
```
ChloeSartorelli:
```suggestion
for caching, through the [symfony/cache](https://symfony.com/doc/current/components/cache.html) library.

Silverstripe provides a PSR-16 interface via the [`Psr16Cache`](https://github.com/symfony/cache/blob/6.1/Psr16Cache.php) class which wraps `symfony/cache`'s [PSR-6](https://www.php-fig.org/psr/psr-6/) compliant adapters.
```

en/02_Developer_Guides/08_Performance/02_Cache_Adapters.md
```diff
@@ -6,19 +6,33 @@ icon: tachometer-alt
...
 
+## Choosing a cache adapter
+
+|Adapter|Pros|Cons|Use cases|
+|---|---|---|---|
+|`PhpFilesAdapter`|Fast, uses opcode cache|Requires opcache, may need memory tuning|General-purpose, good for production if opcache is configured correctly|
+|`FilesystemAdapter`|Simple, reliable|Slower than in-memory caches|Development, fallback when other adapters aren't available|
+|`Memcached`|High-performance, distributed|Requires memcached server and PHP extension|High-traffic websites, distributed caching|
+|`Redis`|In-memory database, persistent, replication|Requires Redis server and PHP extension|High-traffic websites, persistent caching, complex caching scenarios|
+|`APCu`|Simple to set up|Limited to webserver, cannot be shared with CLI, no persistence|Small projects, development, simple caching needs|
```
ChloeSartorelli: I'd prefer not to use this table. I don't think it's a fair representation for comparison.

For example, it says `Memcached` is "High-performance, distributed" but doesn't say the same about redis - which is also high performance and also useful for distributed hosting. I assume the AI just picked that language up from the wording below, which is just a quote from the first sentence on https://www.danga.com/memcached/

There's more to memcached and redis especially than we can easily capture in this format, and afaik both are actively being developed so anything we say about them is liable to becoming stale.

ChloeSartorelli: If you'd like to list the in-memory adapters that are available in a list (i.e. to show them in a way that doesn't require scrolling), a regular list inside the "Adding an in-memory cache adapter" before "Memcached cache" might be better, e.g.

```md
The in-memory cache adapters provided by default are:

- Memcached
- Redis
- APCu
```

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
```diff
@@ -6,23 +6,26 @@ icon: tachometer-alt
 
 # HTTP cache headers
 
-By default, Silverstripe CMS sends headers which signal to HTTP caches
-that the response should be not considered cacheable.
-HTTP caches can either be intermediary caches (e.g. CDNs and proxies), or clients (e.g. browsers).
-The cache headers sent are `Cache-Control: no-cache, must-revalidate`;
+HTTP cache headers instruct HTTP caches (browsers, CDNs, proxies) how to store and serve responses. Proper configuration can significantly improve website performance.
```
ChloeSartorelli:
```suggestion
HTTP cache headers instruct HTTP caches (e.g. browsers, CDNs, and proxies) how to store and serve responses. Proper configuration can significantly improve website performance.
```

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
```diff
@@ -6,23 +6,26 @@ icon: tachometer-alt
...
+
+By default, Silverstripe CMS sends headers that prevent caching: `Cache-Control: no-cache, must-revalidate`.
 
 > [!WARNING]
-> HTTP caching can be a great way to speed up your website, but needs to be properly applied.
-> Getting it wrong can accidentally expose draft pages or other protected content.
+> HTTP caching can greatly speed up your website, but must be configured correctly. Incorrect settings can expose draft pages or other protected content.
```
ChloeSartorelli: I prefer the original wording because:

1. "applied" instead of "configured" - it's not just about the configuration, it's also about when and where you choose to use it
2. "a great way" instead of "greatly" - but this is hugely subjective, you can keep "greatly" if you prefer it
3. "getting it wrong" has more of a warning sound to me than "incorrect settings". Makes me consider more carefully what I'm doing

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
```diff
@@ -6,23 +6,26 @@ icon: tachometer-alt
...
+
+## Understanding HTTP caching
+
+HTTP caching involves storing responses (HTML, CSS, JavaScript, images) to reduce latency and server load. Caches can be located in:
```
ChloeSartorelli:
```suggestion
HTTP caching involves storing responses (e.g. HTML, CSS, JavaScript, images) to reduce latency and server load. Caches can be located in:
```
Can be other types of responses like JSON - instead of adding to the list, just an `e.g.` is fine.

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
```diff
@@ -6,23 +6,26 @@ icon: tachometer-alt
...
+
+- Browsers: Stores responses for a single user.
+- CDNs (Content Delivery Networks): Distributes content across geographically diverse servers.
+- Proxies: Intermediary servers that cache responses for multiple users.
```
ChloeSartorelli: I don't think we need to go into this much detail. These are already mentioned in the opening paragraph, and anyone who doesn't already know what these are can duckduckgo them.

Without this list, the section doesn't really need to exist as its own section. I recommend adapting the initial sentence ("HTTP caching involved storing responses [etc]") to fit inside the opening section instead.

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
```diff
@@ -6,23 +6,26 @@ icon: tachometer-alt
...
 
 ## Cache control headers
 
 ### Overview
 
-In order to support developers in making safe choices around HTTP caching,
-we're using a `HTTPCacheControlMiddleware` class to control if a response
-should be considered public or private. This is an abstraction on the
-`HTTPResponse->addHeader()` low-level API.
+Silverstripe CMS provides `HTTPCacheControlMiddleware` to manage HTTP cache headers safely. This class helps prevent accidental caching of draft content or content generated with active sessions. It's an abstraction on top of `HTTPResponse->addHeader()`.
```
ChloeSartorelli:
```suggestion
Silverstripe CMS provides `HTTPCacheControlMiddleware` to help developers manage HTTP cache headers safely. This class helps prevent accidental caching of private content such as draft content or content generated with active sessions. It's an abstraction on top of the `HTTPResponse->addHeader()` low-level API.
```

Also I'm not sure what you mean by "content generated with active sessions", so that probably needs to be reworded.

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
```diff
@@ -44,30 +47,25 @@ It comes with methods which let developers safely interact with the `Cache-Contr
 
 ### DisableCache()
 
-Simple way to set cache control header to a non-cacheable state.
-Use this method over `privateCache()` if you are unsure about caching details.
-Takes precedence over unforced `enableCache()`, `privateCache()` or `publicCache()` calls.
+Sets cache control headers to a non-cacheable state. Use this method if you are unsure about caching details. Takes precedence over unforced `enableCache()`, `privateCache()` or `publicCache()` calls.
```
ChloeSartorelli:
```suggestion
Sets cache control headers to a non-cacheable state. Use this method instead of `privateCache()` if you are unsure about caching details. Takes precedence over unforced `enableCache()`, `privateCache()` or `publicCache()` calls.
```
This from the original content is important I think.

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
```diff
@@ -44,30 +47,25 @@ It comes with methods which let developers safely interact with the `Cache-Contr
...
 
-Removes all state and replaces it with `no-cache, no-store, must-revalidate`. Although `no-store` is sufficient
-the others are added under [recommendation from Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#Examples)
+Removes all state and replaces it with `no-cache, no-store, must-revalidate`. Although `no-store` is sufficient the others are added under [recommendation from Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#Examples)
 
 Does not set `private` directive, use `privateCache()` if this is explicitly required
 ([details](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching#public_vs_private))
```
ChloeSartorelli: This URL redirects to a document that no longer compares public and private

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
```diff
@@ -44,30 +47,25 @@ It comes with methods which let developers safely interact with the `Cache-Contr
...
-Simple way to set cache control header to a cacheable state.
-Use this method over `publicCache()` if you are unsure about caching details.
+Sets cache control headers to a cacheable state. Use this method if you are unsure about caching details.
 
-Removes the `no-store` directive unless a `max-age` is set; other directives will remain in place.
-Use alongside `setMaxAge()` to activate caching.
+Removes the `no-store` directive unless a `max-age` is set; other directives will remain in place. Use alongside `setMaxAge()` to activate caching.
 
 Does not set `public` directive. Usually, `setMaxAge()` is sufficient. Use `publicCache()` if this is explicitly required
 ([details](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching#public_vs_private))
```
ChloeSartorelli: This URL redirects to a document that no longer compares public and private

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
```diff
@@ -44,30 +47,25 @@ It comes with methods which let developers safely interact with the `Cache-Contr
...
 ### DisableCache()
 
-Simple way to set cache control header to a non-cacheable state.
-Use this method over `privateCache()` if you are unsure about caching details.
-Takes precedence over unforced `enableCache()`, `privateCache()` or `publicCache()` calls.
+Sets cache control headers to a non-cacheable state. Use this method if you are unsure about caching details. Takes precedence over unforced `enableCache()`, `privateCache()` or `publicCache()` calls.
 
-Removes all state and replaces it with `no-cache, no-store, must-revalidate`. Although `no-store` is sufficient
-the others are added under [recommendation from Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#Examples)
+Removes all state and replaces it with `no-cache, no-store, must-revalidate`. Although `no-store` is sufficient the others are added under [recommendation from Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#Examples)
```
ChloeSartorelli: That advice is no longer given, and is now explicitly called out as "basically meaningless":
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cache-Control#preventing_storing

> In theory, if directives are conflicted, the most restrictive directive should be honored. So the example below is basically meaningless because private, no-cache, max-age=0 and must-revalidate conflict with no-store.
> ```text
> # conflicted
> Cache-Control: private, no-cache, no-store, max-age=0, must-revalidate
>
> # equivalent to
> Cache-Control: no-store
> ```

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
```diff
@@ -44,30 +47,25 @@ It comes with methods which let developers safely interact with the `Cache-Contr
...
 
 Does not set `private` directive, use `privateCache()` if this is explicitly required
 ([details](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching#public_vs_private))
 
 ### EnableCache()
 
-Simple way to set cache control header to a cacheable state.
-Use this method over `publicCache()` if you are unsure about caching details.
+Sets cache control headers to a cacheable state. Use this method if you are unsure about caching details.
```
ChloeSartorelli:
```suggestion
Sets cache control headers to a cacheable state. Use this method instead of `publicCache()` if you are unsure about caching details.
```
Again, this part from the original content is important.

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
```diff
@@ -44,30 +47,25 @@ It comes with methods which let developers safely interact with the `Cache-Contr
...
-Use alongside `setMaxAge()` to activate caching.
+Removes the `no-store` directive unless a `max-age` is set; other directives will remain in place. Use alongside `setMaxAge()` to activate caching.
 
 Does not set `public` directive. Usually, `setMaxAge()` is sufficient. Use `publicCache()` if this is explicitly required
 ([details](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching#public_vs_private))
 
 ### PrivateCache()
 
-Advanced way to set cache control header to a non-cacheable state.
+Sets cache control headers to a non-cacheable state.
```
ChloeSartorelli: The fact that this is "advanced" compared to the "simple" `DisableCache()` is important context that's gotten lost.

Maybe something like

```suggestion
Sets cache control headers to a non-cacheable state. This is similar to `DisableCache()`, except
```
and then say what makes it different.

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
```diff
@@ -77,7 +75,7 @@ A private cache (e.g. Web Browser) may store the response. Also removes `public`
 
 ### PublicCache()
 
-Advanced way to set cache control header to a cacheable state.
+Sets cache control headers to a cacheable state.
```
ChloeSartorelli: The fact that this is "advanced" compared to the "simple" `EnableCache()` is important context that's gotten lost.

Maybe something like

```suggestion
Sets cache control headers to a cacheable state. This is similar to `EnableCache()`, except
```
and then say what makes it different.

en/02_Developer_Guides/08_Performance/04_Static_Publishing.md
```diff
@@ -5,15 +5,16 @@ summary: Export web pages as static HTML
 
 # Static publishing
 
-Static publishing is the process of exporting website pages as static HTML files which are then served directly. This means that page content does not need to be recomputed on each request. This can have the following benefits:
+Static publishing exports website pages as static HTML files served directly. This eliminates the need to recompute page content on each request, offering the following benefits:
```
ChloeSartorelli: I actually prefer all of the original language here. I think it's easier to understand.
Especially "this can have" is important, since no caching is ever a magic bullet - I'm sure this can be done in a way that doesn't provide some of these benefits.

en/02_Developer_Guides/08_Performance/04_Static_Publishing.md
```diff
@@ -5,15 +5,16 @@ summary: Export web pages as static HTML
...
 
-- Static HTML files load much faster than dynamically generated pages as they don't require any server-side processing.
-- Static publishing reduces the load on the server, as it doesn't need to connect to the database or render templates for each request.
-- Static publishing can lower hosting costs because there is less need for server resources.
+- Faster load times: Static HTML files load much faster than dynamically generated pages.
+- Reduced server load: No database connection or template rendering is required for each request.
+- Lower hosting costs: Less server resources are needed.
```
ChloeSartorelli:
```suggestion
- Lower hosting costs: Less server resources are needed to serve statically cached content.
```

en/02_Developer_Guides/08_Performance/06_ORM.md
```diff
@@ -5,9 +5,20 @@ summary: Configuration and tips for improving ORM performance
...
+## Identifying ORM performance bottlenecks
+
+ORM performance issues can arise from various factors, including:
+
+- Inefficient database queries
+- Lack of proper indexes
+- Large datasets
+- Complex relationships
+
+You can use [lekoala/silverstripe-debugbar](https://github.com/lekoala/silverstripe-debugbar?tab=readme-ov-file#database-profiling) to identify slow running database queries.
```
ChloeSartorelli: That's not CMS 6 compatible currently.
Maybe call out the [`showqueries` URL variable](https://docs.silverstripe.org/en/6/developer_guides/debugging/url_variable_tools/#database) instead.

en/02_Developer_Guides/08_Performance/06_ORM.md
```diff
@@ -5,9 +5,20 @@ summary: Configuration and tips for improving ORM performance
...
+
 ## Indexes
 
-You can define indexes for your ORM queries using the `$indexes` configuration property in your `DataObject` subclasses. See the [Indexes](/developer_guides/model/indexes) section for more information.
+You can define indexes for your ORM queries using the `$indexes` configuration property in your `DataObject` subclasses. See the [Indexes](/developer_guides/model/indexes) section for more information. Adding indexes to frequently queried fields can significantly improve performance.
```
ChloeSartorelli:
```suggestion
Adding indexes to frequently queried fields can significantly improve performance. You can define indexes for your ORM queries using the `$indexes` configuration property in your `DataObject` subclasses. See the [Indexes](/developer_guides/model/indexes) section for more information.
```

en/02_Developer_Guides/08_Performance/06_ORM.md
```diff
@@ -49,7 +60,7 @@ SilverStripe\ORM\Connect\DBSchemaManager:
 ```
 
 > [!WARNING]
-> Excluding models from database checks can lead to undetected data corruption or other issues. Only exclude models if you are certain of what you are doing.
+> Excluding models from database checks can lead to undetected data corruption or other issues. Only exclude models if you are certain of what you are doing. Consider testing these changes thoroughly in a development environment before applying them to production.
```
ChloeSartorelli:
```suggestion
> Excluding models from database checks can lead to undetected data corruption or other issues. Only exclude models if you are certain of what you are doing.
```
The problem with it is more of a "down the line the checks might fail" - checking it works fine once is likely to be a footgun as you may be fooled into thinking "it worked this time so it'll always work"

en/02_Developer_Guides/08_Performance/07_Eager_Loading.md
```diff
@@ -219,3 +219,30 @@ $teams = Team::get()->eagerLoad([
...
+For example, you can cache the results of an eager-loaded `DataList` using object caching:
+
+```php
+use Psr\SimpleCache\CacheInterface;
+use SilverStripe\Core\Injector\Injector;
+
+$cache = Injector::inst()->get(CacheInterface::class . '.myCache');
+$cacheKey = 'teams_with_players';
+
+$teams = $cache->get($cacheKey)
```
ChloeSartorelli:
```suggestion
$teams = $cache->get($cacheKey);
```

en/02_Developer_Guides/08_Performance/07_Eager_Loading.md
```diff
@@ -219,3 +219,30 @@ $teams = Team::get()->eagerLoad([
...
+
+To further optimize performance, consider combining eager loading with object caching or partial template caching. Eager loading reduces the number of database queries, while caching stores the results for subsequent requests.
+
+For example, you can cache the results of an eager-loaded `DataList` using object caching:
+
+```php
+use Psr\SimpleCache\CacheInterface;
+use SilverStripe\Core\Injector\Injector;
+
+$cache = Injector::inst()->get(CacheInterface::class . '.myCache');
```
ChloeSartorelli: This service doesn't exist - need the YAML snippet as well.

```yaml
SilverStripe\Core\Injector\Injector:
  Psr\SimpleCache\CacheInterface.myCache:
    factory: SilverStripe\Core\Cache\CacheFactory
    constructor:
      namespace: "myCache"
```

May be better to just link to object caching docs instead of giving an example here.

en/02_Developer_Guides/08_Performance/08_Read_only_database_replicas.md
```diff
@@ -5,7 +5,15 @@ summary: Using read-only database replicas to improve performance
 
 # Read-only database replicas
 
-Read-only replicas are additional databases that are used to offload read queries from the primary database, which can improve performance by reducing the load on the primary database.
+Read-only replicas are additional databases that are used to offload read queries from the primary database, which can improve performance by reducing the load on the primary database. This is particularly useful for high-traffic websites with a large number of read operations.
+
+## Benefits of read-only replicas
+
+- By distributing read queries across multiple databases, you can reduce the load on the primary database and improve overall performance.
```
ChloeSartorelli:
```suggestion
```
This first suggestion basically gives the same benefit as the second one - but I prefer the wording of the second one.

en/02_Developer_Guides/08_Performance/08_Read_only_database_replicas.md
```diff
@@ -5,7 +5,15 @@ summary: Using read-only database replicas to improve performance
 
 # Read-only database replicas
 
-Read-only replicas are additional databases that are used to offload read queries from the primary database, which can improve performance by reducing the load on the primary database.
+Read-only replicas are additional databases that are used to offload read queries from the primary database, which can improve performance by reducing the load on the primary database. This is particularly useful for high-traffic websites with a large number of read operations.
+
+## Benefits of read-only replicas
```
ChloeSartorelli:
```suggestion
Some of the benefits of read-only database replicas include:
```

This short list doesn't really warrant its own section

en/02_Developer_Guides/08_Performance/08_Read_only_database_replicas.md
```diff
@@ -63,3 +71,9 @@ DB::withPrimary(function () {
 
 // Code here can use a replica
 ```
+
+## Monitoring replica lag
+
+It's important to monitor replica lag (the delay between changes on the primary database and their replication to the replicas) to ensure data consistency.
```
ChloeSartorelli:
```suggestion
Replica lag is the delay between changes on the primary database being copied to the replicas. It's important to monitor replica lag to ensure data consistency.
```

en/02_Developer_Guides/08_Performance/08_Read_only_database_replicas.md
```diff
@@ -63,3 +71,9 @@ DB::withPrimary(function () {
...
+
+Specific tools to monitor replica lag will depend on the database system being used.
```
ChloeSartorelli:
```suggestion
Specific tools to monitor replica lag will depend on the database system being used and your hosting service.
```
e.g. regardless of what database you use on AWS they probably have their own standardised dashboard that abstracts it.

en/02_Developer_Guides/08_Performance/09_Other_Optimisations.md
```diff
@@ -47,6 +47,50 @@ Here are some strategies for cache warming:
 - Use a cron job to periodically regenerate cache entries.
 - Warm the cache whenever data is changed or added, such as by implementing the `onAfterWrite()` extension hook on your `DataObject`.
 
+For example, to use a [`BuildTask`](api:SilverStripe\Dev\BuildTask) to warm the cache for popular pages:
+
+```php
+// app/src/BuildTasks/CacheWarmerTask.php
```
ChloeSartorelli:
```suggestion
```
The file path is only needed when it's relevant (e.g. content in `_config.php` needs to be explicitly mentioned as living in that special file)

en/02_Developer_Guides/08_Performance/09_Other_Optimisations.md
```diff
@@ -47,6 +47,50 @@ Here are some strategies for cache warming:
...
+        'Home',
+        'About Us',
+        'Contact Us',
+    ];
+
+    protected function execute(InputInterface $input, PolyOutput $output): int
+    {
+        $pages = SiteTree::get()->filter([
+            'Title' => CacheWarmerTask::POPULAR_PAGES,
+        ]);
```
ChloeSartorelli: Use URLSegments, which are less likely to change than titles
```suggestion
        $pages = SiteTree::get()->filter([
            'URLSegment' => CacheWarmerTask::POPULAR_PAGES,
        ]);
```

en/02_Developer_Guides/08_Performance/09_Other_Optimisations.md
```diff
@@ -47,6 +47,50 @@ Here are some strategies for cache warming:
...
+class CacheWarmerTask extends BuildTask
+{
+    protected string $title = 'Cache warmer';
+
+    protected static string $description = 'Warms the cache for the most popular pages.';
+
+    private const POPULAR_PAGES = [
+        'Home',
+        'About Us',
+        'Contact Us',
```
ChloeSartorelli:
```suggestion
        'home',
        'about-us',
        'contact-us',
```

en/02_Developer_Guides/08_Performance/09_Other_Optimisations.md
```diff
@@ -47,6 +47,50 @@ Here are some strategies for cache warming:
...
+    ];
+
+    protected function execute(InputInterface $input, PolyOutput $output): int
+    {
+        $pages = SiteTree::get()->filter([
+            'Title' => CacheWarmerTask::POPULAR_PAGES,
+        ]);
+        foreach ($pages as $page) {
+            $url = Director::absoluteURL($page->Link());
```
ChloeSartorelli:
```suggestion
            $url = $page->AbsoluteLink();
```

en/02_Developer_Guides/08_Performance/09_Other_Optimisations.md
```diff
@@ -47,6 +47,50 @@ Here are some strategies for cache warming:
...
 - Warm the cache whenever data is changed or added, such as by implementing the `onAfterWrite()` extension hook on your `DataObject`.
 
+For example, to use a [`BuildTask`](api:SilverStripe\Dev\BuildTask) to warm the cache for popular pages:
+
+```php
+// app/src/BuildTasks/CacheWarmerTask.php
+namespace App\BuildTasks;
+
+use SilverStripe\CMS\Model\SiteTree;
+use SilverStripe\Control\Director;
```
ChloeSartorelli:
```suggestion
```

en/02_Developer_Guides/08_Performance/09_Other_Optimisations.md
```diff
@@ -47,6 +47,50 @@ Here are some strategies for cache warming:
...
+
+    protected function execute(InputInterface $input, PolyOutput $output): int
+    {
+        $pages = SiteTree::get()->filter([
+            'Title' => CacheWarmerTask::POPULAR_PAGES,
+        ]);
+        foreach ($pages as $page) {
+            $url = Director::absoluteURL($page->Link());
+            file_get_contents($url);
+            $output->write("Warned caches for $url", true);
```
ChloeSartorelli:
```suggestion
            $output->writeln("Warmed caches for $url");
```

en/02_Developer_Guides/08_Performance/09_Other_Optimisations.md
```diff
@@ -47,6 +47,50 @@ Here are some strategies for cache warming:
...
+        }
+        return 0;
```
ChloeSartorelli: Using the const makes it clearer what this returned value represents (otherwise someone may go "oh I should return the number of pages affected instead of 0")
```suggestion
        return Command::SUCCESS;
```
and add `use Symfony\Component\Console\Command\Command;`

en/02_Developer_Guides/08_Performance/09_Other_Optimisations.md
```diff
@@ -47,6 +47,50 @@ Here are some strategies for cache warming:
...
+    ];
+
+    protected function execute(InputInterface $input, PolyOutput $output): int
+    {
+        $pages = SiteTree::get()->filter([
+            'Title' => CacheWarmerTask::POPULAR_PAGES,
+        ]);
+        foreach ($pages as $page) {
+            $url = Director::absoluteURL($page->Link());
+            file_get_contents($url);
```
ChloeSartorelli: Guzzle is one of our fixed dependencies so may as well use it to give a more robust example

```suggestion
            $guzzle = Injector::inst()->get(Client::class);
            $response = $guzzle->get($url);
            if ($response->getStatusCode() !== 200) {
                $output->writeln("<error>Error when trying to warm cache for $url</error>");
            }
```
And add
```php
use SilverStripe\Core\Injector\Injector;
use GuzzleHttp\Client;
```

And put the `writeln` from below in an `else` block

en/02_Developer_Guides/08_Performance/09_Other_Optimisations.md
```diff
@@ -47,6 +47,50 @@ Here are some strategies for cache warming:
...
+use Symfony\Component\Console\Input\InputInterface;
+
+/**
+ * Make HTTP requests to popular pages on the website to warm any caches
+ * This would need to be run after a deployment has fully completed to ensure that
+ * any new assets such as JavaScript bundles are cached in intermediatory caches such as CDNs.
+ */
+class CacheWarmerTask extends BuildTask
+{
+    protected string $title = 'Cache warmer';
```
ChloeSartorelli, line 72:
```suggestion
    protected static string $commandName = 'page-cache-warmer';
    
    protected string $title = 'Cache warmer';
```

en/02_Developer_Guides/08_Performance/index.md
```diff
@@ -7,4 +7,21 @@ icon: tachometer-alt
...
 
+Making your Silverstripe CMS applications fast and scalable is crucial for providing a great user experience and reducing server costs. This section provides guidance on various performance optimization techniques, ranging from caching strategies to database optimizations.
+
+## Key performance considerations
+
+- Implement caching strategies to reduce database queries and server load. See [Partial template caching](partial_template_caching), [Object caching](object_caching), and [HTTP cache headers](http-cache-headers).
+- Optimize your database queries and schema to improve ORM performance. See [ORM performance](orm).
+- Use eager loading to avoid the N+1 query problem and improve database query efficiency. See [Eager loading](eager_loading).
+- Use read-only database replicas to offload read queries from the primary database. See [Read-only database replicas](read-only-database-replicas).
+- Explore other backend performance optimizations, such as increasing server resources, enabling opcode caching, using a CDN, and using a reverse proxy. See [Other optimizations](other_optimisations).
```
ChloeSartorelli: These links have a mix of `-` and `_` - should all be using `_`.
Please check the other links across the performance section as well.

ChloeSartorelli: Also.... is this just listing all the sections? If so, this is redundant to the `[CHILDREN]` block which already has links and descriptions for these sections. It also feels very similar in purpose to "where to start" which I think gives a better idea of what sort of things to consider.

en/02_Developer_Guides/08_Performance/index.md
```diff
@@ -7,4 +7,21 @@ icon: tachometer-alt
...
+
+## Where to start
+
+- Use profiling tools to identify the slowest parts of your application. A good starting point it to simply view the page response time in the network panel in your browsers web developer tools.
```
ChloeSartorelli:
```suggestion
- Use profiling tools to identify the slowest parts of your application.
```

en/02_Developer_Guides/08_Performance/index.md
```diff
@@ -7,4 +7,21 @@ icon: tachometer-alt
...
+- Implementing caching strategies for frequently accessed data and static assets.
+- Focus on optimizing the slowest database queries and adding appropriate indexes.
```
ChloeSartorelli:
```suggestion
- Focus on quick-wins first such as adding appropriate indexes.
```
Not necessarily just slow db queries - but anything that is a quick win should be the first focus. The quick wins for a given project may have nothing to do with the database, in some cases.

GuySartorelli (changes requested):

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/01_Object_Caching.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/01_Object_Caching.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
emteknetnz: Reverted

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/01_Object_Caching.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/02_Cache_Adapters.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
emteknetnz: Reverted

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
emteknetnz: Removed section

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
emteknetnz: Removed link

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
emteknetnz: Removed link

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
emteknetnz: Have removed sentence

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
emteknetnz: Advanced isn't really the right word, the key thing is that it includes the `public` directive while EnableCache() does not

I've updated a few bits to make this clearer.

en/02_Developer_Guides/08_Performance/04_Static_Publishing.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/04_Static_Publishing.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/06_ORM.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/06_ORM.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/06_ORM.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/07_Eager_Loading.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/07_Eager_Loading.md
emteknetnz: I've just deleted the whole section instead

en/02_Developer_Guides/08_Performance/08_Read_only_database_replicas.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/08_Read_only_database_replicas.md
emteknetnz: Updated, and converted the bullets to sentence

en/02_Developer_Guides/08_Performance/08_Read_only_database_replicas.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/08_Read_only_database_replicas.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/09_Other_Optimisations.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/09_Other_Optimisations.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/09_Other_Optimisations.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/09_Other_Optimisations.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/09_Other_Optimisations.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/09_Other_Optimisations.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/09_Other_Optimisations.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/09_Other_Optimisations.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/09_Other_Optimisations.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/index.md
emteknetnz: Updated, there were a handful of one that needed to be changed around the place

emteknetnz: Removed

en/02_Developer_Guides/08_Performance/index.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/index.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/00_Partial_Template_Caching.md
ChloeSartorelli: (done)

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
ChloeSartorelli: "content generated with active sessions" wording is still there without having changed. I still don't know what you mean by that?

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
```diff
@@ -77,8 +68,8 @@ A private cache (e.g. Web Browser) may store the response. Also removes `public`
 
 ### PublicCache()
 
-Advanced way to set cache control header to a cacheable state.
-Indicates that the response may be cached by any cache (e.g. CDNs, Proxies, Web browsers).
+Sets cache control headers to a cacheable state using the `public` directive. This is similar to `EnableCache()`, except it indicates that the response may be cached by any cache (e.g. CDNs, Proxies, Web browsers).
```
ChloeSartorelli: The `EnableCache()` section doesn't say what it can be cached by, so I gather one or more of those (CDNs, Proxies, or Web browsers) isn't applicable for `EnableCache()` - is that right? If so, that should be specified in the `EnableCache()` section.

ChloeSartorelli:
```suggestion
Sets cache control headers to a cacheable state using the `public` directive. This is similar to `EnableCache()`, except it indicates that the response may be cached by any cache (e.g. CDNs, proxies, and browsers).
```
More consistent with how we refer to these in the opening paragraph of the page

en/02_Developer_Guides/08_Performance/06_ORM.md
```diff
@@ -5,9 +5,20 @@ summary: Configuration and tips for improving ORM performance
...
+## Identifying ORM performance bottlenecks
+
+ORM performance issues can arise from various factors, including:
+
+- Inefficient database queries
+- Lack of proper indexes
+- Large datasets
+- Complex relationships
+
+You can append [?showqueries=1](https://docs.silverstripe.org/en/6/developer_guides/debugging/url_variable_tools/#database) to the URL on a dev environment to identify slow running database queries.
```
ChloeSartorelli:
```suggestion
You can use the [`showqueries` variable tool](/developer_guides/debugging/url_variable_tools/#database) on a dev environment to identify slow running database queries.
```
`showqueries` accepts a few different values that give different output, e.g.`showqueries=inline` will show you the values in the SQL instead of `?` parameter placeholders.
This is also consistent with one other link to that page that I'm aware of

en/02_Developer_Guides/08_Performance/09_Other_Optimisations.md
```diff
@@ -47,6 +47,58 @@ Here are some strategies for cache warming:
...
+    ];
+
+    protected function execute(InputInterface $input, PolyOutput $output): int
+    {
+        $pages = SiteTree::get()->filter([
+            'URLSegment' => CacheWarmerTask::POPULAR_PAGES,
+        ]);
+        foreach ($pages as $page) {
+            $guzzle = Injector::inst()->get(Client::class);
+            $url = $url = $page->AbsoluteLink();
```
ChloeSartorelli:
```suggestion
            $url = $page->AbsoluteLink();
```

GuySartorelli (changes requested):
Made a comment in https://github.com/silverstripe/developer-docs/pull/766#discussion_r2153596973 before formally starting the review

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
emteknetnz: Changed to member specific content

en/02_Developer_Guides/08_Performance/03_HTTP_Cache_Headers.md
emteknetnz: Yeah it's a bit of a gray area. enableCache() is honestly just confusing

I've redone this section a bit and put publicCache() up the top because "Yes cache this in everything" is probably what  you want. I see no reason not to include `public` as that clearly communicates intent.

enableCache() remove the `public` directive, so it *might* get cached in a CDN, which isn't really helpful, so I've moved it down the list to give it less prominence

I think that lack of `public` will only *hopefully* prevent caching something like authentication details in a public cache if that public cache happened to have enough internal logic to realise that what was being transmitted was in fact authentication details. But that's a pretty bad reason, instead you should be explicitly calls disableCache() instead (or privateCache(), though that has its own set of problems)

emteknetnz: I've changed it instead to simply `Sets cache control headers to a cacheable state. Will use the `public` directive and remove the `private` if present as that is a contradictory directive.` and I'm not even mentioning enableCache()

en/02_Developer_Guides/08_Performance/06_ORM.md
emteknetnz: Updated

en/02_Developer_Guides/08_Performance/09_Other_Optimisations.md
emteknetnz: Opps fixed

en/02_Developer_Guides/08_Performance/index.md
```diff
@@ -7,4 +7,13 @@ icon: tachometer-alt
 
 # Performance
 
+Making your Silverstripe CMS applications fast and scalable is crucial for providing a great user experience and reducing server costs. This section provides guidance on various performance optimization techniques, ranging from caching strategies to database optimizations.
+
+## Where to start
+
+- Use profiling tools to identify the slowest parts of your application.
+- Implementing caching strategies for frequently accessed data and static assets.
```
michalkleiner:
```suggestion
- Implement caching strategies for frequently accessed data and static assets.
```
otherwise it's the odd one in the list

GuySartorelli (changes requested):
Just Michal's suggested change to go

GuySartorelli (approved):

Issue #770, opened 2025-06-30
Rearrange 6.1.0 changelog


GuySartorelli:
Most of the notes in the CMS 6.1.0 changelog so far are performance improvements, including some of the items under "features and enhancements".

<details>
<summary>Original suggestion</summary>

I think we should move the "performance improvements" section as a level 2 heading, and move any performance related enhancements into that section.
That would rearrange the changelog like so:

- Performance improvements
  - Optimisation of database queries for common CMS operations
  - Improved performance working with assets
  - Composite indexes for `default_sort`
  - Drop indexes in `indexes` configuration
  - Non-blocking file-based sessions
  - Other performance improvements
- Features and enhancements
    - Password strength feedback
    - Other new features
- API changes
- Bug fixes

Obviously anything else that has been added to the changelog since then would fit under either "performance improvements" or "features and enhancements" as appropriate.

The reasons for this suggestion are:
- it allows the individual performance improvements to be highlighted in the table of contents
- It makes it clearer that "performance improvements" isn't just a small part, but is the focus for this release
- some of the "features and enhancements" are enhancements for the sake of performance, and so should be in the performance section

</details>

## Acceptance Criteria

- [ ] Add a blurb to the 6.1 changelog that indicates this release is primarily focused on performance
- [ ] Remove the "Performance improvements" heading
- [ ] Move everything that was under "Performance improvements" into the regular "Features and enhancements" sectio

## PRs
- https://github.com/silverstripe/developer-docs/pull/774

emteknetnz:
Seems kind of weird to me, because we'll essentially have 2x "enhancements" sections, one for performance enhancements, one for not performance enhancements (assuming there are any, there are though obviously non performance features)

Couldn't we just have the intro say "the focus of this release was performance enhancements", and then just get rid of the `### Performance improvements` H3, and move the H4's under to H3's, that way they'll show in the contents?

That way the changelog will still fit that standard style

GuySartorelli:
Yup that's a good alternative.
My main intention is that stuff doesn't get buried since we don't show anything below 2nd level in the ToC. Moving all the performance stuff into the regular "features and enhancements" solves that.

GuySartorelli:
Have updated the issue - lemme know if that's good.

emteknetnz:
Looks good to me

GuySartorelli:
PR merged

PR #771, opened 2025-06-30
DOC Document query caching

GuySartorelli:
Note that this is a performance enhancement so it belongs in the "performance improvements" section of the changelog - but it is also the sort of thing we want people to be aware of, so it should be in the table of contents. I've raised https://github.com/silverstripe/developer-docs/issues/770 to resolve this conundrum.

## Issue

- https://github.com/silverstripe/silverstripe-framework/issues/11767

### DIFF
--- a/en/02_Developer_Guides/08_Performance/06_ORM.md
+++ b/en/02_Developer_Guides/08_Performance/06_ORM.md
@@ -5,6 +5,64 @@ summary: Configuration and tips for improving ORM performance
 
 # ORM performance
 
+## Caching database queries
+
+Database queries made using [`DataList`](api:SilverStripe\ORM\DataList) or [`SQLSelect`](api:SilverStripe\ORM\Queries\SQLSelect) can be cached. The cached query result lasts until the end of the request, unless invalidated during that request.
+
+This is useful if you make the same query multiple times in the same request.
+
+> [!WARNING]
+> Caching always comes with an increase in memory usage. Make sure you are aware of the trade offs, and only cache queries which are repeated multiple times in the same request, and are not already cached in other ways.
+
+To enable caching on a `DataList`, call the [`DataList::setUseCache`](api:SilverStripe\ORM\DataList::setUseCache()) method.
+
+```php
+use App\Model\MyDataObject;
+
+$cachedList = MyDataObject::get()->setUseCache(true);
+```
+
+To enable caching on a `SQLSelect`, call the [`SQLSelect::setUseCache`](api:SilverStripe\ORM\Queries\SQLSelect) method.
+
+```php
+use SilverStripe\ORM\Queries\SQLSelect;
+
+$cachedQuery = SQLSelect::create(/*...*/)->setUseCache(true, 'some-namespace');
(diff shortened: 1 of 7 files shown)


en/02_Developer_Guides/08_Performance/06_ORM.md
```diff
@@ -5,6 +5,64 @@ summary: Configuration and tips for improving ORM performance
...
+### Invalidating query cache
+
+Queries cached through `DataList` or `SQLSelect` are automatically invalidated in the following scenarios:
+
+- One of the following methods is called on a record of that class or one of its subclasses:
+  - [`write()`](api:SilverStripe\ORM\DataObject::write())
+  - [`delete()`](api:SilverStripe\ORM\DataObject::delete())
+  - [`destroy()`](api:SilverStripe\ORM\DataObject::destroy())
+  - [`flushCache()`](api:SilverStripe\ORM\DataObject::flushCache())
+- One of the following is called on a `DataList` that manages that class or one of its subclasses:
```
michalkleiner: The text obviously works ok without the suggestion, not sure if it was intentionally left out or an oversight, so I thought I'd flag it for consideration.

```suggestion
- One of the following methods is called on a `DataList` that manages that class or one of its subclasses:
```

michalkleiner (approved):
Reads well 👍

en/08_Changelogs/6.1.0.md
```diff
@@ -60,6 +60,32 @@ When created, the composite index will always be called `default_sort_composite`
 
 There are some caveats to this, along with some other options you might want to set, so check out the [indexes documentation](/developer_guides/model/indexes/) for more details.
 
+### Database query caching
```
emteknetnz: If this should be in the performance section then it should have a `####` level heading to match the others

Though I don't this is a performance improvement by itself so should be in a different section?  (unless I've misunderstood the PRs?)  Seems like the PRs are adding a cleaner API that developers can use, though we haven't actually added any additional caching so "out of the box" application performance will be identical? The deprecated API like get_one() has been converted to use new API, and the issue in the original description `SearchFilter::getCaseSensitiveByCollation()` already had caching on a private static property?

All the other things in the performance section will make the CMS run faster after composer update

emteknetnz (changes requested):

en/08_Changelogs/6.1.0.md
ChloeSartorelli: > If this should be in the performance section then it should have a `####` level heading to match the others

Good catch. Updated heading to `####`

> Though I don't this is a performance improvement by itself so should be in a different section?

I guess that depends on how we're categorising "performance improvement" or "performance enhancement".... I've taken it to mean any changes we're making that are performance related, which a new caching functionality is.

en/02_Developer_Guides/08_Performance/06_ORM.md
ChloeSartorelli: Updated.

emteknetnz (approved):

PR #775, opened 2025-07-11
DOC Document generated column support in the ORM

GuySartorelli:
## Issue

- https://github.com/silverstripe/silverstripe-assets/issues/693


en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
```diff
@@ -103,6 +104,52 @@ class Car extends DataObject
 > [!NOTE]
 > `Enum` fields will use the first defined value as the default if you don't explicitly declare one. In the example above, the default value would be "New" if it hadn't been declared.
 
+## Generated columns
+
+Generated columns are database columns where the value is generated inside the database, rather than being set by a user. They're usually based on other columns in the database and can be either generated when the record updates and stored, or generated when requested in which case they aren't stored.
```
michalkleiner: Not sure which suggestion is the one you meant. Either
```suggestion
Generated columns are database columns where the value is generated inside the database, rather than being set by a user. They're usually based on other columns in the database and can be either generated when the record is updated and stored, or generated when requested in which case they aren't stored.
```
or
```suggestion
Generated columns are database columns where the value is generated inside the database, rather than being set by a user. They're usually based on other columns in the database and can be either generated when the record updates are stored, or generated when requested in which case they aren't stored.
```

same in the changelog entry

michalkleiner:
It also seems the index support is only available from MySQL 8, should we mention that?

GuySartorelli:
> It also seems the index support is only available from MySQL 8, should we mention that?

The earliest supported version of MySQL is 8.0 [as of CMS 6.0.0](https://docs.silverstripe.org/en/6/changelogs/6.0.0/#mysql-5-support) so I don't think that needs to be mentioned.

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
ChloeSartorelli: Updated

michalkleiner (approved):
Reads well

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
```diff
@@ -50,6 +50,7 @@ Most `DBField` subclasses will be validated using a [`FieldValidator`](api:Silve
 | `Enum` | [`DBEnum`](api:SilverStripe\ORM\FieldType\DBEnum) | An enumeration of a set of strings that can store a single value | Must be one of the defined values |
 | `Float` | [`DBFloat`](api:SilverStripe\ORM\FieldType\DBFloat) | A floating point number | Must be numeric |
 | `ForeignKey` | [`DBForeignKey`](api:SilverStripe\ORM\FieldType\DBForeignKey) | A special `Int` field used for foreign keys in `has_one` relationships | Must be an int |
+| `GeneratedColumn` | [`DBGeneratedColumn`](api:SilverStripe\ORM\FieldType\DBGeneratedColumn) | A column which has its value generated by the database itself, usually based on other column values | N/A |
```
emteknetnz: I'm not sure if we need this entry if we're using an array syntax, because we could use the keyword 'Generated' / 'GeneratedColumn'

If we do need it, change it to 'Generated' / 'DBGenerated'

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
```diff
@@ -103,6 +104,52 @@ class Car extends DataObject
...
+
+Some good use cases for generated columns include:
+
+- sorting in a gridfield on a complex summary field: It's common to use a getter method to get some value derived from your database fields (e.g. a discounted price), and use that in `summary_fields`. With a generated column, you can remove the getter method and the [`GridFieldSortableHeader`](api:SilverStripe\Forms\GridField\GridFieldSortableHeader) component will be able to sort using the column.
+- using functional indexes: generated columns can be included in indexes, allowing you to sort or filter by complex expressions in an efficient way.
+- data integrity: unlike generating values inside an `onBeforeWrite()` extension hook, generated column values will be correct even if you update the record with raw SQL.
+- reduce repetition: instead of using a complex expression in multiple different places, you can just give the expression a name with a generated column.
+
+The format for adding a generated column is `GeneratedColumn("<datatype>", "<expression>", "<STORED|VIRTUAL>")`. Let's break that down:
```
emteknetnz: Need to update for array syntax

en/08_Changelogs/6.1.0.md
```diff
@@ -99,6 +100,27 @@ If you are using the [`DataObjectSchema::databaseIndexes()`](api:SilverStripe\OR
...
 
+### Generated column support {#generated-columns}
+
+Generated columns are database columns where the value is generated inside the database, rather than being set by a user. They're usually based on other columns in the database and can be either generated when the record is updated and stored, or generated when requested in which case they aren't stored.
+
+Some good use cases for generated columns include:
+
+- sorting in a gridfield on a complex summary field: It's common to use a getter method to get some value derived from your database fields (e.g. a discounted price), and use that in `summary_fields`. With a generated column, you can remove the getter method and the [`GridFieldSortableHeader`](api:SilverStripe\Forms\GridField\GridFieldSortableHeader) component will be able to sort using the column.
+- using functional indexes: generated columns can be included in indexes, allowing you to sort or filter by complex expressions in an efficient way.
+- data integrity: unlike generating values inside an `onBeforeWrite()` extension hook, generated column values will be correct even if you update the record with raw SQL.
```
```suggestion
- data integrity: unlike generating values inside an `onBeforeWrite()` extension hook, generated column values will be correct even if you update the record with raw SQL.
```

Probably also restoring old versions of records and archived records which have had columns added since the old record was created

en/08_Changelogs/6.1.0.md
```diff
@@ -99,6 +100,27 @@ If you are using the [`DataObjectSchema::databaseIndexes()`](api:SilverStripe\OR
...
+- reduce repetition: instead of using a complex expression in multiple different places, you can just give the expression a name with a generated column.
```
emteknetnz, line 112: I don't understand what this means?

Normally you'd just have a getMyValue() method on a model? i.e. logic in a single place?

en/08_Changelogs/6.1.0.md
```diff
@@ -99,6 +100,27 @@ If you are using the [`DataObjectSchema::databaseIndexes()`](api:SilverStripe\OR
...
+
+See [data types and casting](/developer_guides/model/data_types_and_casting/#generated-columns) for details about using generated columns.
+
+#### Support in other SQL servers
+
+Generated column support has been added to the MySQL database connector in `silverstripe/framework`, but is not guaranteed to work for other database servers.
+
+If you maintain a module that adds support for another database server, you'll need to implement the new [`DBSchemaManager::makeGenerated()`](SilverStripe\ORM\Connect\DBSchemaManager::makeGenerated()) and [`DBSchemaManager::needRebuildColumn()`](SilverStripe\ORM\Connect\DBSchemaManager::needRebuildColumn()) methods.
```
```suggestion
If you maintain a module that adds support for another database server, you'll need to implement the new [`DBSchemaManager::makeGenerated()`](SilverStripe\ORM\Connect\DBSchemaManager::makeGenerated()) and [`DBSchemaManager::needRebuildColumn()`](SilverStripe\ORM\Connect\DBSchemaManager::needRebuildColumn()) methods to support generated columns.
```

Presumbably you don't HAVE to update the the module, because there's no BC changes?

emteknetnz (changes requested):

en/08_Changelogs/6.1.0.md
ChloeSartorelli: If you have multiple SQL queries that use the same complex expression, you can replace those complex expressions with a single named generated column.

en/08_Changelogs/6.1.0.md
ChloeSartorelli: You do HAVE to update the module to support generated columns if you want to support generated columns lol. I think "to support generated columns" is very heavily implied but I'll add it to remove any hint of confusion.

en/08_Changelogs/6.1.0.md
ChloeSartorelli: I put the diff in a diff checker and there's no change in your suggested change?

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
ChloeSartorelli: Updated to `Generated`

en/08_Changelogs/6.1.0.md
emteknetnz: Opps

Add something like this (feel free to alter)

This also ensures restored and archived records, with may have new columns since the record was created, remain accurate.

en/08_Changelogs/6.1.0.md
emteknetnz: Still not really registering with me

What would an example of this be using the ORM?

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
```diff
@@ -103,6 +104,52 @@ class Car extends DataObject
...
+
+class Car extends DataObject
+{
+    private static $db = [
+        'Condition' => 'Enum("New,Fair,Junk", "Fair")',
+        'Price' => 'Currency',
+        // Discount the car for a fixed percentage based on the condition of the car
+        // phpcs:ignore Generic.Files.LineLength.TooLong
+        'Discount' => 'Generated("Percentage", "CASE WHEN (\\"Condition\\"=\'Junk\') THEN 0.75 WHEN (\\"Condition\\"=\'Fair\') THEN 0.25 ELSE 0 END", "STORED")',
+        'DiscountPrice' => 'Generated("Currency", "\\"Price\\" * (1.0 - \\"Discount\\")", "VIRTUAL")',
```
emteknetnz, line 157: Probably makes sense to use a heredoc for Discount and keep as a one liner for DiscountPrice to show both ways - see https://github.com/silverstripe/silverstripe-framework/pull/11774#issuecomment-3076005752 for example of the heredoc syntax

emteknetnz (changes requested):

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
ChloeSartorelli: No longer applies

en/08_Changelogs/6.1.0.md
ChloeSartorelli: Done

en/08_Changelogs/6.1.0.md
ChloeSartorelli: - I have some record.
- I want to filter by an expression in a template, so I add a controller method that does a custom SQL query that uses that expression. Even something simple like `CONCAT("FieldOne", "FieldTwo")`.
- In another area of the code that's completely separate from the above controller, I am using the same `DataObject` model. I want to filter or sort by that same expression.

Instead of having `CONCAT("FieldOne", "FieldTwo")` in both places, I can just add a generated column named `FieldConcat` and use that field name instead.

en/08_Changelogs/6.1.0.md
ChloeSartorelli: Done

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
ChloeSartorelli: Done.
Used a nowdoc because it's more appropriate here - we don't want any string interpolation.

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
```diff
@@ -103,6 +104,65 @@ class Car extends DataObject
...
+
+use SilverStripe\ORM\DataObject;
+
+class Car extends DataObject
+{
+    private static $db = [
+        'Condition' => 'Enum("New,Fair,Junk", "Fair")',
+        'Price' => 'Currency',
+        // Discount the car for a fixed percentage based on the condition of the car
+        'Discount' => <<<'SPEC'
```
ChloeSartorelli, line 147: See https://github.com/silverstripe/silverstripe-assets/pull/694#discussion_r2208956301 for why I used `SPEC` as the delimiter.

en/02_Developer_Guides/00_Model/04_Data_Types_and_Casting.md
```diff
@@ -22,6 +22,7 @@ use SilverStripe\ORM\DataObject;
 
 class Player extends DataObject
 {
+    // ...
```
ChloeSartorelli, line 25: Adding these to indicate we're skipping some best practices (e.g. `table_name`) for the sake of brevity.

emteknetnz (approved):

PR #778, opened 2025-07-17
DOC Document backporting non-blocking sessions

GuySartorelli:
I thought about saying something about how it can help normalise hosting between CMS 5 and CMS 6, but:
1. I couldn't think of a good way to word it that doesn't just add more confusion
2. There's no _concrete_ plans to remove dynamodb as a required module for cloud hosting yet, so I don't want to open that can of worms before we have answers to likely questions

## Issue

- https://github.com/silverstripe/.github/issues/401


emteknetnz (approved):

emteknetnz (changes requested):
This PR for the changlog is fine

Should add a PR targetting 5.4 doc to mention this module is available - update the docs themselves similar to https://github.com/silverstripe/developer-docs/pull/733/files

GuySartorelli:
I'll add those docs to the new module and make a PR to add them to the optional features section since it's not part of core.

emteknetnz (approved):

Issue #781, opened 2025-07-17
Document that params should be consistent with properties


emteknetnz:
Follow on from https://github.com/silverstripe/silverstripe-framework/pull/11774#discussion_r2208911817

When creating new methods, and the params used are used to set some properties, then the params should match the property name for consistency

## PRs
- https://github.com/silverstripe/developer-docs/pull/782

GuySartorelli:
PR merged

PR #782, opened 2025-07-17
DOC Add parameter naming code convention

emteknetnz:
Issue https://github.com/silverstripe/developer-docs/issues/781


GuySartorelli (approved):
LGTM

I'll try to remember to do this but at least now it's a rule, so it's easier to say "yup will do" when you ask me to change it :sweat_smile:

Issue #796, opened 2025-08-18
Document configuration properties that are only there for performance reasons


GuySartorelli:
The performance documentation doesn't document some of the configuration properties that are there explicitly for performance purposes - these should be identified and documented.

## PRs

- https://github.com/silverstripe/developer-docs/pull/797
- https://github.com/silverstripe/developer-docs/pull/798

PR #798, opened 2025-08-18
DOC Call out improved performance docs in changelog

GuySartorelli:
Seems worthwhile calling this out since otherwise many people may not notice it.

## Issue

- https://github.com/silverstripe/developer-docs/issues/796


emteknetnz (changes requested):
Having this above "Features and enhancements" is confusing and makes it look like it's a headline feature :-)

Move to be below "Features and enhancements" and above "API changes"

GuySartorelli:
I don't know what a "headline feature" is but this kind of is a headline, given this is the version we focused on performance improvements... but I'll move it to avoid ping pong

emteknetnz (approved):

PR #810, opened 2025-08-28
DOC Use initial-scale instead of maximum scale

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1958


GuySartorelli (approved):
Don't think this is needed but it does no harm I guess.

PR #831, opened 2025-10-21
DOC Document that you cannot rely on env vars set in _config.php in yaml

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/8911

This isn't really a bug as yaml is processed before _config.php. I've simply documented that you cannot define things in _config.php and then use them in yaml.


GuySartorelli (approved):
Yup, makes sense. I had misunderstood the issue, thought it was saying normal env vars weren't being picked up. Probably chillu did too given how he responded.

LGTM.

PR #874, opened 2026-02-10
DOC Change tree ins tags to span

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-cms/issues/3122

### DIFF
--- a/en/08_Changelogs/6.2.0.md
+++ b/en/08_Changelogs/6.2.0.md
@@ -73,6 +73,7 @@ Pagination controls in grid fields and asset-admin now have an improved experien
 - The `a.campaign-info__close` element in campaign admin has been changed to `button.campaign-info__close`.
 - The child elements of `.campaign-info` have been reordered.
 - The `ins.jstree-icon` elements now have `tabindex="0"` so they can receive keyboard focus.
+- The site tree expander arrows have been changed from `<ins>` to `<span>` to remove the incorrect semantic markup.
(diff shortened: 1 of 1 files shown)


en/08_Changelogs/6.2.0.md
```diff
@@ -73,6 +73,7 @@ Pagination controls in grid fields and asset-admin now have an improved experien
 - The `a.campaign-info__close` element in campaign admin has been changed to `button.campaign-info__close`.
 - The child elements of `.campaign-info` have been reordered.
 - The `ins.jstree-icon` elements now have `tabindex="0"` so they can receive keyboard focus.
+- The site tree expander arrows have been changed from `<ins>` to `<span>` so the toggle icons no longer carry button semantics.
```
ChloeSartorelli:
```suggestion
- The site tree expander arrows have been changed from `<ins>` to `<span>` to remove the incorrect semantic markup.
```

`ins` doesn't have button semantics, it was just always semantically incorrect. It _used_ to be used as a "this is an icon" pseudo-standard a long time ago, the but [the real semantics of it is "inserted text"](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/ins).

GuySartorelli (changes requested):

en/08_Changelogs/6.2.0.md
emteknetnz: Updated

GuySartorelli (approved):
LGTM

PR #892, opened 2026-03-26
DOC Update X-Forwarded-Proto header documentation

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11973

### DIFF
--- a/en/02_Developer_Guides/08_Performance/02_HTTP_Cache_Headers.md
+++ b/en/02_Developer_Guides/08_Performance/02_HTTP_Cache_Headers.md
@@ -228,11 +228,24 @@ By default, Silverstripe CMS will output a `Vary` header with the following cont
 Vary: X-Forwarded-Protocol
 ```
 
-To change the value of the `Vary` header, you can change this value by specifying the header in configuration.
+> [!IMPORTANT]
+> For historical reasons the default vary header is `X-Forwarded-Protocol` instead of the standard `X-Forwarded-Proto`.
+> If you are using a CDN or proxy which relies on `X-Forwarded-Proto` to determine the protocol of the request, you should change the default vary header to `X-Forwarded-Proto` to ensure that cached content is correctly served over both HTTP and HTTPS.
+
+To change the value of the `Vary` header, you can change this value by disabling the old header and specifying the new header in configuration:
 
 ```yml
-SilverStripe\Control\HTTP:
-  vary: ""
+SilverStripe\Control\Middleware\HTTPCacheControlMiddleware:
+  defaultVary:
+    X-Forwarded-Protocol: false
+    X-Forwarded-Proto: true
+```
+
+You can also remove the default `Vary` header value by setting it to `null`:
+
+```yml
+SilverStripe\Control\Middleware\HTTPCacheControlMiddleware:
+  defaultVary: null
 ```
 
 Note that if you use `Director::is_ajax()` on cached pages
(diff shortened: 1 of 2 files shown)


en/02_Developer_Guides/08_Performance/02_HTTP_Cache_Headers.md
```diff
@@ -228,11 +228,22 @@ By default, Silverstripe CMS will output a `Vary` header with the following cont
...
 
-To change the value of the `Vary` header, you can change this value by specifying the header in configuration.
+> [!IMPORTANT]
+> For historical reasons the default vary header is `X-Forwarded-Protocol` instead of the standard `X-Forwarded-Proto`.
+> If you are using a CDN or proxy which relies on `X-Forwarded-Proto` to determine the protocol of the request, you should change the default vary header to `X-Forwarded-Proto` to ensure that cached content is correctly served over both HTTP and HTTPS.
+
+To change the value of the `Vary` header, you can change this value by specifying the header in configuration:
 
 ```yml
-SilverStripe\Control\HTTP:
```
emteknetnz, line 234: This was simply wrong, I've updated it to use the correct config

en/02_Developer_Guides/09_Security/05_Secure_Coding.md
```diff
@@ -762,13 +762,19 @@ If you wish to change the headers that are used to find the proxy information, y
 `TrustedProxyMiddleware` service:
 
 ```yml
-SilverStripe\Control\TrustedProxyMiddleware:
```
emteknetnz, line 765: This was also wrong, have updated

Note has a different syntax to the `HTTPCacheControlMiddleware` example above as we set the Injector config here meaning there's no need to 'unset' the default values like we do above setting `HTTPCacheControlMiddleware.defaultVary.X-Forwarded-Protocol: false` -- HTTPCacheControlMiddleware lacks the `getDefaultVary()` and `setDefaultVary()` methods so we need to use that approach above

en/02_Developer_Guides/08_Performance/02_HTTP_Cache_Headers.md
```diff
@@ -228,11 +228,24 @@ By default, Silverstripe CMS will output a `Vary` header with the following cont
...
-  vary: ""
+SilverStripe\Control\Middleware\HTTPCacheControlMiddleware:
+  defaultVary:
+    X-Forwarded-Protocol: false
+    X-Forwarded-Proto: true
+```
+
+You can also remove `X-Forwarded-Protocol` from the `Vary` header by setting it to `null`:
```
ChloeSartorelli:
```suggestion
You can also remove the default `Vary` header value by setting it to `null`:
```

GuySartorelli (changes requested):

en/02_Developer_Guides/08_Performance/02_HTTP_Cache_Headers.md
emteknetnz: I think it's worth explicitly saying it's just the `X-Forwarded-Protocol` as there's still a `Vary: Accept-Encoding` header sent, if we say we are removing the default `Vary` header it implies there will be no `Vary` header sent

ChloeSartorelli: The snippet you've provided is of removing _anything and everyhing_ that has been set in the default vary header, which might be altered by third part modules or other config in the project - saying it's only removing the `X-Forwarded-Protocol` header is incorrect.

Either the snippet or the sentence describing the snippet needs to change.

emteknetnz: Updated

GuySartorelli (approved):
LGTM

